PR #4, opened 2020-04-28
Feature/ios rest

tiholic:
Rest implementation for iOS part of the plugin.
Create a rest instance over ably instance and publish messages.
+ dev dependency upgrades
+ log levels on dart side brought to sync with spec

### DIFF
--- a/example/ios/Runner.xcodeproj/project.pbxproj
+++ b/example/ios/Runner.xcodeproj/project.pbxproj
@@ -80,7 +80,6 @@
 				10B8B89ABB4264117E7AB56C /* Pods-Runner.release.xcconfig */,
 				EA3336021AE374E30F8FCF54 /* Pods-Runner.profile.xcconfig */,
 			);
-			name = Pods;
 			path = Pods;
 			sourceTree = "<group>";
 		};
@@ -185,6 +184,7 @@
 				TargetAttributes = {
 					97C146ED1CF9000F007C117D = {
 						CreatedOnToolsVersion = 7.3.1;
+						DevelopmentTeam = S38X8U35U5;
 					};
 				};
 			};
@@ -393,6 +393,7 @@
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CURRENT_PROJECT_VERSION = "$(FLUTTER_BUILD_NUMBER)";
+				DEVELOPMENT_TEAM = S38X8U35U5;
 				ENABLE_BITCODE = NO;
 				FRAMEWORK_SEARCH_PATHS = (
 					"$(inherited)",
@@ -523,6 +524,7 @@
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CURRENT_PROJECT_VERSION = "$(FLUTTER_BUILD_NUMBER)";
+				DEVELOPMENT_TEAM = S38X8U35U5;
 				ENABLE_BITCODE = NO;
 				FRAMEWORK_SEARCH_PATHS = (
 					"$(inherited)",
@@ -546,6 +548,7 @@
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CURRENT_PROJECT_VERSION = "$(FLUTTER_BUILD_NUMBER)";
+				DEVELOPMENT_TEAM = S38X8U35U5;
 				ENABLE_BITCODE = NO;
 				FRAMEWORK_SEARCH_PATHS = (
 					"$(inherited)",
--- a/ios/Classes/AblyFlutterPlugin.m
+++ b/ios/Classes/AblyFlutterPlugin.m
@@ -54,15 +54,11 @@ static FlutterHandler _publishRestMessage = ^void(AblyFlutterPlugin *const plugi
     AblyFlutterMessage *const messageData = message.message;
     NSMutableDictionary<NSString *, NSObject *>* _dataMap = messageData.message;
     NSString *channelName = (NSString*)[_dataMap objectForKey:@"channel"];
-    NSString *eventName = (NSString*)[_dataMap objectForKey:@"name"];
-    NSObject *eventData = (NSString*)[_dataMap objectForKey:@"message"];
+    NSString *const eventName = (NSString*)[_dataMap objectForKey:@"name"];
+    NSObject *const eventData = (NSString*)[_dataMap objectForKey:@"message"];
     ARTRest *client = [ably getRest:messageData.handle];
     ARTRestChannel *channel = [client.channels get:channelName];
 
-    //Handling nil cases, to be generalized based on other encounters
-    NSString *nilStr = nil;
-    if([eventName isKindOfClass:[NSNull class]]){ eventName = nilStr; }
-    if([eventData isKindOfClass:[NSNull class]]){ eventData = nilStr; }
     [channel publish:eventName data:eventData callback:^(ARTErrorInfo *_Nullable error){
         if(error){
             result([
--- a/ios/Classes/AblyFlutterPlugin.m
+++ b/ios/Classes/AblyFlutterPlugin.m
@@ -39,6 +39,44 @@ static FlutterHandler _register = ^void(AblyFlutterPlugin *const plugin, Flutter
     [plugin registerWithCompletionHandler:result];
 };
 
+static FlutterHandler _createRestWithOptions = ^void(AblyFlutterPlugin *const plugin, FlutterMethodCall *const call, const FlutterResult result) {
+    AblyFlutterMessage *const message = call.arguments;
+    LOG(@"message for handle %@", message.handle);
+    AblyFlutter *const ably = [plugin ablyWithHandle:message.handle];
+    // TODO if ably is nil here then an error response, perhaps? or allow Dart side to understand null response?
+    result([ably createRestWithOptions:message.message]);
+};
+
+static FlutterHandler _publishRestMessage = ^void(AblyFlutterPlugin *const plugin, FlutterMethodCall *const call, const FlutterResult result) {
+    AblyFlutterMessage *const message = call.arguments;
+    LOG(@"message for handle %@", message.handle);
+    AblyFlutter *const ably = [plugin ablyWithHandle:message.handle];
+    AblyFlutterMessage *const messageData = message.message;
+    NSMutableDictionary<NSString *, NSObject *>* _dataMap = messageData.message;
+    NSString *channelName = (NSString*)[_dataMap objectForKey:@"channel"];
+    NSString *eventName = (NSString*)[_dataMap objectForKey:@"name"];
+    NSObject *eventData = (NSString*)[_dataMap objectForKey:@"message"];
+    ARTRest *client = [ably getRest:messageData.handle];
+    ARTRestChannel *channel = [client.channels get:channelName];
+
+    //Handling nil cases, to be generalized based on other encounters
+    NSString *nilStr = nil;
+    if([eventName isKindOfClass:[NSNull class]]){ eventName = nilStr; }
+    if([eventData isKindOfClass:[NSNull class]]){ eventData = nilStr; }
+    [channel publish:eventName data:eventData callback:^(ARTErrorInfo *_Nullable error){
+        if(error){
+            result([
+                    FlutterError
+                    errorWithCode:[NSString stringWithFormat: @"%ld", (long)error.code]
+                    message:[NSString stringWithFormat:@"Unable to publish message to Ably server; err = %@", [error message]]
+                    details:nil
+                    ]);
+        }else{
+            result(nil);
+        }
+    }];
+};
+
 static FlutterHandler _createRealtimeWithOptions = ^void(AblyFlutterPlugin *const plugin, FlutterMethodCall *const call, const FlutterResult result) {
     AblyFlutterMessage *const message = call.arguments;
     LOG(@"message for handle %@", message.handle);
@@ -92,6 +130,8 @@ static FlutterHandler _dispose = ^void(AblyFlutterPlugin *const plugin, FlutterM
         @"getPlatformVersion": _getPlatformVersion,
         @"getVersion": _getVersion,
         @"register": _register,
+        @"createRestWithOptions": _createRestWithOptions,
+        @"publish": _publishRestMessage,
         @"createRealtimeWithOptions": _createRealtimeWithOptions,
         @"connectRealtime": _connectRealtime,
         @"dispose": _dispose,
--- a/lib/src/spec/constants.dart
+++ b/lib/src/spec/constants.dart
@@ -1,8 +1,8 @@
 class LogLevel{
-  static final int NONE = 99;
-  static final int VERBOSE = 2;
-  static final int DEBUG = 3;
-  static final int INFO = 4;
-  static final int WARN = 5;
-  static final int ERROR = 6;
+  static final int none = 99;
+  static final int verbose = 2;
+  static final int debug = 3;
+  static final int info = 4;
+  static final int warn = 5;
+  static final int error = 6;
 }
--- a/lib/src/spec/constants.dart
+++ b/lib/src/spec/constants.dart
@@ -1,0 +1,8 @@
+class LogLevel{
+  static final int NONE = 99;
+  static final int VERBOSE = 2;
+  static final int DEBUG = 3;
+  static final int INFO = 4;
+  static final int WARN = 5;
+  static final int ERROR = 6;
+}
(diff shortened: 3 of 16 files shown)


ios/Classes/AblyFlutterPlugin.m
```diff
@@ -39,6 +39,44 @@ -(nullable AblyFlutter *)ablyWithHandle:(NSNumber *)handle;
...
+    NSString *channelName = (NSString*)[_dataMap objectForKey:@"channel"];
+    NSString *eventName = (NSString*)[_dataMap objectForKey:@"name"];
+    NSObject *eventData = (NSString*)[_dataMap objectForKey:@"message"];
+    ARTRest *client = [ably getRest:messageData.handle];
+    ARTRestChannel *channel = [client.channels get:channelName];
+
+    //Handling nil cases, to be generalized based on other encounters
+    NSString *nilStr = nil;
+    if([eventName isKindOfClass:[NSNull class]]){ eventName = nilStr; }
+    if([eventData isKindOfClass:[NSNull class]]){ eventData = nilStr; }
```
tiholic, line 65: @QuintinWillison any scope for improvement here?

As `nil` is not a valid argument for `name` or `data`, we need to pass on an argument of type `NSString`, `NSData`, `NSArray` or `NSDictionary`

example/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -185,6 +184,7 @@
 				TargetAttributes = {
 					97C146ED1CF9000F007C117D = {
 						CreatedOnToolsVersion = 7.3.1;
+						DevelopmentTeam = S38X8U35U5;
```
QuintinWillison, line 187: Who this?

ios/Classes/AblyFlutterPlugin.m
```diff
@@ -39,6 +39,44 @@ -(nullable AblyFlutter *)ablyWithHandle:(NSNumber *)handle;
...
+    result([ably createRestWithOptions:message.message]);
+};
+
+static FlutterHandler _publishRestMessage = ^void(AblyFlutterPlugin *const plugin, FlutterMethodCall *const call, const FlutterResult result) {
+    AblyFlutterMessage *const message = call.arguments;
+    LOG(@"message for handle %@", message.handle);
+    AblyFlutter *const ably = [plugin ablyWithHandle:message.handle];
+    AblyFlutterMessage *const messageData = message.message;
+    NSMutableDictionary<NSString *, NSObject *>* _dataMap = messageData.message;
+    NSString *channelName = (NSString*)[_dataMap objectForKey:@"channel"];
```
QuintinWillison, line 56: Please make pointers `const` unless you have a reason for them not to be. In the case of this line that means:

```objc
NSString *const channelName = [_dataMap objectForKey:@"channel"];
```

ios/Classes/AblyFlutterPlugin.m
QuintinWillison: `objectForKey:` will return `nil` if there is no entry in the data map. Surely you could just not pack an entry on the Dart side?

Also... potentially about to disappear given my suggestion but no need for the `nilStr` intermediary here. `nil` is designed to be assigned directly to any type of Objective-C instance pointer (a.k.a. `id`).

ios/Classes/AblyFlutterPlugin.m
```diff
@@ -39,6 +39,44 @@ -(nullable AblyFlutter *)ablyWithHandle:(NSNumber *)handle;
...
+    NSString *eventName = (NSString*)[_dataMap objectForKey:@"name"];
+    NSObject *eventData = (NSString*)[_dataMap objectForKey:@"message"];
+    ARTRest *client = [ably getRest:messageData.handle];
+    ARTRestChannel *channel = [client.channels get:channelName];
+
+    //Handling nil cases, to be generalized based on other encounters
+    NSString *nilStr = nil;
+    if([eventName isKindOfClass:[NSNull class]]){ eventName = nilStr; }
+    if([eventData isKindOfClass:[NSNull class]]){ eventData = nilStr; }
+    [channel publish:eventName data:eventData callback:^(ARTErrorInfo *_Nullable error){
```
QuintinWillison, line 66: The `_Nullable` here is only of interest to callers, not the implementer, so is superfluous. Also no reason not to make that `error` pointer `const` (swap the `_Nullable` for `const`).

ios/Classes/AblyFlutterPlugin.m
tiholic: @QuintinWillison we do send an entry from dart side, but that entry would be null as passed by package users.

Ex: `restInstance.publish('name', 'message')` will pass both values, but `restInstance.publish()` will pass an map with `{"message": null, "data": null}` which will be available as `nil` in iOS side.

`nil` is not accepted by publish method, as the encoding fails
argument passed: `must be NSString, NSData, NSArray or NSDictionary`

Here is link to the log which states the error:
https://ably-real-time.slack.com/files/U0118SXS2VD/F012UMW8DCZ/untitled

example/ios/Runner.xcodeproj/project.pbxproj
tiholic: That's my personal development team. Seems like this file must be committed.
Let me know if there is a way to ignore this.

Note: this is project file in example app

ios/Classes/AblyFlutterPlugin.m
tiholic: Handled in next PR

ios/Classes/AblyFlutterPlugin.m
tiholic: Handled in next PR

example/ios/Runner.xcodeproj/project.pbxproj
mattheworiordan: @QuintinWillison do we need to commit these XCode files?  I assume we do given this is an example app and we want it to just work, but ideally we'd not commit this stuff.

ios/Classes/AblyFlutterPlugin.m
mattheworiordan: @tiholic I think it would be preferable to have complete PRs, and reduce the sope.  I appreciate there is ongoing work, and you want to get stuff reviewed early, which is good.  However, if the answer is I will fix in the next PR, it means we cannot really review any PRs in full because the feedback is addressed in a separate review, which may or may not be done by the same people.  I think moving forward, we should aim to have all feedback on PRs reviewed and addressed in that PR.  This will ensure we are all happy about code we are reviewing being merged.  WDYT?

ios/Classes/AblyFlutterPlugin.m
```diff
@@ -39,6 +39,44 @@ -(nullable AblyFlutter *)ablyWithHandle:(NSNumber *)handle;
...
 
+static FlutterHandler _createRestWithOptions = ^void(AblyFlutterPlugin *const plugin, FlutterMethodCall *const call, const FlutterResult result) {
+    AblyFlutterMessage *const message = call.arguments;
+    LOG(@"message for handle %@", message.handle);
+    AblyFlutter *const ably = [plugin ablyWithHandle:message.handle];
+    // TODO if ably is nil here then an error response, perhaps? or allow Dart side to understand null response?
+    result([ably createRestWithOptions:message.message]);
+};
+
+static FlutterHandler _publishRestMessage = ^void(AblyFlutterPlugin *const plugin, FlutterMethodCall *const call, const FlutterResult result) {
```
mattheworiordan, line 50: @QuintinWillison I may need some enlightenment here, but could do these methods not directly map to actual IDL methods instead of `publishRestMessage` which does not exist in our IDL.  Is there a reason for this that I am not aware of due to the constraints of Dart?  I do understand there are issues with overloaded methods with Dart, but I am surprised to see code like this where, as the reviewer, I am not entirely clear what I am actually reviewing.

@tiholic separately, could I ask that we include code examples in the README for methods we are delivering as we go along?  If you look at other SDKs, such as Ruby and the publish method, the high level methods are documented in the README - https://github.com/ably/ably-ruby#publishing-a-message-to-a-channel-1.  We don't need to document every method (which is what we have done in all READMEs), however it is imperative for me to see what API interface we are exposing to delivers when using the Flutter SDK.  Could you aim to do that as you go along?

lib/src/spec/constants.dart
```diff
@@ -0,0 +1,8 @@
+class LogLevel{
+  static final int NONE = 99;
+  static final int VERBOSE = 2;
+  static final int DEBUG = 3;
+  static final int INFO = 4;
+  static final int WARN = 5;
+  static final int ERROR = 6;
```
mattheworiordan, line 7: Is it idiomatic to have LogLevel with `NONE` as opposed to `None? 
What do Dart developers prefer / use normally?

ios/Classes/AblyFlutterPlugin.m
tiholic: Yes, agreed @mattheworiordan

tiholic:
@mattheworiordan 
Regarding PR Resolution and README, will take care.

Coming to the idiomatic vs. consistency, as we are pretty consistent in terms on simple method calls. In context of event listeners, we will be having idiomatic approach using Streams.

We are not inclined to following idiomatic approach w.r.t usage of [named constructors](https://dart.dev/guides/language/language-tour#named-constructors). Standard constructors with [asserts in initializer list](https://github.com/dart-lang/sdk/blob/master/docs/language/informal/assert-in-initializer-list.md) is preferred over named constructors as that is most consistent with IDL.

example/ios/Runner.xcodeproj/project.pbxproj
QuintinWillison: Yes, you do need to commit them. They are required for the Xcode project to build and they cannot be automatically built unless you start employing other tech like make which is not idiomatic for iOS devs. Xcode project files are standard currency in Apple development.

ios/Classes/AblyFlutterPlugin.m
QuintinWillison: 💯

ios/Classes/AblyFlutterPlugin.m
QuintinWillison: That makes sense. However I was imagining that you had control, in Dart land, of the production of the data that ends up causing `_dataMap` to be populated once in Objective-C land. Therefore if the Dart code knows that it needs to set a value of `null` for the `name` key then it can simply not pack a value for that key at all. Or am I missing something?

ios/Classes/AblyFlutterPlugin.m
QuintinWillison: To expand on my agreement with Matt's comment, the ideal is that I can click on the 'Resolve Conversation' button for this comment thread, solely in the context of this PR. Thus once I've seen you push an additional commit to this PR that fixes this issue then I can resolve the comment and therefore we can all visually see progress on this PR towards it being suitable to merge.

Clearly this approach of iterating on this PR comes with the slight downside that your subsequent PR underlying branches (#5 and knock on #6) will need to be rebased on top of the branch underlying this one. But then that's the maintenance overhead of having several chained PRs 'in flight' and is pretty much unavoidable.

Hopefully by sticking to the rules and keeping things clean in terms of branch management and maintenance while we go through the review process will keep the process easy and predictable for you. I'm happy to help if any of that is unclear or you have suggestions for an alternative workflow.

lib/src/spec/constants.dart
QuintinWillison: Matt's got a valid point here and it raises something else I should have spotted, sorry!

First, yes, we should be aiming to be as Dart-idiomatic as possible and that means adhering to [Effective Dart](https://dart.dev/guides/language/effective-dart) wherever possible - thus these, as 'constant names' / 'constant variables' should [`PREFER` using `lowerCamelCase`](https://dart.dev/guides/language/effective-dart/style#prefer-using-lowercamelcase-for-constant-names) - i.e. `none`, `verbose`, `debug`, etc..

Secondly, I think you can and should be using `const` rather than `final` as these values can be 'calculated' at compile time. See [this answer](https://stackoverflow.com/a/50431087/392847).

ios/Classes/AblyFlutterPlugin.m
QuintinWillison: Hopefully your commits are suitably granular to support illustrating in this manner but perhaps it would help if you could comment here with a link to the relevant commit so I can resolve [this thread](https://github.com/ably/ably-flutter/pull/4#discussion_r417630602).

Going forward I do want to [encourage atomic iteration on the same PR](https://github.com/ably/ably-flutter/pull/4#discussion_r419337789), but for now happy to unblock merging this PR if you can just comment in this manner here so I can resolve this comment thread. 😁

ios/Classes/AblyFlutterPlugin.m
QuintinWillison: @mattheworiordan ...

The particular block name you've picked up on (`_publishRestMessage`) is an internal implementation detail unique to the Objective-C code (i.e. just a third of the solution, and in this case just this file - it's `static`!). Right now I feel we need to concentrate on the public API which basically means what's presented to Flutter users in pure Dart code. Any illogical mappings will naturally flat to the surface as that evolves.

Regarding your comment to @tiholic relating to examples... The Flutter-plugin-idomatic way of delivering example usage, as I understand it, is in the code in the example folder. It's created for you, out of the box, when you use Flutter's tooling to [create a plugin project](https://flutter.dev/docs/development/packages-and-plugins/developing-packages#step-1-create-the-package-1). It may make sense to also have examples in the readme, or perhaps an indirection from there to the 'live and runnable' Flutter example code. This may need more thought but I think that will shake out a bit more once we have enough implementation in place to do something useful with the plugin!

example/ios/Runner.xcodeproj/project.pbxproj
QuintinWillison: @tiholic my preference would be to remove these lines from this file and then just be mindful whenever you commit changes to this file to make sure you omit chunks that are personalised like this. We can discuss offline how to handle this when doing iOS development (short answer: nice tooling, e.g. [SourceTree](https://www.sourcetreeapp.com/)).

ios/Classes/AblyFlutterPlugin.m
mattheworiordan: Ok, thanks for explaining.  

> Right now I feel we need to concentrate on the public API which basically means what's presented to Flutter users in pure Dart code.

Agreed

> The Flutter-plugin-idomatic way of delivering example usage, as I understand it, is in the code in the example folder.

What wasn't what I meant.  What I meant is is the Ably API we provide from the Flutter SDK idiomatic to Dart developers?  If not, what are the trade offs?  

I was simply requesting that as we go along, we include the code samples in the README to be consistent with our SDKs - hence https://github.com/ably/ably-ruby#publishing-a-message-to-a-channel-1

mattheworiordan:
> Coming to the idiomatic vs. consistency, as we are pretty consistent in terms on simple method calls. In context of event listeners, we will be having idiomatic approach using Streams.

Sounds good.

> We are not inclined to following idiomatic approach w.r.t usage of named constructors. Standard constructors with asserts in initializer list is preferred over named constructors as that is most consistent with IDL.

Great, any reason not to do the latter then?

tiholic:
>  Great, any reason not to do the latter then?

Only reason is that the former is more consistent with IDL.

@paddybyers @QuintinWillison and @Ugbot please share your inputs. We finalized this approach over the call.

ios/Classes/AblyFlutterPlugin.m
tiholic: How would it help if we don't send the key to Obj-C side? `publish` method still expects values and publishing fails at encoding stage.

ios/Classes/AblyFlutterPlugin.m
tiholic: @QuintinWillison https://github.com/ably/ably-flutter/commit/8b2817d305608163c32ae3463a35790511c73089

ios/Classes/AblyFlutterPlugin.m
tiholic: @mattheworiordan @QuintinWillison 

will have the high level API available in README in next PR.

lib/src/spec/constants.dart
tiholic: Updated in 822e6407feea365d548a0282b15880065f798e6f

ios/Classes/AblyFlutterPlugin.m
tiholic: @QuintinWillison as discussed, this is sorted in https://github.com/ably/ably-flutter/commit/6ce0328481388fb75de3f95b21b6306b32f8c723 bu not packing the key form dart side

tiholic:
@mattheworiordan had a long discussion with @QuintinWillison and we decided to follow idiomatic approach over consistency in case of named constructors.

Code not following idiomatic approach will be updated in PRs raised here further.

PR #6, opened 2020-05-03
Pilot implementation for Realtime event based API's - ConnectionStateChange

tiholic:
### Realtime events workflow based on Streams handled for both Android and iOS.

>  Using [StreamChannel](https://pub.dev/packages/streams_channel), which is based on [EventChannel](https://api.flutter.dev/flutter/services/EventChannel-class.html)

* StreamsChannel (EventChannel) is setup on both Android and iOS
* `ConnectionStateChange` is fully supported
* Better way to handle codecs using `CodecPair<Type>(encoder, decoder)`
  * Supported only on Dart and Android side.
  * Objective C has separate reader and writer implementations.

#### Known Issues - Applies only for Event Streams:
* Hot-Restart will not dispose of platform side objects
  * This is a concern only for development phase
* In production, the behavior is as below:
  * On minimising the app (background) - instances stay intact (dart side will be perfectly connected to platform side)
  * On killing the app - all instances will be created afresh and linked as expected

>  NOTE: A fix for the above issue is an experimental WIP, and changes related to the same is not available in the repo.


tiholic:
This PR is a continuation for previous PR's. Number of changes will be reduced as pending PR's are merged.

android/src/main/java/io/ably/flutter/plugin/AblyEventStreamHandler.java
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        @Override
+        public void error(final String s, final String s1, final Object o) {
+            handler.post(() -> eventSink.error(s, s1, o));
+        }
+
+        @Override
+        public void endOfStream() {
+            //TODO work on this if required, or remove this TODO once all features are covered
```
mattheworiordan, line 43: As with other PR, can we ensure TODO tasks carry an issue number if in PR?

android/src/main/java/io/ably/flutter/plugin/AblyEventStreamHandler.java
```diff
@@ -0,0 +1,107 @@
...
+
+import android.os.Handler;
+import android.os.Looper;
+
+import io.ably.lib.realtime.ChannelStateListener;
+import io.ably.lib.realtime.ConnectionStateListener;
+import io.flutter.plugin.common.EventChannel;
+
+
+public class AblyEventStreamHandler implements EventChannel.StreamHandler {
```
mattheworiordan, line 11: Can I ask that we add some high level comments on what AblyEventStreamHandler does, even if just a one-liner for others who review this code in future?

mattheworiordan: I think this applies generally where we have classes that serve a shared function, but don't map to Ably well known types

android/src/main/java/io/ably/flutter/plugin/AblyEventStreamHandler.java
```diff
@@ -0,0 +1,107 @@
...
+            System.out.println("Cannot process null input on cancel");
+            return;
+        }
+        methodCallHandler.<AblyFlutterMessage<String>>ablyDo((AblyFlutterMessage)object, (ablyLibrary, message) -> {
+            String eventName = message.message;
+            switch (eventName) {
+                case "realtime_onConnectionStateChanged":
+                    ablyLibrary.getRealtime(message.handle).connection.off(connectionStateListener);
+                case "realtime_onChannelStateChanged":
+                    // ablyLibrary.getRealtime(handle).connection.off(connectionStateListener);
```
mattheworiordan, line 102: Why is this commented code?

android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
```diff
@@ -110,4 +207,44 @@ private Object readTokenDetails(final ByteBuffer buffer) {
...
+
+    private void writeChannelEvent(ByteArrayOutputStream stream, ChannelEvent e){ writeEnum(stream, _channelEvent, e); }
+    private void writeChannelState(ByteArrayOutputStream stream, ChannelState e){ writeEnum(stream, _channelState, e); }
+    private void writeChannelStateChange(ByteArrayOutputStream stream, ChannelStateListener.ChannelStateChange c){
+        stream.write(_channelStateChange);
+        writeValue(stream, c.current);
+        writeValue(stream, c.previous);
+        writeValue(stream, c.event);
+        writeValue(stream, c.resumed);
+        writeValue(stream, c.reason);
```
mattheworiordan, line 242: As mentioned before in my other PRs, we are now making this problem worse, keeping the ordering of these stream writes and reads in sync.  I really dislike this.

ios/Classes/codec/AblyFlutterReader.h
```diff
@@ -0,0 +1,28 @@
...
+
+NS_ASSUME_NONNULL_END
+
+
+typedef NS_ENUM(UInt8, _Value) {
+    //Ably flutter plugin protocol message
+    _valueAblyMessage = 128,
+    
+    //Other ably objects
+    _valueClientOptions = 129,
```
mattheworiordan, line 17: More craziness...  Sorry. I just cannot warm to this.

android/src/main/java/io/ably/flutter/plugin/AblyEventStreamHandler.java
tiholic: @mattheworiordan it is a WIP, so is commented out.

android/src/main/java/io/ably/flutter/plugin/AblyEventStreamHandler.java
tiholic: Added https://github.com/ably/ably-flutter/issues/9

ios/Classes/codec/AblyFlutterReader.h
tiholic: @mattheworiordan while we can solve the ordering issue by using a map/dict, there is nothing we can do here. Flutter's [implementation to encode and decode default supported types](https://flutter.dev/docs/development/platform-integration/platform-channels?tab=ios-channel-objective-c-tab#codec) also follows [the same approach](https://github.com/flutter/flutter/blob/f139b11009/packages/flutter/lib/src/services/message_codecs.dart#L265), so I would say we should get used to this :)

If you have any inputs/alternate suggestions, we can explore that.

android/src/main/java/io/ably/flutter/plugin/AblyEventStreamHandler.java
```diff
@@ -0,0 +1,107 @@
...
+
+public class AblyEventStreamHandler implements EventChannel.StreamHandler {
+    private final AblyMethodCallHandler methodCallHandler;
+
+    AblyEventStreamHandler(AblyMethodCallHandler methodCallHandler){
+        this.methodCallHandler = methodCallHandler;
+    }
+
+    /**
+     * Refer to the comments on AblyMethodCallHandler.MethodResultWrapper
```
QuintinWillison, line 19: I've just had a look at those comments and couldn't work out why this is required. Is there something missing either there or here or have the comments gone out of sync?

android/src/main/java/io/ably/flutter/plugin/AblyEventStreamHandler.java
```diff
@@ -0,0 +1,107 @@
...
+
+    private class PluginChannelStateListener extends Listener implements ChannelStateListener {
+        PluginChannelStateListener(EventChannel.EventSink eventSink){super(eventSink);}
+        public void onChannelStateChanged(io.ably.lib.realtime.ChannelStateListener.ChannelStateChange stateChange){
+            eventSink.success(stateChange);
+        }
+    }
+
+    @Override
+    public void onListen(Object object, EventChannel.EventSink uiThreadEventSink) {
```
QuintinWillison, line 71: How many times in the lifecycle of the Flutter app and/or plugin does this method get called? If it's only meant to be once then I think it would be helpful to find a way to ensure that you only call `connection.on` once, leaving event routing to the Dart-side. How you ensure that could vary from a simple assert (e.g. has the `connectionStateListener` field already been set to non-`null`?) to restructuring the code to make the intent and/or expectation clearer.

android/src/main/java/io/ably/flutter/plugin/AblyEventStreamHandler.java
QuintinWillison: Perhaps preferable to raise an exception or trigger an assertion failure here so it fails early at runtime.

ios/Classes/AblyFlutterStreamHandler.h
```diff
@@ -0,0 +1,21 @@
...
+@interface AblyFlutterStreamHandler : NSObject<FlutterStreamHandler>
+
++(instancetype)new NS_UNAVAILABLE;
+-(instancetype)init NS_UNAVAILABLE;
+
+-(instancetype)initWithAbly:(AblyFlutterPlugin *const)plugin NS_DESIGNATED_INITIALIZER;
+
+@property(nonatomic, readonly) AblyFlutterPlugin * plugin;
+
+- (FlutterError* _Nullable)onListenWithArguments:(id _Nullable)arguments
```
QuintinWillison, line 16: Really minor thing but in Objective-C there are scenarios, and I think this is one of them, where you can prefix the type with `nullable` which is slightly nicer to read than `_Nullable` ... i.e. `-(nullable FlutterError *)onListenWithArguments:(nullable id)arguments`. Slightly prettier, is all.

Also, cleaner to call the `events` argument name placeholder `eventSink` to match the selector signature. This also makes it a bit clearer that it's a block.

QuintinWillison:
You need to rebase your commits against the target branch as there are conflicts which may be making review harder. Thanks!

tiholic:
@QuintinWillison rebase complete

ios/Classes/codec/AblyFlutterReader.h
QuintinWillison: @tiholic perhaps we could get rid of our custom codec entirely and **always** use a `Map` for objects?

You'll then need a set of serialisers in each language that are capable of encoding and/or decoding the classes/structs that need to go across the native boundary.  Where there are trees of nested structures to be encoded then clearly the encoding serializers also need to nest accordingly to match. Same for the decoding serializers too.

In terms of map keys, those can then be localised to the decoder implementation and can all be of type `String` so that their representations conform exactly to the names in [the spec](https://docs.ably.io/client-lib-development-guide/features/). Essentially, a JSON-like approach - hence a good comparable is the functionality of Apple's [NSJSONSerialization](https://developer.apple.com/documentation/foundation/nsjsonserialization) in terms of how that implementation has been engineered only to understand a focussed subset of the [Foundation](https://developer.apple.com/documentation/foundation?language=objc) data types.

Clearly these encoding and decoding serialisers are no longer coupled to the codec. We leave that in it's Flutter-native form as the 'transport layer'. It does what it does and then we handle our own protocol in our own layer above that.

As a final thought, perhaps worth keeping [YAGNI](https://martinfowler.com/bliki/Yagni.html) in mind so not writing a decoder or encoder for a particular platform if you don't need one. A serializer should be capable of being either a decoder, or encoder, or both, as required.

pubspec.yaml
```diff
@@ -9,17 +9,18 @@ environment:
...
     sdk: flutter
 
   # Specifying a slightly older version minimum for the 'pure' Dart test package
   # because the flutter_test package needs an older version of test_api.
-  test: ^1.14.2
+  test: ^1.9.4
 
   # Stricter Linting
-  pedantic: ^1.9.0
+  pedantic: ^1.0.0
```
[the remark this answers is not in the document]
tiholic: @QuintinWillison downgraded due to incompatibility issues among those 2 dependencies.

example/lib/main.dart
```diff
@@ -29,6 +29,7 @@ class _MyAppState extends State<MyApp> {
   ably.Ably _ablyPlugin;
   ably.Realtime _realtime;
   ably.Rest _rest;
+  ably.ConnectionState _realtimeStatus;
```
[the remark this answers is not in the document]
tiholic: Right, updated in c93bb90952df2a1e78314981c70411793b6a9773

android/src/main/java/io/ably/flutter/plugin/AblyEventStreamHandler.java
tiholic: Added docstring in d59f89a2e4a18ccdcef6443f45d90ac453507501

android/src/main/java/io/ably/flutter/plugin/AblyEventStreamHandler.java
tiholic: @QuintinWillison raising an exception here just kills the app.
`onCancel` method can be called only if a listener is instantiated.

So, onListen is updated to throw runtime exception in d59f89a2e4a18ccdcef6443f45d90ac453507501

example/lib/main.dart
```diff
@@ -147,23 +141,42 @@ class _MyAppState extends State<MyApp> {
 
     ably.Realtime realtime;
     try {
-      realtime = await _ablyPlugin.createRealtime(clientOptions);
+      realtime = await _ablyPlugin.createRealtime(options: clientOptions);
+
+      ///One can listen from multiple listeners on the same event,
```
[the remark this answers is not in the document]
tiholic: Noted.
870dbee7d7106b7529f14de49c63454b8166438f

android/src/main/java/io/ably/flutter/plugin/AblyEventStreamHandler.java
tiholic: @QuintinWillison 

The comment in `AblyMethodCallHandler.MethodResultWrapper` is basically pointing to this issue https://github.com/flutter/flutter/issues/34993#issue-459900986

android/src/main/java/io/ably/flutter/plugin/AblyEventStreamHandler.java
tiholic: This method will be called just once per instance of `AblyEventStreamHandler`, but there will be `n` instances of `AblyEventStreamHandler` for `n` listeners on dart side.

ios/Classes/AblyFlutterStreamHandler.h
tiholic: Handled in 1542ced08593476576c72f0a4262282e34967b10

ios/Classes/codec/AblyFlutterReader.h
tiholic: Created an issue for further discussions on this.
I will explore on your inputs and share my feedback @ https://github.com/ably/ably-flutter/issues/12

android/src/main/java/io/ably/flutter/plugin/AblyEventStreamHandler.java
paddybyers: I discussed this with Quintin and I think we should be minimising the number of places where we need to maintain a correspondence between a dart entity and a plugin entity. I would have expected the dart code to subscribe only once, unconditionally, to connection state events, and the dart code can handle adding/removing listeners without having to add and remove listeners each time on the plugin side.

The entities on the plugin side whose lifecycle needs to be aligned with entities on the dart side are:

- rest/realtime library instance. There are several elements on the plugin side - auth, connection, push, etc - whose lifecycle is the same as the enclosing library instance.

- realtime channel. These are created on the dart side, and a corresponding entity needs to be created on the plugin side. These have state transition events (in the same way as `connection`) but again, there can be a single channel state listener on the plugin side, and all event multicasting happens on the dart side.

I can't think of anything else whose lifecycle needs to be synchronised between the two. All other types can be treated as value types.

tiholic: > we should be minimising the number of places where we need to maintain a correspondence between a dart entity and a plugin entity

@paddybyers current implementation has lesser maintenance. While what you are suggesting sounds like an optimistic approach, we will need to manage all the n-1 mappings for each event.

paddybyers: @tiholic Honestly I think you're underestimating the complexity of the work to keep the set of listeners in sync. Consider this requirement: https://docs.ably.io/client-lib-development-guide/features/#RTE6a. You're going to need a collection of listeners on the dart side, and a corresponding collection on the plugin side; when there is an `emit()`, the native library will determine which listeners to call, and will call them in sequence. For each of them, an event will be streamed back to the corresponding listener on the dart side. Suppose the client's handler for one listener now deletes another listener that hasn't yet been called, but is going to be called: you have to remove it from the local collection, and trigger removal from the plugin, but still ensure that it gets called when an event is streamed from the plugin.

The alternative, ISTM, is very simple. You have a single, permanent listener that the plugin attaches to the native library. You have a collection of listeners on the dart side, which is initially empty. When an event comes from the plugin to the dart side, you just multicast it to the listeners that exist, and implement the eventemitter spec. I just don't see how that's less maintainable - it seems significantly simpler and more maintainable to me.

pubspec.yaml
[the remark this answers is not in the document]
QuintinWillison: Could you please inject some commentary to this effect. Ideally with links to issues or blog posts or SO answers or whatever it was that told you this was the right choice. Someone coming back to here and thinking "hey. I'll just upgrade this" might be wasting their time without some breadcrumbs to start from. Thanks.

android/src/main/java/io/ably/flutter/plugin/AblyEventStreamHandler.java
QuintinWillison: @tiholic I've commented on that commit.

android/src/main/java/io/ably/flutter/plugin/AblyEventStreamHandler.java
QuintinWillison: @tiholic [as above](https://github.com/ably/ably-flutter/pull/6#discussion_r425642598) I've commented on that commit.

android/src/main/java/io/ably/flutter/plugin/AblyEventStreamHandler.java
tiholic: @paddybyers after much thought and discussion with @QuintinWillison here is what I came up with:

### RTE6a is supported with current implementation already
But there is a caveat:

These are possibilities represented by RTE6a:
1. registering a listener (say, `ξ` - Xi) inside another (say, `β`)
   - Number of `ξ` listeners will increase for each trigger of `β` listener.
   - Number of times the `first ξ` listener is triggered will be 1 less than number of times `β` listener is triggered.
   - Similarly, `second ξ` listener will be triggered 2 less than number of times `β` listener is triggered, and so on.

```dart
//RETAINING LISTENER - β
realtime.connection.on().listen((ably.ConnectionStateChange stateChange) async {
    print('RETAINING LISTENER β :: Change event arrived!: ${stateChange.event}');
    //NESTED LISTENER - ξ
    realtime.connection.on().listen((ably.ConnectionStateChange stateChangeX) async {
        //k ξ listeners will be registered and each listener will be called `n-k` times respectively if listener β is called `n` times
        print('NESTED LISTENER ξ: ${stateChangeX.event}');
    });
});
```

2. cancelling a listener inside another. There are 2 ways to do this:
2.1 Registering a listener (say `pre ζ`) before registering the listener that will cancel it (say `γ`)
2.2 Registering a listener (say `post ζ`) after registering the listener that will cancel it (`γ`)
```dart
StreamSubscription preZetaSubscription;
StreamSubscription postZetaSubscription;
preZetaSubscription = realtime.connection.on().listen((ably.ConnectionStateChange stateChangeX) async {
  //This listener "pre ζ" will be cancelled from γ
  print('NESTED LISTENER "pre ζ": ${stateChangeX.event}');
});

//RETAINING LISTENER - γ
realtime.connection.on().listen((ably.ConnectionStateChange stateChange) async {
  print('RETAINING LISTENER γ :: Change event arrived!: ${stateChange.event}');
  if (stateChange.event == ably.ConnectionEvent.connected) {
    //by the time this cancel is triggered, preZeta will already have received current event.
    await preZetaSubscription.cancel();
    //by the time this cancel is triggered, postZeta hasn't received the event yet. And will never receive as it is cancelled.
    await postZetaSubscription.cancel(); 
  }
});

postZetaSubscription = realtime.connection.on().listen((ably.ConnectionStateChange stateChangeX) async {
  //This listener "post ζ" will be cancelled from γ
  print('NESTED LISTENER "post ζ": ${stateChangeX.event}');
});
```

RTE6a succeeds with 1 and 2.1 but **fails with 2.2** even though platform side events are triggered perfectly in all cases (Confirmed this by adding temporary logs on platform side triggers)


### Why is 2.2 failing?

That is how streams and subscriptions are designed. As soon as a subscription is closed, any events triggered further will not be delivered to the listener.

### Why can we not have pure dart implementation to have a n -> 1 mapping?

As we are proceeding with idiomatic approach, we are entitled to use streams. And the failure of 2.2 is due to how streams and subscriptions behave by design.


### We still need to support 2.2. What are the alternate approaches?
1. Extend the default stream and subscription implementation and override the subscription cancel operation to delayed cancellation, and implement the `1 <-> n` approach from `platform <-> dart`. This is just an idea without any POC backing, and is not recommended as dart developers will be expecting the streams to be closed when they close it.

2. Drop streams and follow callback approach.

-----

PS: An example has be added to illustrate this behavior, and a test is written to show that this is a stable behavior as requested by @QuintinWillison 
https://github.com/ably/ably-flutter/commit/cb8090b4e44f6c4681d1ffd5b94e09de3950c8e5

Sorry about blog style comment, couldn't write in simple terms about this behavior 😉

tiholic:
@mattheworiordan @QuintinWillison 

Updated [README](https://github.com/ably/ably-flutter/tree/feature/realtime-pilot) with "how to run example" and "API usage"

9f9a3f8df536495a6a9607eee2396c6d58e89abf

android/src/main/java/io/ably/flutter/plugin/AblyEventStreamHandler.java
mattheworiordan: My immediate thoughts.

1. I am not seeing the advantage of using Dart Streams for connection/channel events, at least from examples seen so far and any ideas I can think of in practice:

```dart
realtime.connection.on((ably.ConnectionStateChange stateChange) async {
    // handle event
});
```

I am not sure how in practice that is better than just sticking with Ably API style, in fact I would say it's more verbose and logically wrong (`on` what?, `on()` does not make sense if you had to read this code, it should be at best `realtime.connection.listen(...)`, but then why?

This still makes sense regardless of what language we use:

```dart
realtime.connection.on(ably.ConnectionStateChange stateChange) {
    // handle event
});
```

In what situations would a customer want to process a stream of Connection of Channel events as a Stream of data?  I get why Dart Streams are idiomatic, but they are designed for streams of data / events that need processing, and support iterable functional styles.  I don't think of any practical ways that would benefit devs.

On your `n-k` listener problem, this is not a Dart problem, it's a problem with your code example.

In Dart:

```dart
/* registers a new listener */
realtime.connection.on().listen((ably.ConnectionStateChange stateChange) async {
    /* registers a new listener on EVERY state change */
    realtime.connection.on().listen((ably.ConnectionStateChange stateChangeX) async {
        /* handle event */
    });
});
```

In Javascript:

```javascript
/* registers a new listener */
realtime.connection.on((stateChange) => {
    /* registers a new listener on EVERY state change */
    realtime.connection.on((ably.ConnectionStateChange stateChangeX) => {
       /* handle event */
    });
});
```

So the problem is not to do with Dart, the problem is that you're registering a new handler on every state change, and this is a caller bug in any platform we support. So at best this should be something more logically like:

```javascript
/* registers a new listener */
realtime.connection.on('connect', (stateChange) => {
    /* registers a new listener on a specific state change with ONCE */
    realtime.connection.once('disconnected', (ably.ConnectionStateChange stateChangeX) => {
       /* handle disconnected event once */
    });
});
```

The only place I can genuinely see idiomatic Dart streams being useful would be with channel message subscriptions, and even if we did do that, we need to think about how it will be human readable.  So any of the following feels better:

```dart
/* Make it explicit when we're giving the user a stream vs a standard Ably IDL */
final stream = channel.subscribeStream;
final subscription = stream.listen((msg) => print('Data: $msg.data'));

/* But would it possible for subscribe to return a stream, and also supports an optional callback */
final eventStream = channel.subscribe('eventName');
final subscription = eventStream.map((msg) => (msg.data)).listen((data) => print('Data: $data'));
final allStream = channel.subscribe();

/* But keep Ably IDL support too */
channel.subscribe('eventName', (msg) => print('Data: $msg.data'));
```

Either way, I still think @paddybyers and @QuintinWillison are right that it would be easier to implement an Ably `EventEmitter` in Dart, which only subscribes once to whatever underlying Ably object it maps to, and thus has only one subscription or no subscription, and that's it.  And the Dart Ably `EventEmitter` could in turn implement Stream support by returning a [Dart broadcast stream](https://api.dart.dev/stable/2.8.2/dart-async/Stream-class.html), which I assume would then listen for `cancel` events on the `SteamSubscription` and detect when none exist and relay the `off` event to the Ably EventEmitter.

I suspect all of this would be easier to review if a PR was done specifically to bring EventEmitter capabilities to Dart in a generic way, that has a set of unit tests around it, and a proxy `EventEmitter` in iOS and Android that the tests would run against.

pubspec.yaml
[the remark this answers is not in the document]
tiholic: @QuintinWillison some commentary was already added on why older package is required https://github.com/ably/ably-flutter/pull/6/files#diff-fb2d52cd953afedcafc9fe7629c6f132R18

android/src/main/java/io/ably/flutter/plugin/AblyEventStreamHandler.java
QuintinWillison: I think, as @paddybyers indicated from the earliest days of development on a Flutter solution, that this really does cut to the core of what we're trying to do here. For what it's worth, @mattheworiordan, I was of the opinion that @tiholic should try to make the approach he was pursuing work but I think I'm now swayed to Paddy's approach based on:

1. The fact that you and Paddy agree on it.
2. The complexity described by Rohit in terms of nested listeners which is still blowing my mind (I don't fully understand it and don't really have the time right now to explore it in depth).

What is important is that we need to keep this simple and consistent with other Ably implementations, especially those on top of which this plugin is built. It's not worth trying to shoehorn Dart-idiomatic constructs on top of our proven fundamentals if we just end up bringing caveats and additional complexities that'll need debugging 'in the field' later.

README.md
```diff
@@ -33,3 +33,83 @@ Your pull requests are welcome but please keep them manageable and focussed.
...
+* Clone the repo
+* cd to `example` folder
+* run `flutter packages get` to install dependencies
+* `flutter run` will start the application on connected android / iOS device
+
+
+## Using the Realtime API
+
+##### Import the package
+```dart
```
QuintinWillison, line 49: As you have lower down in this commit and aligning with the format of the file already, can you leave a line of vertical white space around headings please. While it may not change how it renders, it does affect readability of the source.

README.md
```diff
@@ -33,3 +33,83 @@ Your pull requests are welcome but please keep them manageable and focussed.
...
+```
+
+##### Rest API
+
+```dart
+ably.Rest rest = await ablyPlugin.createRest(options: clientOptions);
+```
+
+Getting a channel instance
+```dart
```
QuintinWillison, line 72: A line of vertical whitespace above blocks of example code is a good idea too, please.

README.md
```diff
@@ -33,3 +33,83 @@ Your pull requests are welcome but please keep them manageable and focussed.
 Equally your input on any pull requests we have in flight at any given time is invaluable to us, so please get involved.
 
 Thanks! :grin:
+
```
QuintinWillison, line 36: More than one line of vertical whitespace is never needed so please remove this additional line and the one below.

Also, can you insert this entire addition above 'Contributing' as is more conventional. As in we're helping users get up and running with the codebase before we encourage them to contribute!

README.md
```diff
@@ -33,3 +33,83 @@ Your pull requests are welcome but please keep them manageable and focussed.
...
+##### Create an Ably instance
+```dart
+final ably.Ably ablyPlugin = ably.Ably();
+```
+
+##### create a ClientOptions
+```dart
+final clientOptions = ably.ClientOptions.fromKey("<YOUR APP KEY>");
+clientOptions.environment = 'sandbox';
```
mattheworiordan, line 61: Please remove `sandbox`.  This is only intended for client SDK developers, no customer would use `sandbox` as they cannot create API keys in sandbox

README.md
```diff
@@ -33,3 +33,83 @@ Your pull requests are welcome but please keep them manageable and focussed.
...
+## Using the Realtime API
+
+##### Import the package
+```dart
+import 'package:ably_flutter_plugin/ably.dart' as ably;
+```
+
+##### Create an Ably instance
+```dart
+final ably.Ably ablyPlugin = ably.Ably();
```
mattheworiordan, line 55: This does not match Ably's conventions in pretty much every other SDK. I don't follow why the user needs to define an instance of an Ably plugin at all, and deviate so significantly from other SDKs.

What this should look like is:

```
import 'package:ably_flutter_plugin/ably.dart' as Ably;
final Ably.Rest ably = Ably.Rest(key: 'key');
Ably.Rest.Channel channel = ably.channels.get('test');
channel.publish(name: 'event');
```

README.md
```diff
@@ -33,3 +33,83 @@ Your pull requests are welcome but please keep them manageable and focussed.
...
+//passing just data
+await channel.publish(data: "Ably");
+//publishing an empty message
+await channel.publish();
+```
+
+##### Realtime API
+
+```dart
+ably.Realtime realtime = await ablyPlugin.createRealtime(options: clientOptions);
```
mattheworiordan, line 92: Why does `createRealtime` and `createRest` need to be async?

tiholic: @mattheworiordan As they require to contact platform methods via platform channel which is an async operation.

README.md
tiholic: I think we should be able to do that, but this requires some changes that are very critical to the complete plugin based approach.

Any comments from @QuintinWillison on this would be helpful!

README.md
tiholic: Done@https://github.com/ably/ably-flutter/commit/083a0a518b96c8e773f98c8353af420b698beaa3

README.md
tiholic: Done@https://github.com/ably/ably-flutter/commit/083a0a518b96c8e773f98c8353af420b698beaa3

README.md
tiholic: Done@https://github.com/ably/ably-flutter/commit/083a0a518b96c8e773f98c8353af420b698beaa3

README.md
tiholic: Done@https://github.com/ably/ably-flutter/commit/d46b88ed11b34b958d0a3483d9001773ef2fb7dd

README.md
mattheworiordan: > I think we should be able to do that, but this requires some changes that are very critical to the complete plugin based approach.

Well the default position should be adherence to our API, unless there is a compelling reason to deviate.  Is there one?

README.md
mattheworiordan: > As they require to contact platform methods via platform channel which is an async operation.

But is that necessary?  
Specifically, if Realtime or REST asynchronously creates an instance of Realtime or REST with iOS or Android, and another method is called immediately afterwards, wouldn't the Dart instance exist synchronously and thus be ready to proxy requests anyway?  If not, does this introduce much complexity to fix?

tiholic: Dart instance will be created only after platform instance is created.

Creating a Rest/Realtime instance on dart side will require a `long` identifier, which will be sent back to dart to store against the dart instance.


#### Here is how it is working now:

Any methods called on dart side will be triggered on the platform side instance  - this instance will be queried using the `long` identifier sent from dart side for each method call.

```dart
createRest ==(async)==> restPlatform(id: 1) ==(async)=> restDart(id: 1)
//await for the above
restInstance = restDart(id: 1)
restDart(id: 1).someMethod(arguments) ==(async)==> restPlatform(id: 1).someMethod(decodedArguments) ==(async)==> responseFromPlatformSide
```

If we don't have that `long` identifier(`id: 1` above) available, we wouldn't know on which platform instance respective method should be invoked.


#### If we created a dart side instance first here is how it might work:

We will need to keep a tab on any methods called immediately after dart side instance is created and trigger platform method call after acquiring the `long` identifier from platforms end.

```dart
createRest ==(async)==> restPlatform(id: 1) ==(async)=> restDart(id: 1)
//don't await for the above
restInstance = restDart(id: null)
restDart.someMethod(arguments) ==(wait until id is acquired)==x
restDart.id = 1;
//then trigger
restDart.someMethod(arguments).__internal__delayed__methodcall__handler.call() ==> restPlatform(id: 1).someMethod(decodedArguments)
```

IMO, this make the dart side code trickier to develop and maintain.

tiholic: Even this should be doable.

tiholic: @mattheworiordan I'd prefer to try and this handled in a separate PR as current PR is purely focused on proving Event handling

@QuintinWillison @paddybyers

README.md
tiholic: @mattheworiordan There is no reason, just that I'm reusing the same codebase developed by Quicktin in early days and parked the thought of optimizing it for later. However, I'd prefer to try and this handled in a separate PR as current PR is purely focused on proving Event handling.

cc: @QuintinWillison @paddybyers

README.md
QuintinWillison: I would go further and remove the `environment` line entirely, certainly for high-level documentation like this which is aimed at users (i.e. Flutter app developers).

README.md
QuintinWillison: Thanks @tiholic - can you please create an issue to do that work under, bidirectionally link it and this conversation and then resolve this conversation.

README.md
QuintinWillison: Moved to handle under [related thread](https://github.com/ably/ably-flutter/pull/6#discussion_r441465807).

README.md
tiholic: Removed `environment` from README c777214e920237f450d5e27b80503becd6bb641e

README.md
tiholic: Issue logged https://github.com/ably/ably-flutter/issues/15

android/src/main/java/io/ably/flutter/plugin/AblyEventStreamHandler.java
tiholic: @QuintinWillison @paddybyers @mattheworiordan 

This conversation is still open for a decision to be made. We have the following choices:

### Using Streams

1. Current PR is good to be merged if we want to document and let go of that one particular failure case of RTE6a and document usage of the hack mentioned above.
2. Implement `EventEmitter` to link `1-n` from platform, which essentially requires for us to write some extra code and provide interface to end users with Streams. (there is no point in doing this IMO, especially when looking at this from SDK Consumer's perspective.)


> Note: RTE6a drawback is due to usage of Dart idiomatic Streams. Refer to 2.2 in my comment [above](https://github.com/ably/ably-flutter/pull/6#discussion_r426137416).

However, there is a hack that users can do in order to make it work. `postZetaSubscription` can be cancelled like this: `Future.delay(Duration.zero, postZetaSubscription.cancel())` which is similar to `setTimeout(fn, 0)` in javascript

### Not using streams
1. Drop streams and implement `EventEmitter` with function callbacks. This can be designed to support all scenarios (including RTE6a)

QuintinWillison: Thanks @tiholic ... I really appreciate the clarity with which you're communicating this. It really helps.

As we've agreed this morning, we are going to proceed with "Using Streams" and you are going to ensure that this is adequately documented in terms of - in particular - divergence from alignment with [RTE6a](https://docs.ably.io/client-lib-development-guide/features/#RTE6a). Your Point 1 above.

I think we had discussed creating an implementation that conforms, in terms of interface, with `EventEmitter`, but I'm not convinced that this is required for users of this library to get the work they need to get done. Can you please create an issue linking back to this thread so we can discuss layering this in under a separate PR later on if the arguments are strong enough that `EventEmitter` is _required_ for this SDK. Your Point 2 above. I trust your opinion here.

tiholic: Issue #16 logged

tiholic: Updated README with necessary details 1d0f376e9828c9d789912fbc74715fc6e6dba0e7

QuintinWillison: Thanks. I could be picky and stall things further by identifying some minor improvements that could be made to the read me, but we also need to get this PR merged so I think we can always just iterate on that further down the line. The key thing is that we've got this front and centre so it should not disappear out of view (fingers crossed).

Issue #12, opened 2020-05-07
map/dict based approach to Encoding and Decoding channel data


tiholic:
To overcome the brittleness of ordered encoding and decoding from iOS, Android and Flutter side, we must follow KVP based serialization and deserialization approach.

Related discussions:
https://github.com/ably/ably-flutter/pull/5#discussion_r419159554
https://github.com/ably/ably-flutter/pull/5#discussion_r419159258
https://github.com/ably/ably-flutter/pull/6#discussion_r420359338

paddybyers:
> we must follow KVP based serialization and deserialization approach

It's not mandatory to use an encoding in which key names are explicit (ie json or msgpack) if we can avoid the brittleness. For example, if we can have a single file that defines the schema (ie the field names and the types for each member of each class), and we can have a way to ensure that all 3 parts (dart, ios, android) implement the same schema then that's ok.

QuintinWillison:
Thanks @paddybyers - that's an interesting approach which would also reduce the amount of code we have to write and transliterate - reducing the "triple-handing" when it comes to serialisation and deserialisation per value.

The last time I played with protocol buffers it always felt a bit turbulent, but then that may have just been the process of building and running the right version of the `protoc` tooling. That may have improved in the last few years.

Regardless, this approach could work with any schema-based code generator. What other options might there be when it comes to that tooling? Or might we consider writing something ourselves?

@tiholic, what do you think? Have you worked with code generation much, yourself? I haven't particularly but it's an area that does interest me.

tiholic:
@QuintinWillison I have worked on json based code generation with a custom generator for java. I feel we can try protoc. I've been exploring on that.

QuintinWillison:
Great. Thanks! Looking forward to seeing what you come up with. 😀

tiholic:
We can close this as it is already implemented. We have #28 for switching codec to a more robust solution.

PR #13, opened 2020-05-23
re-codec using JSON and Codegen

tiholic:
### JSON based encoding and decoding.

Encoders and Decoders will now return `JSON` from `Ably type` and return `Ably type` from `JSON`
Addressing: https://github.com/ably/ably-flutter/issues/12

### Codegen for constants that need to be in Sync on platform side and flutter side

* Generate codec types (Addressing: https://github.com/ably/ably-flutter/pull/5#discussion_r419159258)
* Generate method names for `MethodChannel` and `EventChannel`
* Generate Properties of all transferable objects (Ex: `ClientOptions.authUrl`, `ClientOptions.authMethod`, etc)

**PS**
protobuf based codec didn't turnout to be a good approach as per [trails](https://github.com/ably/ably-flutter/tree/feature/protobuf-codec) and [discussions](https://ably-real-time.slack.com/archives/C030C5YLY/p1589891505276600).

And other approaches like flatbuffers don't have direct support for Objective-C, but support Swift which would require us to port iOS platform code to Swift.


ios/Classes/codec/AblyFlutterReader.m
```diff
@@ -54,87 +66,102 @@ use an explicitly received null from Dart (manifesting this side as a nil id)
...
+    READ_VALUE(o, queueMessages, jsonDict, TxClientOptions.queueMessages);
+    READ_VALUE(o, echoMessages, jsonDict, TxClientOptions.echoMessages);
+    READ_VALUE(o, recover, jsonDict, TxClientOptions.recover);
+    READ_VALUE(o, environment, jsonDict, TxClientOptions.environment);
+    READ_VALUE(o, idempotentRestPublishing, jsonDict, TxClientOptions.idempotentRestPublishing);
+    READ_VALUE(o, fallbackHosts, jsonDict, TxClientOptions.fallbackHosts);
+    READ_VALUE(o, fallbackHostsUseDefault, jsonDict, TxClientOptions.fallbackHostsUseDefault);
+    ON_VALUE(^(const id value) { o.defaultTokenParams = [AblyFlutterReader readTokenParams: value]; }, jsonDict, TxClientOptions.defaultTokenParams);
+    READ_VALUE(o, defaultTokenParams, jsonDict, TxClientOptions.defaultTokenParams);
+    //Following properties not supported by Objective C library
```
QuintinWillison, line 117: I think I would prefer that the message conveyed in this comment and the commented out lines below it (even though they would never compile in their current form given the refactor!) was moved somewhere a bit more prominent. Perhaps the developer notes for now? Or an issue - perhaps with an 'investigate' or 'question' label? I like that you want to keep a record of this but it feels lost here.

This also applies to the writer so I've not separately commented there (i.e. `encodeErrorInfo:` method therein).

ios/Classes/codec/AblyFlutterWriter.m
```diff
@@ -3,59 +3,69 @@
...
+#import "AblyPlatformConstants.h"
 
 
 @implementation AblyFlutterWriter
 
 - (void)writeValue:(id)value {
     if([value isKindOfClass:[ARTErrorInfo class]]){
-        [self writeErrorInfo: value];
+        [self writeByte:errorInfoCodecType];
+        [self writeValue: [self encodeErrorInfo: value]];
```
QuintinWillison, line 14: General observation but there is some inconsistency across your commits in respect of spaces after colons in message sending statements. I would prefer no spaces. So for this line, simply:

```suggestion
        [self writeValue:[self encodeErrorInfo: value]];
```

lib/src/codec.dart
```diff
@@ -96,163 +73,222 @@ class Codec extends StandardMessageCodec {
...
+    v.clientId = readFromJson<String>(jsonMap, TxTokenParams.clientId);
+    v.nonce = readFromJson<String>(jsonMap, TxTokenParams.nonce);
+    v.timestamp = readFromJson<DateTime>(jsonMap, TxTokenParams.timestamp);
+    v.ttl = readFromJson<int>(jsonMap, TxTokenParams.ttl);
+    return v;
   }
 
-  AblyMessage decodeAblyMessage(ReadBuffer buffer){
+  AblyMessage decodeAblyMessage(Map<String, dynamic> jsonMap){
+    if(jsonMap==null) return null;
```
QuintinWillison, line 245: My gut feeling without looking at how this is called more deeply is that this feels like a null argument exception... or do you document this behaviour somewhere?

bin/codegencontext.dart
```diff
@@ -0,0 +1,152 @@
...
+  // Custom type values must be over 127. At the time of writing the standard message
+  // codec encodes them as an unsigned byte which means the maximum type value is 255.
+  // If we get to the point of having more than that number of custom types (i.e. more
+  // than 128 [255 - 127]) then we can either roll our own codec from scratch or,
+  // perhaps easier, reserve custom type value 255 to indicate that it will be followed
+  // by a subtype value - perhaps of a wider type.
+  //
+  // https://api.flutter.dev/flutter/services/StandardMessageCodec/writeValue.html
+
+  //Ably flutter plugin protocol message
```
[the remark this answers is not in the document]
tiholic: Addressed in 5f5c60d9c1fcb292d83f5a892fa628529b55328e

lib/src/codec.dart
QuintinWillison: @tiholic ?

ios/Classes/codec/AblyFlutterReader.m
QuintinWillison: @tiholic ?

(oh, and also, please use proper form "Objective-C", never "Objective C" or - in my opinion worse - "Obj-C" 🤢 )

bin/README.md
```diff
@@ -0,0 +1,17 @@
+## Code generation to keep platform constants in sync
```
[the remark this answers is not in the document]
QuintinWillison: @tiholic have you addressed this one? Thanks.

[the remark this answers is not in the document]
tiholic: Yes. I wasn't able to add a review comment earlier. This has been resolved.

ios/Classes/codec/AblyFlutterReader.m
tiholic: Yes. I wasn't able to add a review comment earlier. This has been resolved.

lib/src/codec.dart
tiholic: @QuintinWillison updated documentation for codec.dart - 237315e9a5b5a65bdc7abe091f4611bd4ef32c77

tiholic:
@QuintinWillison renamed `_CodecEncoder` and `_CodecDecoder`, bit I'd prefer leaving the class name `_CodecPair` as is because the class extending `StandardMessageCodec` is named as `Codec` ([ref](https://github.com/ably/ably-flutter/blob/feature/codec-upgrade/lib/src/codec.dart#L45))

https://github.com/ably/ably-flutter/commit/2a49b9693e08ee53aa8602f213535a0b6bcbc169

PR #18, opened 2020-06-27
Synchronous API for creating instances and Hot-reload fix

tiholic:
This PR mainly focuses on creating Ably instances in a synchronous manner.
The following changes were made possible as all other operations on each ably instances is an asynchronous operation.

## If not async? then, Sync + Lazy = Slazy!!! 😉 
1. All platform calls from dart side are made lazy
2. Singleton Ably library instances on Android side (iOS will be caught up)

A peek on new APIs
```dart
import 'package:ably_flutter_plugin/ably.dart' as ably;

final clientOptions = ably.ClientOptions.fromKey("<YOUR APP KEY>");
clientOptions.logLevel = ably.LogLevel.verbose; //optional
ably.Rest rest = ably.Ably.Rest(options: clientOptions);

function testRest(){
  ably.RestChannel channel = rest.channels.get('test');
  await channel.publish(name: "Hello", data: "Ably");
}

function testRealtime(){
  ably.Realtime realtime = ably.Ably.Realtime(options: clientOptions);
    realtime.connection.on().listen((ably.ConnectionStateChange stateChange) async {
    print('Realtime connection state changed: ${stateChange.event}');
  });
}

```

## Hot-reload issue fix 🌶️  ♻️  🎉  🎉  
3. `registerAbly` is now used to dispose off any registered platform instances
4. Android `StreamsChannel` source is replicated and necessary changes are made to dispose all listeners on hot-reload (Will be removing the streams_channel plugin dependency eventually)

> NOTE: Draft PR supporting Android only.

@QuintinWillison @mattheworiordan @paddybyers Please share your feedback on the API. Will make necessary changes and proceed with implementing iOS platform related changes.


lib/src/ably_implementation.dart
```diff
@@ -1,19 +1,20 @@
 import 'dart:async';
 
-import 'package:ably_flutter_plugin/src/impl/message.dart';
 import 'package:ably_flutter_plugin/src/interface.dart';
+import 'package:flutter/material.dart';
```
QuintinWillison, line 4: Is this an unused import?

lib/src/ably_implementation.dart
```diff
@@ -24,79 +25,75 @@ class AblyImplementation implements Ably {
...
+  Future _initialize() async {
+    if(_initialized) return;
+    _init();
+    // if `_initialized` is false => initialization is in progress.
+    // Let's wait for 250ms and retry
+    //
+    // this is required as many asynchronous `_initialize` calls
+    // will be invoked from different Ably instances
+    while(true){
+      await Future.delayed(Duration(milliseconds: 250));
```
QuintinWillison, line 59: Another seemingly arbitrary delay. Quite a long time, in CPU terms.

android/src/main/java/io/ably/flutter/plugin/StreamsChannel.java
```diff
@@ -0,0 +1,205 @@
...
+            final String method = methodParts[0];
+            switch (method) {
+                case "listen":
+                    onListen(id, call.arguments, reply);
+                    break;
+                case "cancel":
+                    onCancel(id, call.arguments, reply);
+                    break;
+                default:
+                    reply.reply(null);
```
QuintinWillison, line 103: I recognise this is code imported from elsewhere but this smells fishy... should that not be an exception thrown?

android/src/main/java/io/ably/flutter/plugin/AblyEventStreamHandler.java
```diff
@@ -90,15 +96,12 @@ public void onCancel(Object object) {
...
-            switch (eventName) {
-                case PlatformConstants.PlatformMethod.onRealtimeConnectionStateChanged:
-                    ablyLibrary.getRealtime(message.handle).connection.off(connectionStateListener);
-                case PlatformConstants.PlatformMethod.onRealtimeChannelStateChanged:
-                    // ablyLibrary.getRealtime(handle).connection.off(connectionStateListener);
-            }
-        });
+        AblyFlutterMessage<String> message = getMessage(object);
+        String eventName = message.message;
+        switch (eventName) {
```
QuintinWillison, line 101: With a single `case` perhaps this could be an `if` now. 😁 

Regardless, is it the intentional that this method is effectively a no-op if the `eventName` isn't exactly what is expected? Could we not consider failing here to indicate unexpected inbound?

android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java
```diff
@@ -117,26 +105,10 @@ public void onMethodCall(@NonNull MethodCall call, @NonNull MethodChannel.Result
...
-            _ably.dispose();
-        }
-
-        System.out.println("Creating new Ably instance (# " + handle + ").");
-        _ably = new AblyLibrary();
-        _ablyHandle = handle;
-
-        result.success(handle);
+        System.out.println("Registering library instance to clean up any existing instnaces");
+        onHotRestartListener.trigger();
```
QuintinWillison, line 109: Where is this implemented?... I can't seem to work out how this works.

Maybe my `grep` skills are lacking?

```
ably-flutter % git grep trigger
README.md:As per RTE6a, 2nd listener should also be triggered, but it will not be as 2nd listener is registered after 1st listener and  stream subscription is cancelled immediately after first listener is triggered.
android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java:        void trigger();
android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java:        onHotRestartListener.trigger();
example/lib/main.dart:          await preZetaSubscription.cancel();  //by the time this cancel is triggered, preZeta will already have received current event.
example/lib/main.dart:          await postZetaSubscription.cancel(); //by the time this cancel is triggered, postZeta hasn't received the event yet. And will never receive as it is cancelled.
example/test/widget_test.dart:    // Build our app and trigger a frame.
ably-flutter % git grep OnHotRestart
android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java:    public interface OnHotRestart {
android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java:    private OnHotRestart onHotRestartListener;
android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java:    static synchronized AblyMethodCallHandler getInstance(OnHotRestart listener) {
android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java:    private AblyMethodCallHandler(OnHotRestart listener) {
```

And, also, the log statement has a typo - "instances".

README.md
```diff
@@ -57,7 +55,7 @@ clientOptions.logLevel = ably.LogLevel.verbose; //optional
 ##### Rest API
 
 ```dart
-ably.Rest rest = await ablyPlugin.createRest(options: clientOptions);
+ably.Rest rest = ably.Ably.Rest(options: clientOptions);
```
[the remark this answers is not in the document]
mattheworiordan: @Quintin, out of interest, why did you propose introducing `RestClient` instead of:

```dart
var rest = Ably.Rest(options: clientOptions);
```

See comparable https://github.com/ably/ably-js#typescript, https://github.com/ably/ably-js#introduction and https://github.com/ably/ably-java#using-the-realtime-api

P.S. I am not sure why we opt for `ably` as the lib vs `Ably`.  Is that a Dart thing?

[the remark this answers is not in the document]
QuintinWillison: @mattheworiordan A fair observation regarding naming. I guess I was aligning with what we have in Java. Happy to drop the `Client` suffix for both this and `Realtime`. @tiholic, will that work?

In terms of the `ably` package, that is indeed [a Dart thing](https://dart.dev/tools/pub/pubspec#name).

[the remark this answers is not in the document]
tiholic: My concern in writing the README/example that way was to avoid multiple imports for ably library.

1. if we `import 'package:ably_flutter_plugin/ably.dart' as ably;` then `ClientOptions` and other classes/interfaces are available for usage directly as `ably.ClientOptions` along with having access to the library encapsulation class `ably.Ably`.

2. We can `import 'package:ably_flutter_plugin/ably.dart'`, but, this would import all exports from Ably and `ConnectionState` class conflicts ❌  with default flutter classes. So the only way to import is using an alias like in point 1 above.


My suggestion would be to get rid of the `Ably` encapsulation and directly consume the API using `ably.Rest(...)` and `ably.Realtime(...)`.
> Note: the ☝️ lower case `ably` it comes from import alias as in point 1

But, this will have some conflicts with [Ably's flutter spec](https://github.com/ably/ably-flutter/blob/feature/synchronous-specification/lib/src/spec/)

To make this happen, we can rename [abstract Rest definition](https://github.com/ably/ably-flutter/blob/feature/synchronous-specification/lib/src/spec/rest/rest.dart) and [abstract Realtime definition](https://github.com/ably/ably-flutter/blob/feature/synchronous-specification/lib/src/spec/realtime/realtime.dart) from spec to `AblyRest` and `AblyRealtime` -or- `AbstractRest` and `AbstractRealtime`. And similarly rename any other definitions from spec whenever required.


PS: `xxxClient` is not usde anywhere in dart code, so we are cool I suppose.

test/ably_flutter_plugin_test.dart
```diff
@@ -50,9 +44,9 @@ void main() {
         case PlatformMethod.connectRealtime:
         default:
           return null;
-        //  eventsOff,
-        //  eventsOn,
-        //  eventOnce,
+      //  eventsOff,
```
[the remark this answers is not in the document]
tiholic: Nothing 😅 
They are removed 2c26aca9a1cb9e5a8ab67ebd0197c358474fbda3

lib/src/impl/platform_object.dart
```diff
@@ -8,18 +11,34 @@ import 'package:streams_channel/streams_channel.dart';
...
 
-  get handle => _handle;
-  get ablyHandle => _ablyHandle;
+  Future<int> createPlatformInstance();
+
+  bool _registering = false;
+  Future<int> get handle async {
+    if(_registering){
+      //if handle is queried while already being fetched from remote, it must be put on hold..
+      await Future.delayed(Duration(milliseconds: 250));
```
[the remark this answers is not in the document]
tiholic: @QuintinWillison improved the commentary fb7fa6459ff0dbafff2b180ce6474badc5925ed9
If this is still not clear, I can try giving it another shot.

lib/src/ably_implementation.dart
tiholic: @QuintinWillison what could be the optimistic wait time? This has to do a round-trip - a platform method call

android/src/main/java/io/ably/flutter/plugin/StreamsChannel.java
tiholic: @QuintinWillison there is no way of having another method type other than `"listen"` and `"cancel"` here. This code is borrowed from StreamsChannel with minor changes. Would you like to have this completely refactored now? or we can have a issue about this to track and deal with this later?

android/src/main/java/io/ably/flutter/plugin/AblyEventStreamHandler.java
tiholic: @QuintinWillison Yes, it can be an `if`, but soon we will have to switch back to `switch`.
And yes, it can be left as a no-op as this `cancel` event is triggered only when we are cancelling stream subscription. Necessary has been handled on the `onListen` method

android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java
tiholic: @QuintinWillison your grep skills are at 💯  
Just that we are using the not-so-obvious lambda expression (here, method reference, actually) [in AblyFlutterPlugin.java](https://github.com/ably/ably-flutter/blob/feature/synchronous-specification/android/src/main/java/io/ably/flutter/plugin/AblyFlutterPlugin.java#L41) 😉

lib/src/ably_implementation.dart
tiholic: Yes, thanks. Removed da5de0f49b0b5f6f6aa7f3ea269a088e0c33be1a

README.md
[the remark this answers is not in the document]
QuintinWillison: Quite a journey there but it seems like you're getting to what @mattheworiordan was looking for, at least in respect of how the API looks in use.

In terms of those renames you refer to in your penultimate paragraph, I would prefer either `XxxBase` or `XxxInterface` for those abstract class ('interface') names rather than either `AblyXxx` or `AbstractXxx` - i.e. `RestBase` and `RealtimeBase` - assuming, of course, that they need to be presented as abstract base classes.

I believe that will yield a hierarchy looking like:

- `ably.Rest` (top-level function) returning implementation of `RestBase` (abstract class: 'interface') which extends `AblyBase`
- `ably.Realtime` (top-level function) returning implementation of `RealtimeBase` (abstract class: 'interface') which extends `AblyBase`

Alternatively `RestBase` could be named `RestInterface`, and `RealtimeBase` could be named `RealtimeInterface`. I don't have strong opinions either way so would be happy to go with what you feel is most representative of their purpose, @tiholic. Presumably they appear as part of the public API for this client library, right?

Though please do correct me if I'm wrong as my Dart still isn't very strong. We're all here to learn! 😁 

As a side-note, have you had anything that's been a contender for using [mixins](https://resocoder.com/2019/07/21/mixins-in-dart-understand-dart-flutter-fundamentals-tutorial/)?

lib/src/impl/platform_object.dart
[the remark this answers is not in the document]
QuintinWillison: Thanks. Had a look at that.

If you re-read your commentary back I think there's a typo or too (e.g. 'server' > 'served').

Also, 250ms (4Hz) is relatively glacial when most apps should be rendering at a minimum of 60Hz (17ms frame to frame). I would suggest perhaps a delay of no more than 10ms would be more appropriate here. Apart from anything else that is more likely to drive out issues early on in testing too.

And, finally, should there not be a loop here checking `_registering` each time before assuming it's ok to proceed or do I misunderstand the intended outcome? And, if you do end up having a check loop, then there'll need to be some kind of escape route - e.g. timeout - in case the flag never clears. In fact your comment says "a short hold and re-check" but I'm not sure I see evidence of that in the code.

lib/src/ably_implementation.dart
QuintinWillison: Thanks.

I think I spotted that when I pulled this in to my IDE.

Are we not running a linter? I thought `pedantic` was installed in this project but maybe I forget how all that hooks in. I would prefer that we ran as much lint / analysis as we can on every build if possible.

lib/src/ably_implementation.dart
QuintinWillison: I think my [other response](https://github.com/ably/ably-flutter/pull/18#discussion_r450699980) equally applies here.

android/src/main/java/io/ably/flutter/plugin/StreamsChannel.java
QuintinWillison: Will it fail downstream when somebody unexpectedly receives `null` then?

Definitely no need to refactor now but I just want to make sure that this behaviour is documented and that somebody maintaining this months or years down the line will know that this is a downstream outcome.

Is this documented on the interface for `reply`'s `reply` method?

android/src/main/java/io/ably/flutter/plugin/AblyEventStreamHandler.java
QuintinWillison: ok

android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java
QuintinWillison: ok, thanks.

If there's anything you can do to either add commentary explaining that, or better provide some level of interim expression in code (e.g. interim typedef if that's a thing) then that might help. I'm thinking future maintainability - you, me, anybody else coming to this code in months or years time.

QuintinWillison:
Thanks for the responses to my review comments, @tiholic.

I think there are still a few improvements before we can merge this. Thanks.

lib/src/impl/platform_object.dart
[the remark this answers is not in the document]
QuintinWillison: "too" 🤦  😆

android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java
tiholic: Not sure if that is required! I feel the code is explanatory as you can see the usage of definition in below line form your logs...

```
android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java:    static synchronized AblyMethodCallHandler getInstance(OnHotRestart listener) {
```

README.md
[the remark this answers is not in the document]
tiholic: `xxxInterface` sounds like the right way to go as `xxxBase` has a different meaning when it comes to Java (`xxxBase` is used for java and `Ablyxxx` used for android implementation from what I can infer )

And yes, the abstract interfaces are a part of public API.

I've tried to implement `mixins` in this library at very early stages, with [PlatformObject](https://github.com/ably/ably-flutter/blob/feature/synchronous-specification/lib/src/impl/platform_object.dart). But, nope, it turned out they dont fit in here for us (yet, maybe).

android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java
QuintinWillison: ok, just bear in mind that maintainers of this codebase in the future will likely be short of time so we need to make things obvious wherever possible. If you feel this is obvious enough then leave it be. I just wanted to point out that I found this particular part of the event flow difficult to follow.

lib/src/impl/platform_object.dart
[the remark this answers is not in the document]
tiholic: "re-check" is happening via recursion. But as you pointed out, I feel it is appropriate to have a loop instead of recursion.
Also, I've added a timeout of 2 minutes with poll time reduced to 10ms from 250ms.

sha c792574efd37a3cff749aa066970262d0ff6dfec

lib/src/ably_implementation.dart
tiholic: Linter is not enabled in this I see. Should it be enabled via actions?

lib/src/ably_implementation.dart
tiholic: Fixed @ c792574efd37a3cff749aa066970262d0ff6dfec

android/src/main/java/io/ably/flutter/plugin/StreamsChannel.java
tiholic: I'll try by removing this and see while making other changes.

tiholic:
@QuintinWillison Yeah, this PR is still in draft state. I just getting some inputs on how the final API should look like.
Now that we have enough information to proceed further, I will finish iOS implementation and try to address all the unresolved comments from above.

README.md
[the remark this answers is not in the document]
mattheworiordan: Thanks for the feedback and thoughts, it's all getting exciting 👍

lib/src/impl/platform_object.dart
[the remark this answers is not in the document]
QuintinWillison: > Also, I've added a timeout of 2 minutes with poll time reduced to 10ms from 250ms.

Just checked your source... thankfully you've implemented "2 **seconds**". 😁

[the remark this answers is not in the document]
tiholic: Oops, my bad 😛

android/src/main/java/io/ably/flutter/plugin/StreamsChannel.java
tiholic: @QuintinWillison after further investigation into this, I've observed that StreamsChannel has re-used the code that was implemented in flutter's default EventChannel. And the `reply.reply()` in the default case is also copied from the same. So, I'd say let's leave it that way. WDYT?

tiholic: And can you point out the inconsistency on usage of final?

tiholic:
@QuintinWillison iOS side caught up, if the runway is clear for landing, we can merge this.

Regarding the `pubspec.lock` changes, that was probably due to automatic minor version upgrade behavior.
`^A.B.C` will upgrade packages if minor version releases are available. If we need strict version for dependencies, we can go with `package: 'A.B.C'` instead of `package: '^A.B.C'` (ref: https://dart.dev/tools/pub/dependencies)

Updated DeveloperNotes with instructions on static code analysis (includes all build time checks)

android/src/main/java/io/ably/flutter/plugin/StreamsChannel.java
QuintinWillison: ok, as [I said before](https://github.com/ably/ably-flutter/pull/18#discussion_r450702980) no need to refactor... however I'm not sure what you're referring to when [you ask](https://github.com/ably/ably-flutter/pull/18#discussion_r453219515):

> And can you point out the inconsistency on usage of final?

QuintinWillison:
Regarding [`pubspec.lock` changes](https://github.com/ably/ably-flutter/pull/18#issuecomment-657256973), I'm happy to leave things as they are for now and thank you for explaining. 😁 

There is perhaps a wider discussion to be had here at Ably, across all client libraries, around dependency version management / auto management.

ios/Classes/AblyFlutterPlugin.m
```diff
@@ -105,35 +96,43 @@ -(nullable AblyFlutter *)ablyWithHandle:(NSNumber *)handle;
...
+    // Initializing reader writer and method codecs
     FlutterStandardReaderWriter *const readerWriter = [AblyFlutterReaderWriter new];
     FlutterStandardMethodCodec *const methodCodec = [FlutterStandardMethodCodec codecWithReaderWriter:readerWriter];
     
-    FlutterMethodChannel *const channel =
-        [FlutterMethodChannel methodChannelWithName:@"io.ably.flutter.plugin" binaryMessenger:[registrar messenger] codec:methodCodec];
-    AblyFlutterPlugin *const plugin =
-        [[AblyFlutterPlugin alloc] initWithChannel:channel];
+    // initializing event channel for event listeners
+    AblyStreamsChannel *const streamsChannel = [AblyStreamsChannel streamsChannelWithName:@"io.ably.flutter.stream" binaryMessenger:registrar.messenger codec: methodCodec];
```
QuintinWillison, line 113: For some of these longer lines could you consider formatting them so they're a little less long. I'm definitely not a fan of strict 80/120/whatever line length limits but there's always a sensible level - whether viewed in the IDE or a GitHub PR - when it's worth breaking them down. For Objective-C I think this is particularly easy to do, often aligned around argument colons. e.g.:

```suggestion
    AblyStreamsChannel *const streamsChannel =
        [AblyStreamsChannel streamsChannelWithName:@"io.ably.flutter.stream"
                                   binaryMessenger:registrar.messenger
                                             codec: methodCodec];
```

ios/Classes/AblyStreamsChannel.h
```diff
@@ -0,0 +1,26 @@
+//  Copyright (c) 2018 Loup Inc.
```
QuintinWillison, line 1: Why do we have Loup's copyright at the top of here when we have renamed it to have an `Ably` prefix? If anything I would expect an attribution or acknowledgement, or that we should be using this in original form. What's going on here? WDYT, @paddybyers ?

ios/Classes/AblyStreamsChannel.m
```diff
@@ -0,0 +1,138 @@
...
+    } else {
+      [self->_messenger sendOnChannel:name message:[self->_codec encodeSuccessEnvelope:event]];
+    }
+  };
+  stream.handler = factory(call.arguments);
+  
+  [_streams setObject:stream forKey:key];
+  [_listenerArguments setObject:call.arguments forKey:key];
+  
+  FlutterError* error = [stream.handler onListenWithArguments:call.arguments eventSink:stream.sink];
```
QuintinWillison, line 112: This pointer should be immutable. It is very typical for local variable `error` pointers to be passed to methods using `&error` so this code is a trip hazard for maintainers. Make your intent clearer, please. i.e.:

```suggestion
  FlutterError const* error = [stream.handler onListenWithArguments:call.arguments eventSink:stream.sink];
```

applies elsewhere too. Please try to default to using `const` wherever possible. I think I've mentioned this before.

ios/Classes/AblyStreamsChannel.m
```diff
@@ -0,0 +1,138 @@
...
+  if (error) {
```
QuintinWillison, line 113: I notice this logic is used in a couple of places. Could you move this to the implementation of `encodeErrorEnvelope:` to decide? It could accept `nil` values for `error` and return a success envelope in that case. This this line would just become:

```suggestion
  callback([self->_codec encodeErrorEnvelope:error]);
```

With corresponding clear commentary on that interface to explain this behaviour.

ios/Classes/AblyStreamsChannel.m
```diff
@@ -0,0 +1,138 @@
...
+    callback([self->_codec encodeErrorEnvelope:error]);
+  } else {
+    callback([self->_codec encodeSuccessEnvelope:nil]);
+  }
+}
+  
+- (void)cancelForCall:(FlutterMethodCall*)call withKey:(NSNumber*)key usingCallback:(FlutterBinaryReply)callback andFactory:(NSObject<FlutterStreamHandler> *(^)(id))factory {
```
QuintinWillison, line 120: The various `self->` prefixes for ivar access are superfluous in this method implementation and just add noise. You only need explicit `self` when dealing with blocks.

ios/Classes/AblyFlutterMessage.h
```diff
@@ -7,8 +7,8 @@ NS_ASSUME_NONNULL_BEGIN
 +(instancetype)new NS_UNAVAILABLE;
 -(instancetype)init NS_UNAVAILABLE;
 
--(instancetype)initWithHandle:(NSNumber *)handle
-                      message:(id)message NS_DESIGNATED_INITIALIZER;
+-(instancetype)initWithMessage:(id)message
+                      handle:(NSNumber *)handle NS_DESIGNATED_INITIALIZER;
```
QuintinWillison, line 11: Minor but the previous signature aligned colons to keep things readable.

DeveloperNotes.md
```diff
@@ -57,3 +57,22 @@ Some files in the project are generated to maintain sync between
...
+
+Running `flutter analyze` in project root will analyze dart files in complete project,
+ i.e., plugin code and example code
+
+
+Or, use the good old dart analyzer
+
+```bash
+cd <root>/lib/
+dartanalyzer --fatal-warnings **/*.dart
```
QuintinWillison, line 74: You can get rid of the preceding `cd` commands by modifying these commands to be executable from root:

```suggestion
dartanalyzer --fatal-warnings lib/**/*.dart
```

ios/Classes/AblyStreamsChannel.m
tiholic: Bad news! `encodeErrorEnvelope` method is from flutter's source.

ios/Classes/AblyStreamsChannel.m
tiholic: Yes, I remember mentioning regarding `const`. I deliberately ignore any syntax related optimization inside code extracted from loup_v's StreamsChannel code.

I tried to change it to constant, but XCode hints:
`Sending 'const FlutterError *__strong' to parameter of type 'FlutterError * _Nonnull' discards qualifiers`
as the error is being passed on to `encodeErrorEnvelope`

ios/Classes/AblyStreamsChannel.h
paddybyers: We should adopt the following practice (shamelessly stolen from Google):

> Files that copy lines of code from third-party files, or that directly translated or derived from third-party files, must retain any copyright notices present in the third-party file(s). AnAbly copyright notice can be added to the license header as a separate copyright notice, beneath any existing notices, but above the license text.
> 
> Any files that are exact copies of files taken from a third-party source must retain any existing copyright notices and must not include any new copyright notices.
> 
> Entirely new, Ably-authored files being added to a project that is forked or derived from a third-party source should use the license header used by other files in the project, omitting any third-party copyright notices. AnAbly copyright notice can be added to the header above the license text.

ios/Classes/AblyFlutter.m
```diff
@@ -25,19 +34,11 @@ -(instancetype)init {
     return self;
 }
 
-#define ASSERT_NOT_DISPOSED \
```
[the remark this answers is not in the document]
tiholic: As we are using `sharedInstance`, we would not need to check if ably instance is present and not disposed. There will always be 1 and only 1 instance which will never be disposed.

But all Rest, Realtime, etc instances created inside Ably sharedInstance will be cleared on hot-restart

ios/Classes/AblyFlutter.h
```diff
@@ -8,6 +8,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface AblyFlutter : NSObject
 
++ (id)instance;
```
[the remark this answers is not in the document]
tiholic: Addressed in 86deac257bdc6054137ddedbfe31e85cd716412c

ios/Classes/AblyFlutterPlugin.m
tiholic: Addressed in 86deac257bdc6054137ddedbfe31e85cd716412c

ios/Classes/AblyStreamsChannel.m
tiholic: Addressed in 86deac257bdc6054137ddedbfe31e85cd716412c

ios/Classes/AblyFlutterMessage.h
tiholic: Addressed in 86deac257bdc6054137ddedbfe31e85cd716412c

DeveloperNotes.md
tiholic: Addressed in 86deac257bdc6054137ddedbfe31e85cd716412c

lib/src/ably_implementation.dart
QuintinWillison: No, don't worry for now. I'll sort this out later, outside of the scope of this PR.

ios/Classes/AblyFlutter.m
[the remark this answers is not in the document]
QuintinWillison: Fair enough. Back in my box, etc.. 😁  🤦

ios/Classes/AblyStreamsChannel.m
QuintinWillison: ok, perhaps an opportunity for your own wrapper - even if just a static local function?

ios/Classes/AblyStreamsChannel.m
QuintinWillison: Damn... my bad... shows how long it's been since I did any Objective-C (or even C) in anger... that should be `const*`, not `*const`! (i.e. const *pointer*!).

This is the problem with coming up with suggestions in GitHub's Web UI, not Xcode. 😄

ios/Classes/AblyStreamsChannel.h
QuintinWillison: @tiholic can you come up with something ^ or do you need us to assist?

android/src/main/java/io/ably/flutter/plugin/StreamsChannel.java
QuintinWillison: ^ @tiholic I still don't understand to what you were referring to in [your last question](https://github.com/ably/ably-flutter/pull/18#discussion_r453219515).

ios/Classes/AblyStreamsChannel.h
tiholic: I need help, please

android/src/main/java/io/ably/flutter/plugin/StreamsChannel.java
tiholic: I mean if you can point out the inconsistency by sharing the line numbers, to make it a bit clearer for me. Anyhow as you said, we can ignore this.

QuintinWillison: Oh, I get you now - you're referring the [my outer comment](https://github.com/ably/ably-flutter/pull/18#pullrequestreview-440140108).

I think the specific example I found was the [onMessage implementation](https://github.com/ably/ably-flutter/blob/d99a8e40f166333522d84a33c7ee7c23c98714bc/android/src/main/java/io/ably/flutter/plugin/StreamsChannel.java#L76), where for some non-apparent reason only the `reply` argument has been made `final`.

Nowhere else in that source file have arguments been made `final` which is inconsistent in and of itself as a single unit of source code.

I think we can ignore this for now as there's no real benefit on improving this right now. It may be something we can look for later on using lint / checkstyle. I'll resolve this thread. 😅

ios/Classes/AblyStreamsChannel.h
QuintinWillison: I'll push a commit.

QuintinWillison: Conformed: https://github.com/ably/ably-flutter/pull/18/commits/86df27e38e1dafead6485b55270e8237e8db8b14

@tiholic, please resolve this conversation if you're happy with my commit.

tiholic: Cool, thanks 😄

ios/Classes/AblyStreamsChannel.m
QuintinWillison: @tiholic Let's move on. This is unimportant.

lib/src/impl/platform_object.dart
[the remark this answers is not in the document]
zoechi: The PR is merged already but I stumbled upon this code in another PR.

First I'd suggest to make the getter a function. 
In my opinion getters with side effects (except caching) are usually not a good idea (I admit this is a bit controversial) 
but after I saw this code
```dart
    this.handle;  //proactively acquiring handle
```
I wondered what the purpose of accessing a field might be. 
Then of course I thought this might be a getter.
Then I saw the comment and my bad feeling about this code got reinforced.

I'd suggest to make it a function named `acquireHandle()`

About the content. I think 

```dart
Future<int> acquireHandle() => _handle ??= createPlatformInstance()
    .timeout(const Duration(milliseconds: 2));
```
would do the same as the current code.

Is it OK to continue the discussion here or should I create an issue or should it be discussed in Slack?

Also it feels like either `createPlatformInstance` https://github.com/ably/ably-flutter/blob/0c2f7074b7ffe6abcb5ba0a0a99aed7ab4917bd9/lib/src/impl/realtime/channels.dart#L34 
or `handle` should be `@protected` but I might not fully understand the purpose of the code yet.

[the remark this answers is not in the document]
zoechi: Not sure if this is related to this PR but perhaps the `reassemble` callback might be utilized if the handle needs to be reinitialized after hot-reload/restart 
https://proandroiddev.com/some-widely-used-override-methods-in-flutter-50696f08d42a#dc1c

PR #21, opened 2020-07-24
Realtime message subscription

tiholic:
Listening to messages on a Realtime Channel

This implementation follows dart idiomatic approach

#### Subscribing
```dart
Stream<ably.Message> messageStream = channel.subscribe();
StreamSubscription<ably.Message> channelMessageSubscription = messageStream.listen((ably.Message message){
  print("New message arrived ${message.data}");
});
```

#### Unsubscribing
```dart
await channelMessageSubscription.cancel();
```

### DIFF
--- a/DeveloperNotes.md
+++ b/DeveloperNotes.md
@@ -58,6 +58,26 @@ Some files in the project are generated to maintain sync between
 
 [Read about generation of platform specific constant files](bin/README.md)
 
+## Implementing new codec types
+
+1. Add type here along with value `bin/codegencontext.dart~_types`. Note, values must be unique
+2. Add an object definition here `bin/codegencontext.dart~objects` with object name and all properties.
+ This will create `Tx<ObjectName>` under which all properties are accessible.
+ 
+Generate platform constants and continue
+
+3. update `lib.src.codec.Codec~getCodecType` so new codec type is returned based on runtime type
+4. update `lib.src.codec.Codec~codecPair` so new encoder/decoder is assigned for new type
+5. update `android.src.main.java.io.ably.flutter.plugin.AblyMessageCodec~writeValue` so new codec type is obtained from runtime type
+6. update `android.src.main.java.io.ably.flutter.plugin.AblyMessageCodec~codecMap` so new encoder/decoder is assigned for new type
+7. update `ios.classes.codec.AblyFlutterWriter~writeValue` so new codec encoder assigned
+8. update `ios.classes.codec.AblyFlutterReader.getDecoder` so new codec decoder assigned
+
+## Implementing new platform methods
+
+1. Add new method name `bin/codegencontext.dart~_platformMethods`
+
+Generate platform constants and use wherever required
 
 ## Static plugin code analyzer
 
--- a/DeveloperNotes.md
+++ b/DeveloperNotes.md
@@ -60,22 +60,26 @@ Some files in the project are generated to maintain sync between
 
 ## Implementing new codec types
 
-1. Add type here along with value `bin/codegencontext.dart~_types`. Note, values must be unique
-2. Add an object definition here `bin/codegencontext.dart~objects` with object name and all properties.
+1. Add new type along with value in `_types` list at [bin/codegencontext.dart](bin/codegencontext.dart)
+2. Add an object definition  with object name and its properties to `objects` list at [bin/codegencontext.dart](bin/codegencontext.dart)
  This will create `Tx<ObjectName>` under which all properties are accessible.
  
 Generate platform constants and continue
 
-3. update `lib.src.codec.Codec~getCodecType` so new codec type is returned based on runtime type
-4. update `lib.src.codec.Codec~codecPair` so new encoder/decoder is assigned for new type
-5. update `android.src.main.java.io.ably.flutter.plugin.AblyMessageCodec~writeValue` so new codec type is obtained from runtime type
-6. update `android.src.main.java.io.ably.flutter.plugin.AblyMessageCodec~codecMap` so new encoder/decoder is assigned for new type
-7. update `ios.classes.codec.AblyFlutterWriter~writeValue` so new codec encoder assigned
-8. update `ios.classes.codec.AblyFlutterReader.getDecoder` so new codec decoder assigned
+3. update `getCodecType` in [lib.src.codec.Codec](lib/src/codec.dart) so new codec type is returned based on runtime type
+4. update `codecPair` in [lib.src.codec.Codec](lib/src/codec.dart)  so new encoder/decoder is assigned for new type
+5. update `writeValue` in [android.src.main.java.io.ably.flutter.plugin.AblyMessageCodec](android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java)
+ so new codec type is obtained from runtime type
+6. update `codecMap` in [android.src.main.java.io.ably.flutter.plugin.AblyMessageCodec](android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java)
+ so new encoder/decoder is assigned for new type
+7. add new codec encoder method in [ios.Classes.codec.AblyFlutterWriter](ios/Classes/codec/AblyFlutterWriter.m)
+ and update `writeValue` so that new codec encoder is called
+8. add new codec encoder method in [ios.classes.codec.AblyFlutterReader](ios/Classes/codec/AblyFlutterReader.m)
+ and update `getDecoder` so that new codec decoder is called
 
 ## Implementing new platform methods
 
-1. Add new method name `bin/codegencontext.dart~_platformMethods`
+1. Add new method name in `_platformMethods` list at [bin/codegencontext.dart](bin/codegencontext.dart)
 
 Generate platform constants and use wherever required
 
--- a/README.md
+++ b/README.md
@@ -137,8 +137,8 @@ await channel.detach();
 Subscribing to channel messages
 
 ```dart
-Stream<ably.Message> messageStream = channel.subscribe();
-StreamSubscription<ably.Message> channelMessageSubscription = messageStream.listen((ably.Message message){
+var messageStream = channel.subscribe();
+var channelMessageSubscription = messageStream.listen((ably.Message message){
   print("New message arrived ${message.data}");
 });
 ```
--- a/README.md
+++ b/README.md
@@ -143,7 +143,7 @@ StreamSubscription<ably.Message> channelMessageSubscription = messageStream.list
 });
 ```
 
-_use `channel.subscribe(name: "shifu")` or `channel.subscribe(names: ["shifu", "oogway"])` to listen to specific named messages_
+_use `channel.subscribe(name: "event1")` or `channel.subscribe(names: ["event1", "event2"])` to listen to specific named messages_
 
 UnSubscribing from channel messages
 
--- a/README.md
+++ b/README.md
@@ -134,6 +134,23 @@ Detaching from channel
 await channel.detach();
 ```
 
+Subscribing to channel messages
+
+```dart
+Stream<ably.Message> messageStream = channel.subscribe();
+StreamSubscription<ably.Message> channelMessageSubscription = messageStream.listen((ably.Message message){
+  print("New message arrived ${message.data}");
+});
+```
+
+_use `channel.subscribe(name: "shifu")` or `channel.subscribe(names: ["shifu", "oogway"])` to listen to specific named messages_
+
+UnSubscribing from channel messages
+
+```dart
+await channelMessageSubscription.cancel();
+```
+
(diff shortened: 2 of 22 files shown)


android/src/main/java/io/ably/flutter/plugin/AblyEventStreamHandler.java
```diff
@@ -89,6 +104,35 @@ public void onChannelStateChanged(ChannelStateChange stateChange){
...
+
+        PluginChannelMessageListener(EventChannel.EventSink eventSink){
+            super(eventSink);
+        }
+
+        public void onMessage(Message message){
+            System.out.print("message received with data: ");
+            System.out.println(message.data);
+            try {
+//                if(message.data instanceof JsonElement){
```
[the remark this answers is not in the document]
tiholic: Oops! I should have removed these after I fixed the encoding part in codec.
Addressed a027dfa8bbfc40b6ed233fa7fe82fbfe8d66eb52

android/src/main/java/io/ably/flutter/plugin/AblyEventStreamHandler.java
```diff
@@ -89,6 +104,35 @@ public void onChannelStateChanged(ChannelStateChange stateChange){
...
     }
 
+    private class PluginChannelMessageListener extends Listener implements Channel.MessageListener {
+
+        PluginChannelMessageListener(EventChannel.EventSink eventSink){
+            super(eventSink);
+        }
+
+        public void onMessage(Message message){
+            System.out.print("message received with data: ");
```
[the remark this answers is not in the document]
tiholic: Created: https://github.com/ably/ably-flutter/issues/24

README.md
```diff
@@ -134,6 +134,23 @@ Detaching from channel
...
+Subscribing to channel messages
+
+```dart
+Stream<ably.Message> messageStream = channel.subscribe();
+StreamSubscription<ably.Message> channelMessageSubscription = messageStream.listen((ably.Message message){
+  print("New message arrived ${message.data}");
+});
+```
+
+_use `channel.subscribe(name: "shifu")` or `channel.subscribe(names: ["shifu", "oogway"])` to listen to specific named messages_
```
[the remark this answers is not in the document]
tiholic: addressed @ 6a7e9b0c4b101585358d24613ebe8019b7c46b60

DeveloperNotes.md
```diff
@@ -58,6 +58,26 @@ Some files in the project are generated to maintain sync between
 
 [Read about generation of platform specific constant files](bin/README.md)
 
+## Implementing new codec types
+
+1. Add type here along with value `bin/codegencontext.dart~_types`. Note, values must be unique
```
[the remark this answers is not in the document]
tiholic: Addressed @ f40547b3856c13d9c9d1cd376d8ed63ffd474593

DeveloperNotes.md
```diff
@@ -58,6 +58,26 @@ Some files in the project are generated to maintain sync between
 
 [Read about generation of platform specific constant files](bin/README.md)
 
+## Implementing new codec types
```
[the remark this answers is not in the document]
tiholic: Addressed @ f40547b3856c13d9c9d1cd376d8ed63ffd474593

android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
```diff
@@ -257,5 +279,18 @@ private TokenDetails readTokenDetails(Map<String, Object> jsonMap) {
         return jsonMap;
     }
 
+    private Map<String, Object> encodeChannelMessage(Message c){
+        if(c==null) return null;
```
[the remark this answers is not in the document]
tiholic: Would you like to have an issue for this?
I can only think about moving this null check to `encodeAblyMessage` in `Codec.dart` once we implement all data types that required to be encoded.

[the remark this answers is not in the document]
tiholic: @QuintinWillison Would you like to have an issue for this?
I can only think about moving this null check to `encodeAblyMessage` in `Codec.dart` once we implement all data types that required to be encoded.

lib/src/impl/realtime/channels.dart
```diff
@@ -136,14 +136,29 @@ class RealtimePlatformChannel extends PlatformObject implements spec.RealtimeCha
...
+            if(message.name==name){
+              sink.add(message);
+            }
+          } else {
+            sink.add(message);
+          }
+        }
+      )
+    );
+    return stream;
```
zoechi, line 161: It's not clear to me what `StreamTransformer.fromHandlers(...)` is required for.
My suggestion would be
```dart
  Stream<spec.Message> subscribe({String name, List<String> names}) {
    final subscribedNames =
        {if (name != null) name, ...?names}.where((n) => n != null).toList();
    return listen(PlatformMethod.onRealtimeChannelMessage, _payload)
        .map((message) => message as spec.Message)
        .where((message) =>
            subscribedNames.isEmpty ||
            subscribedNames.any((n) => n == message.name));
  }
```

lib/src/spec/message.dart
```diff
@@ -2,16 +2,40 @@ import 'enums.dart';
 import 'rest/channels.dart';
 
 class Message {
-  Message.fromEncoded(Map jsonObject, ChannelOptions channelOptions);
-  Message.fromEncodedArray(List<Map> jsonArray, ChannelOptions channelOptions);
+
```
zoechi, line 5: Are there any comments in similar classes in implementations for other languages that are more elaborate/exact?
If there are, it might be worth considering to copy them, to perhaps have better comments
and comments that are at least somewhat consistent with the other implementations.

"The timestamp for this message" is a quite redundant comment.

lib/src/spec/message.dart
```diff
@@ -2,16 +2,40 @@ import 'enums.dart';
...
   String connectionId;
-  dynamic data;
+
+  /// Any transformation applied to the data for this message
   String encoding;
-  dynamic extras;
-  String id;
+
+  /// The message payload
+  dynamic data;
```
zoechi, line 22: I think `dynamic` is almost always a bad idea (except when implementing `noSuchMethod`).
Either use `Object` or any more concrete class (or interface) if possible.
@QuintinWillison what do you think?

lib/src/spec/message.dart
tiholic: These comments are actually copied from java implementation

lib/src/spec/message.dart
tiholic: I think object should do.
Acceptable types for `data` are `null`, `String`, `Map` and `List`

lib/src/spec/message.dart
zoechi: that's sad🥴

lib/src/spec/message.dart
zoechi: ok, in this case object (and dynamic) are the only options

lib/src/impl/realtime/channels.dart
tiholic: Also, https://github.com/ably/ably-flutter/pull/20#discussion_r464451601

`on` method in `realtime/channels.dart` and `realtime/connection` to be optimized in same fashion

tiholic: Addressed 5359c91a64e4da5d7988a18d022be5e22038013f

lib/src/spec/message.dart
tiholic: addressed 9b9d8a44b61653d442cd5b90c49481fbb67b78ff

example/lib/main.dart
```diff
@@ -324,6 +331,37 @@ class _MyAppState extends State<MyApp> {
...
+          "\tisNull: ${message.data == null}\n"
+          "\tisString ${message.data is String}\n"
+          "\tisMap ${message.data is Map}\n"
+          "\tisList ${message.data is List}\n");
+        setState((){
+          channelMessage = message;
+        });
+      });
+      print("Channel messages subscribed");
+      setState((){});
```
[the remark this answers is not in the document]
tiholic: It is neither 😂 
An honest mistake in copying code from another method.

Addressed 246e02fbe90297be1cd51c653e3c213c613426ea

README.md
```diff
@@ -134,6 +134,23 @@ Detaching from channel
...
 ```
 
+Subscribing to channel messages
+
+```dart
+Stream<ably.Message> messageStream = channel.subscribe();
+StreamSubscription<ably.Message> channelMessageSubscription = messageStream.listen((ably.Message message){
+  print("New message arrived ${message.data}");
+});
+```
```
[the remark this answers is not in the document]
tiholic: addressed d709f2d8421140cf7d69b519588e6dda49d5a9c6

android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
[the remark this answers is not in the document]
QuintinWillison: @zoechi am I fussing over nothing here? I would value your opinion on this. Thanks!

lib/src/spec/message.dart
QuintinWillison: Thanks both. My preference will always be strongly typed but, alas, we're somewhat beholden to [the spec](https://docs.ably.io/client-lib-development-guide/features/#TM2d).

There could be an opportunity to do something more strongly typed here so I would be open to suggestions. Perhaps `data` could present an API that offers the app developer to use language akin to the `asString()`, `asBytes()` seen elsewhere?

Feels like it cuts to the core of Ably's "conformed APIs" approach (vs idiomatic) so would be interested in views from @paddybyers, @mattheworiordan and @SimonWoolf.

tiholic: @QuintinWillison _Just FYI_, in `ably-java` we have **message payload** as `Object` and in `ably-cocoa` as `id`

README.md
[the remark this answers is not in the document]
zoechi: I would have removed the `ably.Message` for the closure parameter as well.
I'll leave it to the linter rules decisions which would complain about that anyway if the related rules were enabled.

android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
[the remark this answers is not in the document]
zoechi: `null` is a valid JSON value and if the message passed is `null`, then `null` is the correct conversion result.
So, I do not see anything wrong with this.

It would be different if `null` were not a valid value for the parameter `c`, then an assert or throwing an exception would be more appropriate.

lib/src/spec/message.dart
zoechi: Ok, I leave it at that for now. 
Perhaps when I understand the API better I can come up with a better comment.

lib/src/spec/message.dart
zoechi: I have the same opinion about strongly typed, but it looks there are no good options that would be a real improvement.
A wrapper class could be used with a field per property value type, but `Map` and `List` would need generic parameters to actually provide notably more type safety than Object.

In the message receive case that's hardly possible because I guess we usually do not know what type the message will have at compile time.

android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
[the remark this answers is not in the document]
QuintinWillison: Back in my box. 😝

lib/src/spec/message.dart
QuintinWillison: We don't but, in my opinion, the app developer might reasonably know - therefore they then call the correct helper method (or is that getter) on the returned object.

zoechi: We could wait until Dart supports union types which seems to be part of the next bigger chunk on the roadmap https://github.com/dart-lang/language/issues/546 which might take a few years 
or create a wrapper class similar to a union type ourselves until Dart provides that.

zoechi: I'd suggest to create a new issue for that instead of making in part of this PR

mattheworiordan: Personally I think we should go with what's easiest to implement now and avoid introducing new "special" types that deviate significantly from other SDKs.

zoechi: @QuintinWillison Sorry, I didn't mean to dismiss your suggestion with different accessor methods/getters. 
It's rather that I can't imagine how that would work out, but that's definitely because I do not know Ably's API well yet.

tiholic: Created an issue to move this discussion forward: https://github.com/ably/ably-flutter/issues/35

I've also added some inputs based on @QuintinWillison's suggestion

PR #26, opened 2020-07-31
Fix: iOS hot-restart issue with active listeners

tiholic:
Dropping key from streams dict while enumerating - [err: Collection was mutated while being enumerated]
Fixed by using a `mutableCopy` of `_streams`

### DIFF
--- a/ios/Classes/AblyStreamsChannel.m
+++ b/ios/Classes/AblyStreamsChannel.m
@@ -99,7 +99,7 @@
 }
 
 - (void) reset{
-    for (NSString* key in _streams) {
+    for (NSString* key in [_streams mutableCopy]) {
         AblyStreamsChannelStream *stream = _streams[key];
         [stream.handler onCancelWithArguments:nil];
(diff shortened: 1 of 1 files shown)


ios/Classes/AblyStreamsChannel.m
```diff
@@ -99,7 +99,7 @@ - (void)setStreamHandlerFactory:(NSObject<FlutterStreamHandler> *(^)(id))factory
 }
 
 - (void) reset{
-    for (NSString* key in _streams) {
+    for (NSString* key in [_streams mutableCopy]) {
```
QuintinWillison, line 102: I know this has been merged but, fwiw, that could just be `copy` as it's not needed to be mutable. Cosmetic, admittedly. 😁

tiholic: @QuintinWillison Good to know there is an immutable copy for dicts 😄 

Addressed at efc5104 in PR #20

PR #32, opened 2020-08-05
Drive test example

zoechi:
Because flutter drive tests need to be a flutter application project I created such a project as sub-project in the `test_integration` directory. 
I'm not sure if this is a good name.  
I didn't want to name it `test_driver` because that is normally used to contain driver tests directly.  
There is not really a rule that would limit what the name could be.

The `test_driver.sh` batch file executes all driver tests in that sub-project.

`test_integration/README.md` contains further information.


test_integration/test_driver/app_test.dart
```diff
@@ -0,0 +1,48 @@
...
+
+      final result = await driver.requestData(message.toJson());
+      final response = TestControlMessage.fromJson(result);
+
+      expect(response.testName, message.testName);
+
+      // Use the `driver.getText` method to verify the counter starts at 0.
+      expect(response.payload['platformVersion'], 'Android 10');
+      expect(response.payload['ablyVersion'], 'android-1.1.10');
+    });
```
tiholic, line 32: Not sure if we should hardcode the version details but considering this as just an example, cool!

zoechi: You are right of course. 
I don't know the API well enough yet to know what test makes sense.
This is mostly about demonstrating how to write and run such tests.

test_driver.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
```
QuintinWillison, line 1: I think just remove the `cd` commands and move this script into the `test_integration` folder. I would like to keep the root as clean as possible.

test_integration/.gitignore
```diff
@@ -0,0 +1,44 @@
+# Miscellaneous
```
QuintinWillison, line 1: I think much of the contents of this file may well be already in the root ignore file. Could you clean it up a bit? I tend to try to have sub-folder Git ignores only have the stuff they absolutely need and then keep the boilerplate ignores for the root.

test_integration/README.md
```diff
@@ -0,0 +1,123 @@
+# Ably Realtime integration tests
+
+This is for tests that depend on the Ably plugin to be available and
+functional and for tests to be executed on Android and iOS devices or 
+emulators.
+
+## Execute the Flutter Driver Tests
+
+Run the bash script in the project root directory
```
QuintinWillison, line 9: Again, if you remove the `cd` commands from the script then this can simplify.

test_integration/.gitignore
zoechi: Good idea. I didn't even look at it. It's auto-generated by `flutter create ...`

test_driver.sh
zoechi: Done

test_integration/.gitignore
zoechi: Done

test_integration/README.md
zoechi: Done

PR #33, opened 2020-08-08
REST authCallback

tiholic:
Support for `authCallback` client option

### Checklist
- [x] Android
- [x] iOS

### How it works
1. `Rest` instance will be created with `authCallback` in `ClientOptions`
2. `RestChannel#publish` will be triggered from user code
3. ably library code calls a proxy `authCallback` on platform code and that will immediately return null, but will make a platform call to dart side for a token data.
4. dart side receives the MethodChannel call and calls `authCallback` and return the value to platform code.
5. platform code will set this response token data in temporary storage.
6. `RestChannel#publish` from `#2` will receive a `80019` error which will suppress the error and will check a boolean on dart side if the authentication update was complete every 100ms and then retry publishing again.
7. This time, platform code will return ably library with the token data stored temporarily in `#5` above


zoechi (changes requested):

ios/Classes/AblyFlutter.h
```diff
@@ -3,13 +3,16 @@
...
 @class ARTClientOptions;
+#import <Flutter/Flutter.h>
 
 NS_ASSUME_NONNULL_BEGIN
 
 @interface AblyFlutter : NSObject
 
 + (instancetype)sharedInstance;
 
+-(instancetype) setChannel:(FlutterMethodChannel *const)channel;
```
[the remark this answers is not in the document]
QuintinWillison: Actually I've now found where you do use it and this is not an acceptable pattern - you're basically returning `self` as a side effect of a setter. Please, no. I'll comment at the calling site further...

ios/Classes/AblyFlutterPlugin.m
```diff
@@ -223,7 +223,7 @@ -(instancetype)initWithChannel:(FlutterMethodChannel *const)channel streamsChann
     if (!self) {
         return nil;
     }
-    _ably = [AblyFlutter sharedInstance];
+    _ably = [[AblyFlutter sharedInstance] setChannel: channel];
```
QuintinWillison, line 226: As commented on earlier where you define `setChannel:`...
At a minimum I would expect something like:

```suggestion
    _ably = [AblyFlutter sharedInstance];
    [_ably setChannel:channel];
```

ios/Classes/types/ARTClientOptions+AblyFlutterClientOptions.h
```diff
@@ -0,0 +1,11 @@
+#import "ARTClientOptions.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface ARTClientOptions (AblyFlutterClientOptions)
```
QuintinWillison, line 5: I can see what you're trying to do by using a category here but I am concerned that your new property is mutable when, as I imagine, for the lifetime of an instance it will not need to change once initially set.

Perhaps this might be an example where composition would serve more accurately what you're looking to achieve. Something like:

```objc
@interface AblyFluterClientOptions : NSObject

-(instancetype)initWithClientOptions:(ARTClientOptions *)clientOptions
                     hasAuthCallback:(BOOL)hasAuthCallback NS_DESIGNATED_INITIALZER;

@property (nonatomic, readonly) ARTClientOptions* clientOptions;
@property (nonatomic, readonly) BOOL hasAuthCallback;

@end
```

ios/Classes/types/ARTClientOptions+AblyFlutterClientOptions.m
```diff
@@ -0,0 +1,16 @@
+#import "ARTClientOptions+AblyFlutterClientOptions.h"
+#import <objc/runtime.h>
+
+@implementation ARTClientOptions (AblyFlutterClientOptions)
+
+- (void) setHasAuthCallback:(BOOL) value {
+    NSNumber *number = [NSNumber numberWithBool: value];
+    objc_setAssociatedObject(self, @selector(hasAuthCallback), number, OBJC_ASSOCIATION_RETAIN_NONATOMIC);
```
QuintinWillison, line 8: Oh, yes, even more reasons not to use a category here! Oooohhhh.... 😬

android/src/main/java/io/ably/flutter/plugin/AblyEventStreamHandler.java
```diff
@@ -125,6 +124,7 @@ public void onListen(Object object, EventChannel.EventSink uiThreadEventSink) {
         final AblyEventMessage<Object> eventMessage = ablyMessage.message;
         final String eventName = eventMessage.eventName;
         try {
+            ChannelOptions channelOptions;
```
QuintinWillison, line 127: I'm not sure why this local variable has to sit outside at this scope. There should be an instance of it for each case that uses it, is all.

android/src/main/java/io/ably/flutter/plugin/types/PlatformClientOptions.java
```diff
@@ -0,0 +1,14 @@
+package io.ably.flutter.plugin.types;
+
+import io.ably.lib.types.ClientOptions;
+
+public class PlatformClientOptions extends ClientOptions {
```
QuintinWillison, line 5: Akin to my comment on the Objective-C code earlier, this does not feel like the right place for inheritance. I would use composition here as well. e.g. something adopting:

```java
interface PlatformClientOptions {
    ClientOptions getClientOptions();
    boolean hasAuthCallback();
}
```

or similar.

QuintinWillison (changes requested):
I've just focussed on the Java and Objective-C code for now, ignoring the generated stuff.

example/lib/provisioning.dart
```diff
@@ -37,17 +37,26 @@ class AppKey {
...
     throw HttpException('Server didn\'t return success. Status: ' + response.statusCode.toString());
   }
+  return jsonDecode(response.body);
+}
 
-  // Cherry pick the pieces from the JSON response that we need.
-  final Map result = jsonDecode(response.body);
-  final List keys = result['keys'];
-  final Map key = keys[0];
+Future<AppKey> provision(final String environmentPrefix) async {
```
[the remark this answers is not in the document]
tiholic: We must wait for #38 too!

example/lib/provisioning.dart
```diff
@@ -37,17 +37,26 @@ class AppKey {
   String toString() => _keyStr;
 }
 
-Future<AppKey> provision(final String environmentPrefix) async {
+Future<Map> _provisionApp(final String environmentPrefix) async {
```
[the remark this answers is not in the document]
tiholic: 5397f8c227c3a43f6a3100f5c98deb3dcd8e9f85

example/lib/provisioning.dart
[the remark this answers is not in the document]
tiholic: 5397f8c227c3a43f6a3100f5c98deb3dcd8e9f85

example/lib/provisioning.dart
```diff
@@ -37,17 +37,26 @@ class AppKey {
...
-  final Map result = jsonDecode(response.body);
-  final List keys = result['keys'];
-  final Map key = keys[0];
+Future<AppKey> provision(final String environmentPrefix) async {
+  final Map result = await _provisionApp(environmentPrefix);
+  final Map key = result['keys'][0];
   return AppKey(key['keyName'], key['keySecret'], key['keyStr']);
 }
+
+Future<Map> getTokenRequest() async {
```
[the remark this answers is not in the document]
tiholic: 5397f8c227c3a43f6a3100f5c98deb3dcd8e9f85

lib/src/impl/rest/rest.dart
```diff
@@ -7,6 +7,7 @@ import '../../spec/spec.dart' as spec;
 import '../platform_object.dart';
 import 'channels.dart';
 
+Map<int, Rest> restInstances = {};
```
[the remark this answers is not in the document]
tiholic: @zoechi restInstances is accessed in `AblyMethodCallHandler#onAuthCallback` (method_call_handler.dart)

We can alternatively have a static method on `Rest` like `Rest.getInstanceByHandle(int handle)` to get a rest instance and make `restInstances` as private.

lib/src/spec/common.dart
```diff
@@ -74,31 +74,179 @@ abstract class StatsMessageTraffic {
...
 }
 
+/// A class providing parameters of a token request.
+/// These params are used when invoking
+/// Auth#authorize, Auth#requestToken and Auth#createTokenRequest
+///
+/// spec: https://docs.ably.io/client-lib-development-guide/features/#TK1
 class TokenParams {
+
+  /// Capability of the token. If the token request is successful,
```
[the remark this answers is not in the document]
tiholic: d52d03a8ab85639ac368e4a7e04e16059d02e4db

lib/src/spec/common.dart
```diff
@@ -74,31 +74,179 @@ abstract class StatsMessageTraffic {
   StatsMessageTypes webhook;
 }
 
+/// A class providing parameters of a token request.
+/// These params are used when invoking
```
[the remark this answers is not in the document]
tiholic: d52d03a8ab85639ac368e4a7e04e16059d02e4db

lib/src/spec/common.dart
```diff
@@ -74,31 +74,179 @@ abstract class StatsMessageTraffic {
   StatsMessageTypes webhook;
 }
 
+/// A class providing parameters of a token request.
```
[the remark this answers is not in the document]
tiholic: d52d03a8ab85639ac368e4a7e04e16059d02e4db

lib/src/method_call_handler.dart
```diff
@@ -0,0 +1,25 @@
...
+import 'package:ably_flutter_plugin/src/impl/message.dart';
+import 'package:flutter/services.dart';
+
+class AblyMethodCallHandler {
+  AblyMethodCallHandler(MethodChannel channel) {
+    channel.setMethodCallHandler((call) async {
+      switch (call.method) {
+        case PlatformMethod.authCallback:
+          return await onAuthCallback(call.arguments);
+      }
```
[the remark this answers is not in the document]
tiholic: Throwing `PlatformException`: 06ae0bffb8c89d324f9d7aa5bcc560b2cafa35c9

lib/src/codec.dart
```diff
@@ -122,7 +132,8 @@ class Codec extends StandardMessageCodec {
 
     // =========== ENCODERS ===========
     /// Writes [value] for [key] in [map] if [value] is not null
-    writeToJson(Map<String, dynamic> map, key, value){
+    writeToJson(Map<String, dynamic> map, String key, Object value){
```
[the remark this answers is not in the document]
tiholic: I'll leave this file as is for this PR? Explore and fix all the issues as detailed in 43 as a part of separate PR after linting?

ios/Classes/AblyFlutterPlugin.m
tiholic: 772c54ec3cada003fed9a2b8230cfd3f5c51822f

lib/src/codec.dart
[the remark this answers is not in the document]
zoechi: Works for me

ios/Classes/types/ARTClientOptions+AblyFlutterClientOptions.h
tiholic: Done, 167da5cf297b94e8025bb6f27ebc6abbec1075ef

ios/Classes/types/ARTClientOptions+AblyFlutterClientOptions.m
tiholic: 167da5cf297b94e8025bb6f27ebc6abbec1075ef

ios/Classes/AblyFlutterPlugin.m
```diff
@@ -52,11 +52,14 @@ -(void)registerWithCompletionHandler:(FlutterResult)completionHandler;
     AblyFlutterMessage *const messageData = message.message;
     NSMutableDictionary<NSString *, NSObject *> *const _dataMap = messageData.message;
     NSString *const channelName = (NSString*)[_dataMap objectForKey:@"channel"];
-    NSString *const eventName = (NSString*)[_dataMap objectForKey:@"name"];
+    NSString * eventName = (NSString*)[_dataMap objectForKey:@"name"];
+    if(eventName==[NSNull null]){
```
tiholic, line 56: @QuintinWillison any better way to do this?

Dart side will not be sending the key if there is no value, and ably-cocoa will throw errors when name is NSNull

https://github.com/ably/ably-cocoa/blob/main/Source/ARTMessage.m#L52

android/src/main/java/io/ably/flutter/plugin/AblyEventStreamHandler.java
tiholic: It will be duplicate definition if declared inside each switch case

android/src/main/java/io/ably/flutter/plugin/types/PlatformClientOptions.java
tiholic: c9b9c89bd121150fe50d02d2fc03bd8216ad0fbe

lib/src/impl/rest/rest.dart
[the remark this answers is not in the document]
zoechi: It's not part of this PR but at line 38
```dart
@override
ClientOptions options;
```
probably should be
```dart
@override
final ClientOptions options;
```
otherwise the asserts in the constructor do not make sense.

For this you probably need to change `spec/rest/ably_base.dart`
```dart
  ClientOptions options;
``` 
to
```dart
  ClientOptions get options;
```
Please also check all other fields in `ably_base.dart` if they should actually provide a setter as well, or if just getters are enough for these interfaces.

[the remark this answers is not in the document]
zoechi: I assume that accesses from outside this file should only be allowed to read, but not to add, remove, or replace?
In this case I'd use
```dart
import 'dart:collection';
...
Map<int, Rest> _restInstances = {};
Map<int, Rest> _restInstancesUnmodifiableView;
Map<int, Rest> get restInstances => _restInstancesUnmodifiableView ??= UnmodifiableMapView(_restInstances);
```
and change update access in the rest of the file to the private `_restInstances`

A bit verbose, but it communicates clearly that modifications are not supposed to be made.
`UnmodifiableMapView` does not provide feedback through static analyis, it only throws at modification attempts. The advantage is that it passes as regular `Map`.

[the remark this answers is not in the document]
tiholic: @zoechi 
About your previous comment:
> It's not part of this PR but at line 38...

These values can be changed any time, they need not be locked on initializing. At least that's what I could infer from [the IDL](https://docs.ably.io/client-lib-development-guide/features/#idl)

There are many other classes with such behavior in `spec/` folder. Can we have an issue to discuss this?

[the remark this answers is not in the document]
tiholic: 25bda8181091fce8e7b86b12804d26e55ce5e581

ios/Classes/AblyFlutterPlugin.m
QuintinWillison: I think just leave it as is for now - it's an internal detail so we can see how it scales over time.

lib/src/impl/rest/rest.dart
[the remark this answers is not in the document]
QuintinWillison: @zoechi can you resolve this conversation if you're happy with @tiholic's change?

QuintinWillison (approved):

lib/src/spec/common.dart
```diff
@@ -74,31 +75,189 @@ abstract class StatsMessageTraffic {
...
+    this.capability,
+    this.clientId,
+    this.nonce,
+    this.timestamp,
+    this.ttl
+  });
+
 }
 
+/// TokenDetails is a type containing the token request
```
zoechi, line 141: This comment should be improved. 
Using the name of the commented item in the comment usually leads to comments that do not add much value. Also that `TokenDetail` is a type is not useful, because `class TokenDetails` tells that already.

My suggestion

```dart
/// Response to a REST requestToken request.
```

lib/src/spec/common.dart
```diff
@@ -74,31 +75,189 @@ abstract class StatsMessageTraffic {
...
+///
+/// response from the REST requestToken endpoint
+///
+/// spec: https://docs.ably.io/client-lib-development-guide/features/#TD1
 class TokenDetails {
-  TokenDetails(this.token);
-  String capability;
-  String clientId;
+
+  /// TD2
```
zoechi, line 148: Can we use `Ref: ...` consistently to such ids? (like `/// Ref: TK2a`
`TD2` is rather confusing for someone not familiar with the organization of this project.
`Ref:` is not perfect, but is at least a hint that somewhere else more information can be found.

lib/src/spec/common.dart
```diff
@@ -74,31 +75,189 @@ abstract class StatsMessageTraffic {
...
+    token = map['token'] as String;
+    expires = map['expires'] as int;
+    issued = map['issued'] as int;
+    capability = map['capability'] as String;
+    clientId = map['clientId'] as String;
+  }
+
+}
+
+/// TokenRequest is a type containing the token request
```
zoechi, line 195: This line is pure noise and prone to upset people looking for helpful information.

lib/src/spec/common.dart
zoechi: Or even much better 
```dart
/// spec: https://docs.ably.io/client-lib-development-guide/features/#TE1
```

lib/src/spec/common.dart
```diff
@@ -74,31 +75,189 @@ abstract class StatsMessageTraffic {
...
+/// details sent to the REST requestToken endpoint
+///
+/// spec: https://docs.ably.io/client-lib-development-guide/features/#TE1
+class TokenRequest {
+
+  /// The keyName of the key against which this request is made.
```
zoechi, line 201: The comment doesn't provide much insight beyond what the field name already tells, but at least `The keyName` should be changed to `The name`.

lib/src/spec/common.dart
```diff
@@ -74,31 +75,189 @@ abstract class StatsMessageTraffic {
...
-  String mac;
+
+  /// An opaque nonce string of at least 16 characters to ensure
+  ///	uniqueness of this request. Any subsequent request using the
+  ///	same nonce will be rejected.
+  ///
+  /// TE2, TE5
   String nonce;
+
+  /// The Message Authentication Code for this request. See the Ably
```
zoechi, line 213: There should be an empty line before `See the Ably ...`

zoechi (changes requested):

lib/src/spec/common.dart
tiholic: @zoechi  There are 2 parts of an Ably Access Key: `keyName` and `keyToken`.
They are together called as `Key` ( `<keyName>:<keyToken>` is the format for a `Key` )

Let me know if we can improvised based on this context!

lib/src/spec/common.dart
tiholic: @zoechi there are 2 lines that makes helpful information, not sure if that would be really helpful, we can remove that altogether if it wouldn't make sense.

```
/// TokenRequest is a type containing the token request
/// details sent to the REST requestToken endpoint
```

This content is copied from the docs here: https://docs.ably.io/client-lib-development-guide/features/#TE1

lib/src/impl/rest/rest.dart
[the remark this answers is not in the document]
zoechi: Sorry, I missed these comments. 
Why did you change it when they are supposed to be changeable?
Turned it out that they actually shouldn't be modifiable?

lib/src/spec/common.dart
tiholic: @zoechi TD2 is the item right under TD1 in the spec, the class commentary consists of link to TD1, so I felt just mentioning the spec ID would be sufficient. Let me know if it is a better idea to explicitly mention it for each of the fields.

lib/src/spec/common.dart
tiholic: I think we better say:

"An encapsulating class TokenDetails acquired by making a request to `requestToken` endpoint"

lib/src/spec/common.dart
zoechi: I'd rather have no comments than comments that just state the obvious.
Others might have different opinions.
Actually I'd just remove it alltogether (except the reference)

lib/src/spec/common.dart
zoechi: Your comment above would be already quite an improvement to the comment. I'm just wondering where the mentioned associated token can be found. Also access key says more than just key. Is there a way to add a link to the spec of what access key means?

tiholic: That would be helpful I suppose. Let me see if I can find anything on that front.

tiholic: Ah! there. https://docs.ably.io/client-lib-development-guide/features/#RSA11

lib/src/spec/common.dart
zoechi:
```
Response to a `requestToken` endpoint
```
seems to be enough for me, except that's I'm not sure what `endpoint` means here. Perhaps 
```
Response to a `requestToken` request
```
would be better. `ecapsulating class` and `TokenDetails` doesn't add any value in my opinion.

lib/src/spec/common.dart
zoechi: I'd add full links everywhere like shown in https://dart.dev/guides/language/effective-dart/documentation#markdown

```
/// spec: https://docs.ably.io/client-lib-development-guide/features/#TE1
```
or
```
/// [spec: TE1](https://docs.ably.io/client-lib-development-guide/features/#TE1)
```
or
```
/// [spec: TE1][https://docs.ably.io/client-lib-development-guide/features/#TE1]
```
Doesn't make much of a difference, but if someone browses API docs in https://pub.dev the links help to navigate directly to the spec

lib/src/spec/common.dart
zoechi: So the value in `keyName` is a combination of key name and token separated with a `:`?
I'd definitely add that link.

lib/src/spec/common.dart
```diff
@@ -74,31 +75,189 @@ abstract class StatsMessageTraffic {
...
-abstract class TokenRequest {
+  /// stringified capabilities JSON
+  ///
+  /// TD5
   String capability;
+
+  /// clientId assigned to the token.
+  ///
+  /// If clientId is not set (i.e. null), then the token is prohibited
+  /// from assuming a clientId in any operations, however if clientId
```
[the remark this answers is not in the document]
tiholic: 0bf7a8e5317502191c32c116ed5bca4fad8c43a0

lib/src/spec/common.dart
tiholic: a162720d5a628526e5257889ace180d9fbfe0a48

lib/src/spec/common.dart
tiholic: a162720d5a628526e5257889ace180d9fbfe0a48

lib/src/spec/common.dart
tiholic: 28208cf1cc513eb384ca2a99b314abfa74672508

lib/src/spec/common.dart
tiholic: updated: 28208cf1cc513eb384ca2a99b314abfa74672508

zoechi (approved):

Issue #30, opened 2020-08-04
Add linting rules


tiholic:
Rules: https://dart-lang.github.io/linter/lints/

Discussion: https://ably-real-time.slack.com/archives/C0178QGC093/p1596456192026600

tiholic:
https://github.com/ably/ably-flutter/pull/20#discussion_r464403047

ben-xD:
https://pub.dev/packages/pedantic has been deprecated. 

> Instead, please see package:lints which is the lint recommendation from the Dart team, or package:flutter_lints which extends that for Flutter. These are better than the Google lint set because they take into account the needs of all Dart users, not just Google engineers.
>
> To migrate away from package:pedantic please see the instructions here.

In the latest release from Flutter, they mentioned https://pub.dev/packages/flutter_lints 

We should switch to use this instead

ben-xD:
Closing this because of the above comment, a new issue has been created: https://github.com/ably/ably-flutter/issues/168

PR #34, opened 2020-08-10
Basis for discussion about what linter rules to enable

zoechi:
Especially the rules with a comment`personal preference` need to be decided per project/team,
but others might as well need adjustment to specific requirements in this project or preferences of the team.

Some of the enabled rules might not yet be supported in the Dart version you are using.

The easiest way is to check out the branch and open a Dart file.
Then the DartAnalyzer should start analyzing and show all violations in the error list.

I think a good way for a discussion would be to link to code examples where you think the suggestions
make the code worse and if possible some rationale.

Many linter rules (but far from all) provide quick fixes that make it easy to adjust the code to conform to the rules. Most IDEs should support this as they are usually all driven by DartAnalyzer.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -116,7 +116,6 @@ linter:
     - library_names
     - library_prefixes
     - lines_longer_than_80_chars
-# Too annoying when long strings with interpolations are broken into multiple lines which we do often
     - missing_whitespace_between_adjacent_strings
     - no_default_cases             # experimental
     - no_runtimeType_toString
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -31,9 +31,9 @@ linter:
     - avoid_slow_async_io
     - avoid_types_as_parameter_names
     - avoid_web_libraries_in_flutter
-    # subject to false positives, perhaps worth supressing in code in such cases
+    # subject to false positives, perhaps worth suppressing in code in such cases
     - cancel_subscriptions
-    # subject to false positives, perhaps worth supressing in code in such cases
+    # subject to false positives, perhaps worth suppressing in code in such cases
     - close_sinks
     - comment_references
     - control_flow_in_finally
@@ -53,8 +53,6 @@ linter:
     - throw_in_finally
     - unnecessary_statements
     - unrelated_type_equality_checks
-    - unsafe_html
-    - use_key_in_widget_constructors
     - valid_regexps
 
 ## Style Rules
@@ -63,7 +61,7 @@ linter:
 #    - always_put_control_body_on_new_line # works against dartfmt
     - always_put_required_named_parameters_first
     - always_require_non_null_named_parameters
-#    - always_specify_types        # this works agains general Dart style guides
+#    - always_specify_types        # this works against general Dart style guides
     - annotate_overrides
     - avoid_annotating_with_dynamic
 #    - avoid_as                    # obsolete - this was considered good practice in Dart 1
@@ -91,10 +89,9 @@ linter:
     - avoid_returning_null_for_void
     - avoid_returning_this
     - avoid_setters_without_getters
-    - avoid_shadowing_type_parameters # personal preference
+    - avoid_shadowing_type_parameters
     - avoid_single_cascade_in_expression_statements
     - avoid_types_on_closure_parameters
-    - avoid_unnecessary_containers
     - avoid_unused_constructor_parameters
     - avoid_void_async
     - await_only_futures
@@ -102,14 +99,13 @@ linter:
     - camel_case_types
     - cascade_invocations
     - constant_identifier_names
-    - curly_braces_in_flow_control_structures # personal preference
+    - curly_braces_in_flow_control_structures
     - directives_ordering
-    - do_not_use_environment # depends on project requirements
+    - do_not_use_environment
     - empty_catches
     - empty_constructor_bodies
     - exhaustive_cases
     - file_names
-#    - flutter_style_todos         # I find the required ":" annoying and pointless
     - implementation_imports
     - join_return_with_assignment
     - leading_newlines_in_multiline_strings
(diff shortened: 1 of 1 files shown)


zoechi:
Related to #30

analysis_options.yaml
```diff
@@ -1,2 +1,215 @@
+# Updated according to
+# https://dart-lang.github.io/linter/lints/index.html
+# at 2020-08-10
+
 include: package:pedantic/analysis_options.yaml
 
+analyzer:
+  enable-experiment:
+    - non-nullable
```
tiholic, line 9: Can we disable this experiment?
It doesn't allow returning null values from methods.

tiholic, line 9: Can we disable this experiment?
It doesn't allow returning null values from methods.

analysis_options.yaml
```diff
@@ -1,2 +1,215 @@
...
+# annoying when introducing a class to be used as interface that has only one member
+#    - one_member_abstracts
+    - only_throw_errors
+    - overridden_fields
+    - package_api_docs
+    - package_prefixed_library_names
+    - parameter_assignments        # personal preference
+    - prefer_adjacent_string_concatenation
+    - prefer_asserts_in_initializer_lists
+#    - prefer_asserts_with_message # personal preference
```
tiholic, line 137: We can enable this

analysis_options.yaml
```diff
@@ -1,2 +1,215 @@
...
+    - prefer_is_not_empty
+    - prefer_is_not_operator
+    - prefer_iterable_whereType
+    - prefer_mixin
+    - prefer_null_aware_operators
+    - prefer_single_quotes
+    - prefer_spread_collections
+    - prefer_typing_uninitialized_variables
+    - provide_deprecation_message
+#    - public_member_api_docs      # personal preference
```
tiholic, line 173: Let's enable this too

analysis_options.yaml
```diff
@@ -1,2 +1,215 @@
...
+    - recursive_getters
+    - sized_box_for_whitespace
+    - slash_for_doc_comments
+    - sort_child_properties_last
+#      - sort_constructors_first           # personal preference
+#      - sort_unnamed_constructors_first   # personal preference
```
tiholic, line 179: We can enable these 2 settings: `sort_constructors_first` and `sort_unnamed_constructors_first`

analysis_options.yaml
```diff
@@ -1,2 +1,215 @@
...
+    - cascade_invocations
+    - constant_identifier_names
+    - curly_braces_in_flow_control_structures # personal preference
+    - directives_ordering
+    - do_not_use_environment # depends on project requirements
+    - empty_catches
+    - empty_constructor_bodies
+    - exhaustive_cases
+    - file_names
+#    - flutter_style_todos         # I find the required ":" annoying and pointless
```
tiholic, line 114: Can be ignored!

analysis_options.yaml
```diff
@@ -1,2 +1,215 @@
...
+    - implementation_imports
+    - join_return_with_assignment
+    - leading_newlines_in_multiline_strings
+    - library_names
+    - library_prefixes
+    - lines_longer_than_80_chars
+# Too annoying when long strings with interpolations are broken into multiple lines which we do often
+#    - missing_whitespace_between_adjacent_strings
```
tiholic, line 122: @zoechi isn't it helpful to avoid missing spaces?

analysis_options.yaml
```diff
@@ -1,2 +1,215 @@
...
+    - no_duplicate_case_values
+    - no_logic_in_create_state
+    - prefer_relative_imports      # personal preference
+    - prefer_void_to_null
+    - test_types_in_equals
+    - throw_in_finally
+    - unnecessary_statements
+    - unrelated_type_equality_checks
+    - unsafe_html
+    - use_key_in_widget_constructors
```
tiholic, line 59: Will flutter specific rules be valid here?

tiholic (commented):
My 7 cents.
Everything else should be good from what I can see.

analysis_options.yaml
zoechi: definitely, that's not available for Flutter yet anyway

analysis_options.yaml
zoechi: I find it a bit noisy but I'm fine mit whatever the others want.

analysis_options.yaml
zoechi: Ok, that's ambitious. I hope this does not lead to lots of low quality comments.
I prefer no comment over one that doesn't add real value even for public members.

analysis_options.yaml
zoechi: I'll check again.
I remember this warning about code that's actually fine.

analysis_options.yaml
zoechi: As far as I know they don't interfere but won't have any effect.
I'll remove it.

analysis_options.yaml
tiholic: @zoechi I think this will help us in identifying the exposed members without having to deliberately browse through the code.

analysis_options.yaml
tiholic: I agree, lets keep this disabled.

analysis_options.yaml
tiholic: 5fcf63eb4c20ecf7a2363430688dfa6ee977f093

analysis_options.yaml
tiholic: I'm enabling this for now as this wouldn't affect current source in anyway, can be disabled later if it doesn't fit

c59284f2906efdd2cf1c6bc7a14125db33d5b37f

analysis_options.yaml
tiholic: Not enabling this at the moment

analysis_options.yaml
tiholic: Removed both `unsafe_html` and `use_key_in_widget_constructors`

analysis_options.yaml
tiholic: Ignoring this for now. We can track and decide if we'd need this later on.

zoechi:
Superseded by #49

PR #38, opened 2020-08-25
realtime authcallback

tiholic:
Support for `authCallback` Realtime operations

### Current status 
**Android**
- [x] Realtime connection
- [x] Realtime publish
- [ ] in-band re-authentication

**iOS**
- [x] Realtime connection
- [x] Realtime publish
- [ ] in-band re-authentication

### How it works
1. Realtime instance will be created with `authCallback` in `ClientOptions`
2. `Realtime#connect` will be triggered from user code
3. ably library code calls a proxy `authCallback` on platform code and that will immediately return `null`, but will make a MethodChannel call (with name `realtimeAuthCallback`) to dart side for token data.
4. dart side receives the `MethodChannel` call and triggers authCallback to return the value to platform code.
platform code will set this response token data in temporary storage.
5. Trigger from `#2` will receive an error with code `80019` which will suppress the error and will check a boolean on dart side if the authentication update was complete and then retry publishing again.
6. Now, the platform code will return with the token data stored temporarily in #5 above for next `authCallback` request from ably library.

#### On token expiry
On token expiry, the connection state listeners will be triggered with connection state set to `disconnected` and with reason as Error with code `80019`. As soon as this event is encountered, `realtimeAuthCallback` method channel api will be triggered from platform side and `Realtime#connect` will be called immediately.

A similar flow occurs in case of publishing to a channel or listening on a channel.

### DIFF
--- a/lib/src/impl/realtime/channels.dart
+++ b/lib/src/impl/realtime/channels.dart
@@ -8,9 +8,8 @@ import 'package:flutter/services.dart';
 
 import '../platform_object.dart';
 
-
-class RealtimePlatformChannel extends PlatformObject implements spec.RealtimeChannel{
-
+class RealtimePlatformChannel extends PlatformObject
+    implements spec.RealtimeChannel {
   @override
   spec.AblyBase ably;
 
@@ -23,44 +22,42 @@ class RealtimePlatformChannel extends PlatformObject implements spec.RealtimeCha
   @override
   spec.RealtimePresence presence;
 
-  RealtimePlatformChannel(this.ably, this.name, this.options): super() {
-    this.handle;  //proactively acquiring handle
-    this.state = spec.ChannelState.initialized;
-    this.on().listen((ChannelStateChange event) {
-      this.state = event.current;
+  RealtimePlatformChannel(this.ably, this.name, this.options) : super() {
+    state = spec.ChannelState.initialized;
+    on().listen((ChannelStateChange event) {
+      state = event.current;
     });
   }
 
-  Realtime get realtimePlatformObject => this.ably as Realtime;
+  Realtime get realtimePlatformObject => ably as Realtime;
 
   /// createPlatformInstance will return realtimePlatformObject's handle
   /// as that is what will be required in platforms end to find realtime instance
   /// and send message to channel
   @override
-  Future<int> createPlatformInstance() async => await realtimePlatformObject.handle;
+  Future<int> createPlatformInstance() async =>
+      await realtimePlatformObject.handle;
 
   @override
-  Future<spec.PaginatedResult<spec.Message>> history([spec.RealtimeHistoryParams params]) {
+  Future<spec.PaginatedResult<spec.Message>> history(
+      [spec.RealtimeHistoryParams params]) {
     // TODO: implement history
     return null;
   }
 
-  Map<String,dynamic> __payload;
-  Map<String, dynamic> get _payload => __payload ??=
-  {
-    "channel": name,
-    if(options != null) "options": options
-  };
+  Map<String, dynamic> __payload;
+
+  Map<String, dynamic> get _payload =>
+      __payload ??= {'channel': name, if (options != null) 'options': options};
 
   @override
-  Future<void> publish({
-    spec.Message message,
-    List<spec.Message> messages,
-    String name,
-    dynamic data
-  }) async {
-    bool hasAuthCallback = this.ably.options.authCallback!=null;
-    while (hasAuthCallback && this.realtimePlatformObject.authCallbackInProgress) {
+  Future<void> publish(
+      {spec.Message message,
+      List<spec.Message> messages,
+      String name,
+      dynamic data}) async {
+    var hasAuthCallback = ably.options.authCallback != null;
+    while (hasAuthCallback && realtimePlatformObject.authCallbackInProgress) {
       await Future.delayed(Duration(milliseconds: 100));
     }
     try {
@@ -81,8 +78,9 @@ class RealtimePlatformChannel extends PlatformObject implements spec.RealtimeCha
         'messages': messages,
       });
     } on PlatformException catch (pe) {
-      if (hasAuthCallback && pe.code == ErrorCodes.authCallbackFailure.toString()) {
-        this.realtimePlatformObject.authCallbackInProgress = true;
+      if (hasAuthCallback &&
+          pe.code == ErrorCodes.authCallbackFailure.toString()) {
+        realtimePlatformObject.authCallbackInProgress = true;
         await publish(name: name, data: data);
       } else {
         throw spec.AblyException(pe.code, pe.message, pe.details);
@@ -108,7 +106,7 @@ class RealtimePlatformChannel extends PlatformObject implements spec.RealtimeCha
   @override
   Future<void> attach() async {
     try {
-      await this.invoke(PlatformMethod.attachRealtimeChannel, _payload);
+      await invoke(PlatformMethod.attachRealtimeChannel, _payload);
     } on PlatformException catch (pe) {
       throw spec.AblyException(pe.code, pe.message, pe.details);
     }
@@ -117,7 +115,7 @@ class RealtimePlatformChannel extends PlatformObject implements spec.RealtimeCha
   @override
   Future<void> detach() async {
     try {
-      await this.invoke(PlatformMethod.detachRealtimeChannel, _payload);
+      await invoke(PlatformMethod.detachRealtimeChannel, _payload);
     } on PlatformException catch (pe) {
       throw spec.AblyException(pe.code, pe.message, pe.details);
     }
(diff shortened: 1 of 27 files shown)


lib/src/impl/realtime/channels.dart
```diff
@@ -59,6 +59,10 @@ class RealtimePlatformChannel extends PlatformObject implements spec.RealtimeCha
     String name,
     dynamic data
   }) async {
+    bool hasAuthCallback = this.ably.options.authCallback!=null;
+    while (hasAuthCallback && this.realtimePlatformObject.authCallbackInProgress) {
```
zoechi, line 63: This busy-waiting looks really bad.
I haven't found `hasAuthCallback` being changed anywhere after the while loop, so I assume the while loop is only about `this.realtimePlatformObject.authCallbackInProgress`, therefore I'd write this as
```dart
if(hasAuthCallback) {
  while(this.realtimePlatformObject.authCallbackInProgress) {
    ...
  }
}
```
also instead of busy-waiting 
`realtimePlatformObject` should return a `Future` that completes when no authCallback is in progress anymore.

lib/src/impl/realtime/channels.dart
```diff
@@ -68,7 +72,12 @@ class RealtimePlatformChannel extends PlatformObject implements spec.RealtimeCha
         if(data!=null) "data": data
       });
     } on PlatformException catch (pe) {
-      throw spec.AblyException(pe.code, pe.message, pe.details);
+      if (hasAuthCallback && pe.code == ErrorCodes.authCallbackFailure.toString()) {
```
zoechi, line 75: Instead of using `toString()` everywhere, `ErrorCodes.xxx` should all be strings instead of integers.

lib/src/impl/realtime/connection.dart
```diff
@@ -15,6 +15,10 @@ class ConnectionPlatformObject extends PlatformObject implements Connection {
     this.handle;  //proactively acquiring handle
     this.state = ConnectionState.initialized;
     this.on().listen((ConnectionStateChange event) {
+      if (event.reason?.code == ErrorCodes.authCallbackFailure) {
```
zoechi, line 18: I see, here `ErrorCodes` is used as `int`. Is there a way to unify this to `int` or `String`?

lib/src/impl/realtime/realtime.dart
```diff
@@ -48,13 +53,25 @@ class Realtime extends PlatformObject implements spec.RealtimeInterface {
...
   @override
   Future<void> close() async => await invoke(PlatformMethod.closeRealtime);
 
+  bool authCallbackInProgress = false;
+
   @override
-  Future<void> connect() async => await invoke(PlatformMethod.connectRealtime);
+  Future<void> connect() async {
+    bool hasAuthCallback = this.options.authCallback!=null;
+    while (hasAuthCallback && authCallbackInProgress) {
```
zoechi, line 66: There really shouldn't be a need for this kind of busy-waiting, like ever.

lib/src/method_call_handler.dart
```diff
@@ -25,4 +27,20 @@ class AblyMethodCallHandler{
     return callbackResponse;
   }
 
+  bool realtimeAuthInProgress = false;
```
zoechi, line 30: I think this should be a `Future` that is returned to all calls if not `null` instead of a boolean flag.
This way callers can wait for the callback to complete.

lib/src/method_call_handler.dart
```diff
@@ -25,4 +27,20 @@ class AblyMethodCallHandler{
...
+  onRealtimeAuthCallback(AblyMessage message) async {
```
zoechi, line 31: Public members should have proper return types.

zoechi (changes requested):

lib/src/impl/realtime/channels.dart
tiholic: @zoechi Yes, this part needs to be changed completely. I'll try using completers if that fit well here.

lib/src/impl/realtime/channels.dart
tiholic: in PlatformException, it will be a string but in `ErrorInfo` it is an integer. This codes are exposed as public and so is ErrorInfo. I think it is better to have them as int.

lib/src/impl/realtime/connection.dart
tiholic: @zoechi Not sure! As per the spec, `ErrorInfo#code` is supposed to be int, and `PlatformException#code` is of type string. Not sure if we can unify them both.

@QuintinWillison Any thoughts?

lib/src/impl/realtime/channels.dart
zoechi: Ok, I just wanted to be sure this is intentional.

zoechi: Ok, lets resolve here. The discussion is open in another place already.

lib/src/method_call_handler.dart
tiholic: @zoechi Caller here is Android/iOS platform which would take no action when `null` is returned. And returning the same future will return same results to platform side which will have undesirable effects on the platform side.

zoechi (changes requested):

lib/src/impl/realtime/connection.dart
QuintinWillison: I'm not quite sure what's being proposed, by from my bird's eye view I think it's got to be lowest common denominator, surely? You can safely represent an `int` in a `String`, but not vice-versa.

zoechi: I'd then suggest we leave it as it is, when there is no concrete reason to change it.
I just wanted to be sure nothing was missed.

lib/src/impl/realtime/channels.dart
tiholic: @zoechi addressed 34417e964349e49ba2071e2ae93316ecf25a8982

lib/src/impl/realtime/realtime.dart
tiholic: addressed: 34417e964349e49ba2071e2ae93316ecf25a8982

lib/src/method_call_handler.dart
```diff
@@ -15,30 +15,26 @@ class AblyMethodCallHandler {
...
       return null;
     }
     realtimeAuthInProgress = true;
     ably.TokenParams tokenParams = message.message as ably.TokenParams;
     ably.Realtime realtime = ably.realtimeInstances[message.handle];
     Object callbackResponse = await realtime.options.authCallback(tokenParams);
-    Future.delayed(Duration.zero, (){
-      realtime.authUpdateComplete();
-    });
+    Future.delayed(Duration.zero, () => realtime.authUpdateComplete());
```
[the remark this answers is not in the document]
tiholic: Fixed

lib/src/method_call_handler.dart
```diff
@@ -15,30 +15,26 @@ class AblyMethodCallHandler {
...
 
-  Future<dynamic> onAuthCallback(AblyMessage message) async {
+  Future<Object> onAuthCallback(AblyMessage message) async {
     ably.TokenParams tokenParams = message.message as ably.TokenParams;
     ably.Rest rest = ably.restInstances[message.handle];
     final callbackResponse = await rest.options.authCallback(tokenParams);
-    Future.delayed(Duration.zero, () {
-      rest.channels.all.forEach((c) => c.authUpdateComplete());
-    });
+    Future.delayed(Duration.zero, () => rest.authUpdateComplete());
```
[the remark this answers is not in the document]
tiholic: Fixed

lib/src/method_call_handler.dart
tiholic: Done, handled

lib/src/impl/rest/rest.dart
```diff
@@ -9,36 +9,40 @@ import 'channels.dart';
...
-    final String key
-  }) :
-      assert(options!=null || key!=null),
-      this.options = (options==null)?ClientOptions.fromKey(key):options,
-      super()
-  {
-    this.channels = RestPlatformChannels(this);
+class Rest extends PlatformObject
+    implements spec.RestInterface<RestPlatformChannels> {
+  Rest({ClientOptions options, final String key})
```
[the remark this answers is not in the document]
tiholic: @zoechi both are not mandatory, either one is required!

lib/src/method_call_handler.dart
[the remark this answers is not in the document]
tiholic: Done

lib/src/method_call_handler.dart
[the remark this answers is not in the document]
tiholic: Done

lib/src/impl/rest/rest.dart
```diff
@@ -9,36 +9,40 @@ import 'channels.dart';
...
-      AblyMessage(options)
-    );
+    var handle = await invokeRaw<int>(
+        PlatformMethod.createRestWithOptions, AblyMessage(options));
     restInstances[handle] = this;
     return handle;
   }
 
+  void authUpdateComplete() {
+    channels.all.forEach((c) => c.authUpdateComplete());
```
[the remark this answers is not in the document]
tiholic: @zoechi I think the reference comment is lost, can you please point exactly?

lib/src/impl/rest/rest.dart
```diff
@@ -9,36 +9,40 @@ import 'channels.dart';
...
-      assert(options!=null || key!=null),
-      this.options = (options==null)?ClientOptions.fromKey(key):options,
-      super()
-  {
-    this.channels = RestPlatformChannels(this);
+class Rest extends PlatformObject
+    implements spec.RestInterface<RestPlatformChannels> {
+  Rest({ClientOptions options, final String key})
+      : assert(options != null || key != null),
+        options = (options == null) ? ClientOptions.fromKey(key) : options,
```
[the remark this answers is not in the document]
tiholic: b85617edc2c162cb15c7fc520668898c37ddbeb0

zoechi (approved):

lib/src/method_call_handler.dart
```diff
@@ -9,17 +9,32 @@ class AblyMethodCallHandler {
...
+    var rest = ably.restInstances[message.handle];
     final callbackResponse = await rest.options.authCallback(tokenParams);
-    Future.delayed(Duration.zero, () {
-      rest.channels.all.forEach((c) => c.authUpdateComplete());
-    });
+    Future.delayed(Duration.zero, rest.authUpdateComplete);
+    return callbackResponse;
+  }
+
+  bool realtimeAuthInProgress = false;
```
zoechi, line 26: I missed that this should be private

zoechi (changes requested):

lib/src/impl/rest/rest.dart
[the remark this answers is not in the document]
zoechi: Sorry, I didn't look properly. 
It's an `||` in the `assert(...)` not a `&&`.

lib/src/impl/rest/rest.dart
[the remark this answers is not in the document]
zoechi: Sorry again. I was sloppy. 
Here it can't be changed in the other cases because the function is required to accept a parameter..

lib/src/method_call_handler.dart
tiholic: Addressed 0ab957b7fed08750fc0893be45b0f5a0f13878af

zoechi (approved):

lib/src/spec/realtime/realtime.dart
```diff
@@ -16,6 +16,6 @@ abstract class RealtimeInterface extends AblyBase {
   void close();
   void connect();
   final Connection connection;
-  RealtimeChannels channels;
+  C channels;
```
zoechi, line 19: I assume this should be `C get channels;` 
See also my comment in #33 `impl/rest/rest.dart`

zoechi (changes requested):

lib/src/spec/realtime/realtime.dart
tiholic: converted Channels and Connection to getters 9b2fc58aa8f37ae7d0e970599b689437ae58859e

from 33 => 68022946db52951e7223052bdfe67687d907b515

lib/src/impl/realtime/channels.dart
```diff
@@ -23,53 +24,114 @@ class RealtimePlatformChannel extends PlatformObject implements spec.RealtimeCha
...
-  Map<String,dynamic> __payload;
-  Map<String, dynamic> get _payload => __payload ??=
-  {
-    "channel": name,
-    if(options != null) "options": options
-  };
+  Map<String, dynamic> __payload;
+
+  Map<String, dynamic> get _payload =>
+      __payload ??= {'channel': name, if (options != null) 'options': options};
```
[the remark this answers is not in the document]
tiholic: c52463b5ab6df5f959a13a85bb095d7b9b33f2d5

zoechi (approved):

PR #39, opened 2020-08-26
Improve RestPlatformChannel publish

zoechi:
replaces #36

### DIFF
--- a/lib/src/impl/rest/channels.dart
+++ b/lib/src/impl/rest/channels.dart
@@ -1,5 +1,6 @@
 import 'dart:async';
-
+import 'dart:collection';
+import 'package:pedantic/pedantic.dart';
 import 'package:ably_flutter_plugin/ably.dart';
 import 'package:ably_flutter_plugin/src/impl/rest/rest.dart';
 import 'package:ably_flutter_plugin/src/spec/spec.dart' as spec;
@@ -7,8 +8,9 @@ import 'package:flutter/services.dart';
 
 import '../platform_object.dart';
 
-
-class RestPlatformChannel extends PlatformObject implements spec.RestChannel{
+class RestPlatformChannel extends PlatformObject implements spec.RestChannel {
+  static const String clientConfiguredAuthenticationProviderRequestFailed =
+      '80019';
 
   /// [Rest] instance
   @override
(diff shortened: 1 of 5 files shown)


lib/src/method_call_handler.dart
```diff
@@ -3,26 +3,23 @@ import 'package:ably_flutter_plugin/src/generated/platformconstants.dart';
...
     ably.TokenParams tokenParams = message.message as ably.TokenParams;
     ably.Rest rest = ably.restInstances[message.handle];
-    Object callbackResponse = await rest.options.authCallback(tokenParams);
-    Future.delayed(Duration.zero, (){
+    await rest.options.authCallback(tokenParams);
+    Future.delayed(Duration.zero, () {
       rest.channels.all.forEach((c) => c.authUpdateComplete());
     });
-    return callbackResponse;
+    return;
```
tiholic, line 23: @zoechi the returned `callbackResponse` will be returned back to platform. So, we'd need this return statement to be as is.

lib/src/impl/rest/channels.dart
```diff
@@ -1,14 +1,16 @@
...
 import 'package:ably_flutter_plugin/src/impl/rest/rest.dart';
 import 'package:ably_flutter_plugin/src/spec/spec.dart' as spec;
 import 'package:flutter/services.dart';
 
 import '../platform_object.dart';
 
-
-class RestPlatformChannel extends PlatformObject implements spec.RestChannel{
+class RestPlatformChannel extends PlatformObject implements spec.RestChannel {
+  static const String clientConfiguredAuthenticationProviderRequestFailed =
```
tiholic, line 12: @zoechi A new commit (e9e0939) has been cherry-picked from another latest branch into `feature/realtime-authcallback`.

You may rebase and use that constant

tiholic (changes requested):

lib/src/method_call_handler.dart
zoechi: The function should have a proper return type then.
The type annotation in the original code is `Object` (line 21), but `AuthCallback` is defined to return `dynamic`. 

I think I suggested to not use `dynamic` at all already (except a few exceptions where it's unavoidable).
I'll leave `dynamic` here anyway for consistency.

If we actually want to to consistently not use `dynamic`, we should create an issue for this and make a PR where we get rid of it everywhere.

lib/src/impl/rest/channels.dart
zoechi: `PlatformException` has `final String code`, but `ErrorCodes` has `static const int authCallbackFailure`
Does this type inconsistency `String`/`int` make any sense?

lib/src/method_call_handler.dart
tiholic: Sure, let's get rid of everything `dynamic`!

lib/src/impl/rest/channels.dart
tiholic: This discussion is continued either https://github.com/ably/ably-flutter/pull/38#discussion_r484375487 or https://github.com/ably/ably-flutter/pull/38#discussion_r484379978

tiholic (approved):

PR #41, opened 2020-08-27
Add integration tests

zoechi:
- report unhandled exceptions back to the driver test
- report `FlutterError`s back to the driver test
- make `lib/main.dart` and `TestDispatcher` reusable

- improve platform version and ably version tests
- improve appkey provision test

#### Tests failing in GitHub Actions on Android emulator only

I think https://github.com/flutter/flutter/issues/42433 is the cause.

When I split the tests so that there is only one or sometimes 2 tests run at once like in https://github.com/ably/ably-flutter/blob/feature/basic-integration-tests/test_integration/test_driver/rest_publish_test.dart (in contrary to all in one file https://github.com/ably/ably-flutter/blob/feature/basic-integration-tests/test_integration/test_driver/all_test.dart) it works but it's quite cumbersome.


zoechi:
- [ ] base on integration/stage-2

QuintinWillison (approved):

PR #48, opened 2020-09-29
Channel history & Paginated Result API for Rest

tiholic:
`Rest#Channel#Channel#history` will return `PaginatedResult`.
PaginatedResult supports methods: `hasNext()`, `isLast()`, `async next()` and `async first()` as per the [docs#TG](https://docs.ably.io/client-lib-development-guide/features/#TG1)

```dart
void getHistory([ably.RestHistoryParams params]) async {
    // getting channel history, by passing or omitting the optional params
    var result = await channel.history(params);

    var messages = result.items; //get messages
    var hasNextPage = result.hasNext(); //tells whether there are more results
    if(hasNextPage){    
      result = await result.next();  //fetches next page results
      messages = result.items;
    }
    if(!hasNextPage){
      result = await result.first();  //fetches first page results
      messages = result.items;
    }
}

// history with default params
getHistory();

// sorted and filtered history
getHistory(
  ably.RestHistoryParams(
    direction: 'forwards',
    limit: 10,
    start: DateTime.fromMillisecondsSinceEpoch(0),
    end: DateTime.now()
  )
);
```


zoechi (changes requested):

lib/src/spec/common.dart
```diff
@@ -398,17 +442,31 @@ abstract class EventEmitter<E, G> {
...
-  List<T> items();
-  Future<PaginatedResult<T>> first();
-  Future<PaginatedResult<T>> next();
-  Future<PaginatedResult<T>> current();
+/// PaginatedResult [TG1](https://docs.ably.io/client-lib-development-guide/features/#TG1)
+///
+/// A type that represents page results from a paginated query.
+/// The response is accompanied by metadata that indicates the
+/// relative queries available.
+abstract class PaginatedResultInterface<T> {
```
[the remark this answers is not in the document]
tiholic: We are exporting all the definitions in the final package `ably/ably.dart` which is leading to duplicate declarations.

lib/src/spec/common.dart
```diff
@@ -398,17 +442,31 @@ abstract class EventEmitter<E, G> {
...
+  /// returns true if there are further pages (TG6)
   bool hasNext();
+
+  /// returns true if this page is the last page (TG7)
   bool isLast();
 }
 
-abstract class HttpPaginatedResponse extends PaginatedResult<String> {
-  List<String> items();
+abstract class HttpPaginatedResponse extends PaginatedResultInterface<String> {
```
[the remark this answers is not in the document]
tiholic: [same as another comment] We are exporting all the definitions in the final package `ably/ably.dart` which is leading to duplicate declarations.

lib/src/spec/common.dart
```diff
@@ -260,37 +260,81 @@ class TokenRequest {
...
+      {DateTime start,
+      DateTime end,
+      String direction,
+      int limit,
+      this.untilAttach})
+      : super(start: start, end: end, direction: direction, limit: limit);
 }
 
-abstract class RestPresenceParams {
+class RestPresenceParams {
```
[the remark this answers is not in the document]
tiholic: This class will be enhanced to match the exact requirement while implementing `RestPresence`

lib/src/spec/common.dart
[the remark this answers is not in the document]
zoechi: But `Interface` is misleading then. `Base` might be a better extension then.

lib/src/spec/common.dart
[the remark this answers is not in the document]
zoechi: If classes are used as interfaces it's better to make the fields getters, and if required add matching setters on demand. 
Introducing fields in classes that might be extended is error prone. 

I'd prefer to have such classes "locked down" by default and open up on demand, instead of later trying to comb through all code to figure out if some of it actually needs to be fields, or needs to be writeable.

lib/src/spec/common.dart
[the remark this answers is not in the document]
zoechi: Do they have the same name in other implementations?

lib/src/impl/paginated_result.dart
```diff
@@ -0,0 +1,61 @@
...
+  /// instantiated by the codec. So the code invoking platform method
+  /// is bound to update this [_pageHandle]
+  ///
+  /// [PaginatedResult.fromAblyMessage] will act as a utility to update
+  /// this property. See [next] and [first] for usages
+  int _pageHandle;
+
+  /// items contain page of results
+  @override
+  List<T> items;
```
zoechi, line 22: This should probably be a private field with a public getter

zoechi (changes requested):

lib/src/spec/common.dart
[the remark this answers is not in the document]
tiholic: @zoechi I created these base classes (spec) so that even if we migrate from a plugin based approach to a pure dart package, we should be able to do it smoothly.

And currently: whenever it is required to implement any of these classes, renaming the class in spec with `XXXImplementation` and then implementing a platform object for the same is serving the purpose.

What do you think would be the best approach / suggestion for modification in relation to above comments?

lib/src/impl/paginated_result.dart
tiholic: @zoechi addressed at 5df822ddbe75e9bd504bb69471273002d38bedde

ios/Classes/codec/AblyFlutterWriter.m
```diff
@@ -27,6 +27,8 @@ + (UInt8) getType:(id)value{
         return messageCodecType;
     }else if([value isKindOfClass:[ARTTokenParams class]]){
         return tokenParamsCodecType;
+    }else if([value isKindOfClass:[ARTPaginatedResult class]]){
```
QuintinWillison, line 30: I'm not going to block this pull request landing on this, but I am interested to find out why our linter didn't pick up the fact that there's no space before the `else`, nor after the `if`... nor before the final `{` actually! 🤔 

It should be configured so this fails the build.

ios/Classes/codec/AblyPlatformConstants.h
```diff
@@ -164,3 +170,21 @@ extern NSString *const TxMessage_data;
 extern NSString *const TxMessage_name;
 extern NSString *const TxMessage_extras;
 @end
+
+@interface TxPaginatedResult : NSObject
```
QuintinWillison, line 174: These `@interface` declarations and the corresponding `@implementation` blocks do nothing. I would prefer that we remove them as the inference is that they add value, but they don't.

example/lib/main.dart
```diff
@@ -364,6 +365,31 @@ class _MyAppState extends State<MyApp> {
     child: Text('Publish'),
   );
 
+  Widget getRestChannelHistory() => FlatButton(
```
QuintinWillison, line 368: For what it's worth, I think this code could benefit from some commentary. If nothing else for each of the three cases in the `if` statement. It's a little bit opaque on a brief scan. I'm finding myself puzzling over why `_restHistory` could be `null`, but also why you use `?`/`??` at the start but a direct comparison to `== null` later.

This is meant to be example code so we cannot have too much commentary. Many developers taking a look will be time poor so need their hands held.

lib/src/generated/platformconstants.dart
```diff
@@ -161,3 +167,21 @@ class TxMessage {
     static const String name = "name";
     static const String extras = "extras";
 }
+
+class TxPaginatedResult {
```
QuintinWillison, line 171: [avoid_classes_with_only_static_members ](https://dart-lang.github.io/linter/lints/avoid_classes_with_only_static_members.html) 🤔

lib/src/spec/common.dart
```diff
@@ -261,37 +261,81 @@ class TokenRequest {
...
+  /// end must be equal to or greater than start and is unaffected by the request direction
+  ///
+  /// RLS2b1
+  final DateTime end;
+
+  /// Sorting history backwards or forwards
+  ///
+  /// if omitted the direction defaults to the REST API default (backwards)
+  /// RLS2b2
+  final String direction;
```
QuintinWillison, line 282: The commentary here should specify really clearly what values can be set. Though, surely an enum would be more appropriate? This is a good example of somewhere we could do better in terms of using the compiler.

QuintinWillison (approved):
I've not looked at this pull request at the level of depth I would like to, but I'm short on time and we need to get a release out.

None of the comments I've made need block this pull request from landing so, @tiholic, please don't add any more commits to the underlying branch. For each of my comments, could you please either:

1. Create an issue to defer the work to fix it until later.
2. Create a pull request to fix it now if you have time (in which case that pull request's underlying branch should branch from the latest branch you've been working to allow the entire backlog to fold down correctly)

And then add a comment to the conversation to point towards which approach you'll take. I'll resolve the conversations later.

QuintinWillison:
Conflicts. 🤔  I'm resolving them now.

ios/Classes/codec/AblyFlutterWriter.m
tiholic: Lint was fixed in another PR #67  dedicated purely for linting.
Even though it majorly focuses on dart code linting, I've also applied formatting to java source and auto indent to ObjC source.

lib/src/generated/platformconstants.dart
tiholic: We disabled this lint rule.
https://github.com/ably/ably-flutter/blob/feature/lint-dart/analysis_options.yaml#L77

lib/src/spec/common.dart
tiholic: Yes, I too had a thought on this. We can change this to an enum with 2 values and handle it in codec.

https://github.com/ably/ably-flutter/issues/83

tiholic: Commentary was later improved in another dedicated PR for docs #69

ios/Classes/codec/AblyFlutterWriter.m
QuintinWillison: Thanks. Sorry, I should have perhaps looked ahead in order not to add this noise here. 😅

lib/src/generated/platformconstants.dart
QuintinWillison: ok, fair enough. Thanks. 😁

lib/src/spec/common.dart
QuintinWillison: Perfect, thanks. 😍

PR #57, opened 2020-11-02
Fixes for GitHub actions

tiholic:
Fix github actions
1. fix iOS device UUID query
2. remove duplicate tests
3. Wakelock on devices to run without the device going to sleep (makes sense only for physical devices)


QuintinWillison (approved):
Very difficult to review without a lot of research, but I see this fixes the workflow so let's land it.

PR #59, opened 2020-11-04
Pre-release dev.1

tiholic:


QuintinWillison (changes requested):
This PR should not be landed to `main` until we're sure that all other stage 1 PRs have been merged.

Regardless, I notice your commit includes changes to `pubspec.lock` - I would definitely not expect changes to that file within the scope of a release PR. It should be version "bump" for this library only plus change log.

In terms of the change log entry, it is helpful for everyone to get review input on that early, so thanks for getting this opened.

tiholic:
@QuintinWillison removed the changes in `pubspec.lock`

PR #62, opened 2020-11-09
Feature/initial release fixes

### LINKED ISSUE #56: Inconsistent ENUM Ordering between platform SDK's
The enum ordering is inconsistent between [`ably-java's ChannelEvent`](https://github.com/ably/ably-java/blob/5cf9e0193c4dff0e48770f20db9756491c7a7e2c/lib/src/main/java/io/ably/lib/realtime/ChannelEvent.java#L6) and [`ably-cocoa's ARTChannelEvent`](https://github.com/ably/ably-cocoa/blob/33f4a96c3540f1cfcac225d1df79428b1d7f0704/Source/ARTTypes.m#L212) (Same applies to `ChannelState` too)

When it comes to `ably-flutter`, these enum’s will be encoded from platform side as the ordinal value of these enums (see: [encoding in cocoa](https://github.com/ably/ably-flutter/blob/main/ios/Classes/codec/AblyFlutterWriter.m#L64) & [encoding in java](https://github.com/ably/ably-flutter/blob/main/android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java#L105)) and dart side will decode it using [ChannelState in dart](https://github.com/ably/ably-flutter/blob/main/lib/src/spec/enums.dart#L1) from the same index. This leads to decoding the values wrongly when it comes to android as dart ordering aligns with cocoa library which is same as mentioned in [line 184 in Interface Definition](https://docs.ably.io/client-lib-development-guide/features/#idl)

@QuintinWillison @paddybyers

### LINKED ISSUE #58: ConnectionStateChange#previous is none for the first event from ably-cocoa
When a realtime instance is created, it is supposed to have `initialized` as connection state so that the first ConnectionChange Event must emit a `ConnectionStateChange` object with `previous` attribute set as `initialized`

This works as expected in Android, but `ConnectionStateChange#previous` from ios is `null`

tiholic:
1. Fixes #58 : wrong `ConnectionStateChange#previous` and `ChannelStateChange#previous` values from iOS
2. Fixes #56 : ENUM Ordering issue - addressed by converting ENUM values to constant strings on both platforms and then passing them to flutter side

Updating Podfile.lock for example app and integration tests to upgrade ably-cocoa to 1.2.3

### DIFF
--- a/android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
+++ b/android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
@@ -14,7 +14,11 @@ import java.util.function.Consumer;
 
 import io.ably.flutter.plugin.generated.PlatformConstants;
 import io.ably.flutter.plugin.types.PlatformClientOptions;
+import io.ably.lib.realtime.ChannelEvent;
+import io.ably.lib.realtime.ChannelState;
 import io.ably.lib.realtime.ChannelStateListener;
+import io.ably.lib.realtime.ConnectionEvent;
+import io.ably.lib.realtime.ConnectionState;
 import io.ably.lib.realtime.ConnectionStateListener;
 import io.ably.lib.rest.Auth;
 import io.ably.lib.rest.Auth.TokenDetails;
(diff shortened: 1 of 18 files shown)


QuintinWillison (approved):
The repetition with the code generation gets me every time, but on a briefer-than-usual scan thru this looks good. Thanks!

PR #66, opened 2020-11-16
Feature/pub feedback

tiholic:
Fixing issues reported by pub

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [1.0.0+dev.2](https://github.com/ably/ably-flutter/tree/1.0.0-dev.2)
+
+Conforming to pub recommendations to improvise pub scoring
+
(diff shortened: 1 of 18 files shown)


QuintinWillison (approved):
Thanks, @tiholic.

It would be good, if you can find the time, to perhaps add some of these learnings to the developer notes document we have in this repository. I'm thinking, in particular, around how "internal" files need to be imported using relative folder syntax whereas files outside of the core package still need to use the `package:`-prefixed form.

Alternatively, could not existing (or new) tooling in our build have picked this up automatically? Is there a rule to spot this somewhere?

QuintinWillison:
I'll wait until checks have completed before landing this PR and doing the release.

tiholic:
@QuintinWillison the relative imports is addressed in linting PR. And rest assured we will not be facing this issue anymore if we release after all CI checks are passed.

PR #69, opened 2020-11-26
Feature/docstrings

tiholic:
1. Adding docstrings to all public members

2. Rename few abstract methods as `xxxInterface` and name the implementations as `xxx`
> ex: `abstract RestChannels` -> `abstract RestChannelsInterface` and `RestChannelPlatformObject` -> `RestChannel`

This is done as a prerequisite to improvising the members of library that are supposed to be available via library import ( `import 'package:ably_flutter_plugin/ably_flutter_plugin.dart';`)

> _No logical/API changes were made as a part of this PR - and no changes to integration tests_

### DIFF
--- a/.github/workflows/flutter_integration.yaml
+++ b/.github/workflows/flutter_integration.yaml
@@ -30,7 +30,7 @@ jobs:
         with:
           channel: "stable"
       - name: "Run Flutter Driver tests"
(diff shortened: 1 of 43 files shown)


QuintinWillison (approved):
This is a significant step forward. Thanks, @tiholic! 😁

PR #89, opened 2021-02-16
Preview Release - 1.2.0 preview.1

QuintinWillison:

PR #90, opened 2021-02-17
Rename Package to remove the _plugin suffix

QuintinWillison:
We'll land this pull request before #89.

As [discussed in Slack](https://ably-real-time.slack.com/archives/C0178QGC093/p1613505401090500), I'm also going to create a private fork of this repository so we can push a discontinuation to the [current package](https://pub.dev/packages/ably_flutter_plugin). Clearly that falls outside of the scope of this pull request.


lib/src/spec/common.dart
```diff
@@ -1,4 +1,4 @@
-import 'package:ably_flutter_plugin/src/spec/spec.dart';
+import 'package:ably_flutter/src/spec/spec.dart';
```
tiholic, line 1: Not sure how I missed this earlier, but linting should have picked up [prefer_relative_imports](https://dart-lang.github.io/linter/lints/prefer_relative_imports.html)

tiholic: This can be ignored in this PR and I can take it up later. I'll be checking regularly for any linting pointers for every commit.

tiholic: Also, in upcoming PR's we are capturing these lint pointers in workflows: https://github.com/ably/ably-flutter/blob/b917e0e86fdd5d575b3ab2b57d0646580d00992f/.github/workflows/flutter_integration.yaml

tiholic: I just checked, it was fixed in another branch later: https://github.com/ably/ably-flutter/blob/feature/realtime-presence-get-history/lib/src/spec/common.dart

tiholic (approved):
LGTM

PR #97, opened 2021-03-19
Add docs workflow

QuintinWillison:
Exploratory, work in progress. 😁

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -46,7 +46,7 @@ jobs:
           route: POST /repos/:repository/deployments
           repository: ${{ github.repository }}
           ref: ${{ fromJson(steps.get_pull_request_information.outputs.data).head.ref }}
-          environment: sdk-static
+          environment: sdk-static/pull/${{ github.event.pull_request.number }}
           auto_merge: false
           required_contexts: '[]'
         env:
@@ -61,7 +61,7 @@ jobs:
           route: POST /repos/:repository/deployments/:deployment/statuses
           repository: ${{ github.repository }}
           deployment: ${{ fromJson(steps.create_deployment.outputs.data).id }}
-          environment: sdk-static
+          environment: sdk-static/pull/${{ github.event.pull_request.number }}
           environment_url: https://sdk.ably.com/builds/ably/ably-flutter/pull/${{ github.event.pull_request.number }}/docs/api-reference/
           log_url: https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}
           state: in_progress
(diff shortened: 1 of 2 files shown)


QuintinWillison:
Thanks @lmars for debugging and fixing the AWS S3 piece! 🎉 

In terms of the bits I had worked out, thinking about it further in respect of how we're using Github's exposed [default environment variables]((https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables))...

An alternative approach for detecting whether we've been triggered from a pull request or from a push to `main` is to first inspect `GITHUB_EVENT_NAME`. If we find `pull_request` there, then we know that the event data must include a pull request number.

I might push another commit to implement that as I think it's cleaner. Relying on `jq` returning us `null` when it can't find a key in the event json feels icky. FYI, @owenpearson.

QuintinWillison:
This is our first end-to-end proof of life in respect of pushing content out to the SDK Team S3 bucket.

I think the only piece that is missing is how we might create a deployment for when the docs are build from the `main` branch... but I've not fully got my head around where/how that would be applicable, and I've not seen examples of this being done elsewhere (though I've not searched _very_ hard).

lmars (approved):
🚀

owenpearson (approved):
awesome 💯

QuintinWillison:
Back to the drawing board in respect of `environment` naming, now that I've seen the results of adding #99 to the mix.

More detail [in Slack](https://ably-real-time.slack.com/archives/C0178QGC093/p1616498074026600).

QuintinWillison:
Still some work to do on this pull request... next up is to create a deployment when run against the `main` branch. That's more difficult to test, clearly, given we're on a pull request against a feature branch while iterating this but I have a few ideas there. I'll carry on tomorrow on this so shall leave this PR in Draft state for now.

QuintinWillison:
ok, I think I've got this to a happy place now. Yet to be proven on the `main` branch, but we'll find out about that when this pull request lands! 🤞 

@owenpearson and I have put a fair bit of thought into environment names:

- `staging/pull/97/dartdoc` for a pull request
- `staging/main/dartdoc` for the `main` branch

Where:

- the `staging` prefix is the "where" (where `sdk.ably.com` is our https-accessible staging site)
- the middle bit (`pull/97` or `main`) is the "context / what"
- the `dartdoc` suffix is the "which bit"

We're keen to keep the suffixes short, as these environment names are visible from the repository home page (right hand side column). e.g.:

Screenshot 2021-03-24 at 10 53 23

We've also decided to keep a 1:1 relationship between the environment name (beyond the `staging` prefix) with the remote path we serve at `sdk.ably.com`. And, of course, this is all also aligned with GitHub-native paths also. Hopefully keeps everything straight and logical.

So, sorry @lmars and @owenpearson, but I need to re-request review from you both please as a lot's changed since you approved this PR in a much earlier state.

lmars (approved):

owenpearson (approved):
LGTM

QuintinWillison:
Thanks for the re-approves.

@tiholic and @tbedford, I'm keen to respond to any feedback you may have after this pull request has landed as we definitely want to iterate further if there's anything you spot that's awry!

Right now, however, I'm going to land this so we can see if the deployment works on the `main` branch too. 🤞

PR #99, opened 2021-03-23
Refine docs workflow

QuintinWillison:
The goals of this PR include:

1. Understanding and asserting what should happen in this workflow when executed against the `main` branch (see [this comment](https://github.com/ably/ably-flutter/pull/97#issuecomment-804717512))
2. Squeezing maximum juice out of the deployment API (e.g. use of `description` in requests) to ensure we're presenting in the GitHub UI in the best possible way


QuintinWillison:
As far as I can tell these `description` values are not surfaced anywhere helpful in the GitHub user interface so I'm going to close this PR.

Issue #82, opened 2021-02-15
Migrate to null safety


tiholic:
Refer to the migration guide: https://dart.dev/null-safety/migration-guide

_Note: null safety is supported in dart SDK > 2.12_

heshanka:
Is there any progress about this issue?

QuintinWillison:
Hello @heshanka - we definitely feel that migration to null safety will bring benefits to developers using this plugin so this issue is definitely towards the top of our backlog, however we can't currently promise when it'll be delivered as there are still a few fundamental features which we're working on ahead of this.

PR #123, opened 2021-05-24
Add support message extras and its extensions [TM2i]

### LINKED ISSUE #11: [TM2i] eventExtras for realtime publish
As per this thread https://github.com/ably/ably-flutter/pull/5#discussion_r419158894 

Development guidelines reference: https://docs.ably.io/client-lib-development-guide/features/#TM2i

tiholic:
- Simplify for `MessageExtras`
- Add support for `DeltaExtras`
    - readonly extension if delta compression is enabled on the channel

fixes #11

### DIFF
--- a/android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
+++ b/android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
@@ -652,8 +652,8 @@ public class AblyMessageCodec extends StandardMessageCodec {
 
   private Map<String, Object> encodeChannelMessageExtras(MessageExtras c) {
     if (c == null) return null;
-    final HashMap<String, Object> jsonMap = new HashMap<>();
-    writeValueToJson(jsonMap, PlatformConstants.TxMessageExtras.extras, c.asJsonObject());
+    final HashMap<String, Object> jsonMap =
+        new Gson().<HashMap<String, Object>>fromJson(c.asJsonObject().toString(), HashMap.class);
     DeltaExtras deltaExtras = c.getDelta();
     if (deltaExtras != null) {
       final HashMap<String, Object> deltaJson = new HashMap<>();
--- a/android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
+++ b/android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
@@ -26,8 +26,10 @@ import io.ably.lib.types.AblyException;
 import io.ably.lib.types.AsyncPaginatedResult;
 import io.ably.lib.types.ChannelOptions;
 import io.ably.lib.types.ClientOptions;
+import io.ably.lib.types.DeltaExtras;
 import io.ably.lib.types.ErrorInfo;
 import io.ably.lib.types.Message;
+import io.ably.lib.types.MessageExtras;
 import io.ably.lib.types.Param;
 import io.ably.lib.types.PresenceMessage;
 import io.flutter.plugin.common.StandardMessageCodec;
@@ -105,6 +107,8 @@ public class AblyMessageCodec extends StandardMessageCodec {
             new CodecPair<>(self::encodeErrorInfo, null));
         put(PlatformConstants.CodecTypes.messageData,
             new CodecPair<>(null, self::decodeChannelMessageData));
+        put(PlatformConstants.CodecTypes.messageExtras,
+            new CodecPair<>(self::encodeChannelMessageExtras, self::decodeChannelMessageExtras));
         put(PlatformConstants.CodecTypes.message,
             new CodecPair<>(self::encodeChannelMessage, self::decodeChannelMessage));
         put(PlatformConstants.CodecTypes.presenceMessage,
@@ -155,6 +159,8 @@ public class AblyMessageCodec extends StandardMessageCodec {
       return PlatformConstants.CodecTypes.channelStateChange;
     } else if (value instanceof PresenceMessage) {
       return PlatformConstants.CodecTypes.presenceMessage;
+    } else if (value instanceof MessageExtras) {
+      return PlatformConstants.CodecTypes.messageExtras;
     } else if (value instanceof Message) {
       return PlatformConstants.CodecTypes.message;
     }
@@ -444,6 +450,13 @@ public class AblyMessageCodec extends StandardMessageCodec {
     return decodeMessageData(jsonMap.get(PlatformConstants.TxMessage.data));
   }
 
+  private MessageExtras decodeChannelMessageExtras(Map<String, Object> jsonMap) {
+    if (jsonMap == null) return null;
+    Object extras = jsonMap.get(PlatformConstants.TxMessageExtras.extras);
+    final Gson gson = new Gson();
+    return new MessageExtras(gson.fromJson(gson.toJson(extras, Map.class), JsonObject.class));
+  }
+
   private Message decodeChannelMessage(Map<String, Object> jsonMap) {
     if (jsonMap == null) return null;
     final Message o = new Message();
@@ -452,10 +465,7 @@ public class AblyMessageCodec extends StandardMessageCodec {
     readValueFromJson(jsonMap, PlatformConstants.TxMessage.name, v -> o.name = (String) v);
     readValueFromJson(jsonMap, PlatformConstants.TxMessage.data, v -> o.data = decodeMessageData(v));
     readValueFromJson(jsonMap, PlatformConstants.TxMessage.encoding, v -> o.encoding = (String) v);
-    readValueFromJson(jsonMap, PlatformConstants.TxMessage.extras, v -> {
-      final Gson gson = new Gson();
-      o.extras = gson.fromJson(gson.toJson(v, Map.class), JsonObject.class);
-    });
+    readValueFromJson(jsonMap, PlatformConstants.TxMessage.extras, v -> o.extras = (MessageExtras) v);
     return o;
   }
 
@@ -637,6 +647,18 @@ public class AblyMessageCodec extends StandardMessageCodec {
     writeValueToJson(jsonMap, PlatformConstants.TxChannelStateChange.event, encodeChannelEvent(c.event));
     writeValueToJson(jsonMap, PlatformConstants.TxChannelStateChange.resumed, c.resumed);
     writeValueToJson(jsonMap, PlatformConstants.TxChannelStateChange.reason, encodeErrorInfo(c.reason));
+    return jsonMap;
+  }
+
+  private Map<String, Object> encodeChannelMessageExtras(MessageExtras c) {
+    if (c == null) return null;
+    final HashMap<String, Object> jsonMap = new HashMap<>();
+    writeValueToJson(jsonMap, PlatformConstants.TxMessageExtras.extras, c.asJsonObject());
+    DeltaExtras deltaExtras = c.getDelta();
+    final HashMap<String, Object> deltaJson = new HashMap<>();
+    writeValueToJson(deltaJson, PlatformConstants.TxDeltaExtras.format, deltaExtras.getFormat());
+    writeValueToJson(deltaJson, PlatformConstants.TxDeltaExtras.from, deltaExtras.getFrom());
+    writeValueToJson(jsonMap, PlatformConstants.TxMessageExtras.delta, deltaJson);
     return jsonMap;
   }
 
--- a/android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
+++ b/android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
@@ -72,6 +72,7 @@ public class AblyMessageCodec extends StandardMessageCodec {
   }
 
   private Map<Byte, CodecPair> codecMap;
+  private static final Gson gson = new Gson();
 
   AblyMessageCodec() {
     final AblyMessageCodec self = this;
@@ -187,9 +188,9 @@ public class AblyMessageCodec extends StandardMessageCodec {
 
   private void WriteJsonElement(ByteArrayOutputStream stream, JsonElement value) {
     if (value instanceof JsonObject) {
-      super.writeValue(stream, (new Gson()).fromJson(value, Map.class));
+      super.writeValue(stream, (gson).fromJson(value, Map.class));
     } else if (value instanceof JsonArray) {
-      super.writeValue(stream, (new Gson()).fromJson(value, ArrayList.class));
+      super.writeValue(stream, (gson).fromJson(value, ArrayList.class));
     }
   }
 
@@ -198,7 +199,6 @@ public class AblyMessageCodec extends StandardMessageCodec {
    * returns null if these 2 types are a no-match
    */
   static JsonElement readValueAsJsonElement(final Object object) {
-    final Gson gson = new Gson();
     if (object instanceof Map) {
       return gson.fromJson(gson.toJson(object, Map.class), JsonObject.class);
     } else if (object instanceof ArrayList) {
@@ -453,8 +453,10 @@ public class AblyMessageCodec extends StandardMessageCodec {
   private MessageExtras decodeChannelMessageExtras(Map<String, Object> jsonMap) {
     if (jsonMap == null) return null;
     Object extras = jsonMap.get(PlatformConstants.TxMessageExtras.extras);
-    final Gson gson = new Gson();
-    return new MessageExtras(gson.fromJson(gson.toJson(extras, Map.class), JsonObject.class));
+    // extras received from dart side could be a nested map with dynamic value types
+    // So, converting to a json string and then using that to create a JSONObject
+    String extrasJson = gson.toJson(extras, Map.class);
+    return new MessageExtras(gson.fromJson(extrasJson, JsonObject.class));
   }
 
   private Message decodeChannelMessage(Map<String, Object> jsonMap) {
@@ -653,7 +655,7 @@ public class AblyMessageCodec extends StandardMessageCodec {
   private Map<String, Object> encodeChannelMessageExtras(MessageExtras c) {
     if (c == null) return null;
     final HashMap<String, Object> jsonMap =
-        new Gson().<HashMap<String, Object>>fromJson(c.asJsonObject().toString(), HashMap.class);
+        gson.<HashMap<String, Object>>fromJson(c.asJsonObject().toString(), HashMap.class);
     DeltaExtras deltaExtras = c.getDelta();
     if (deltaExtras != null) {
       final HashMap<String, Object> deltaJson = new HashMap<>();
--- a/android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
+++ b/android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
@@ -655,10 +655,12 @@ public class AblyMessageCodec extends StandardMessageCodec {
     final HashMap<String, Object> jsonMap = new HashMap<>();
     writeValueToJson(jsonMap, PlatformConstants.TxMessageExtras.extras, c.asJsonObject());
     DeltaExtras deltaExtras = c.getDelta();
-    final HashMap<String, Object> deltaJson = new HashMap<>();
-    writeValueToJson(deltaJson, PlatformConstants.TxDeltaExtras.format, deltaExtras.getFormat());
-    writeValueToJson(deltaJson, PlatformConstants.TxDeltaExtras.from, deltaExtras.getFrom());
-    writeValueToJson(jsonMap, PlatformConstants.TxMessageExtras.delta, deltaJson);
+    if (deltaExtras != null) {
+      final HashMap<String, Object> deltaJson = new HashMap<>();
+      writeValueToJson(deltaJson, PlatformConstants.TxDeltaExtras.format, deltaExtras.getFormat());
+      writeValueToJson(deltaJson, PlatformConstants.TxDeltaExtras.from, deltaExtras.getFrom());
+      writeValueToJson(jsonMap, PlatformConstants.TxMessageExtras.delta, deltaJson);
+    }
     return jsonMap;
   }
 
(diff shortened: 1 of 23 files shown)


test_integration/lib/utils/encoders.dart
```diff
@@ -1,5 +1,17 @@
 import 'package:ably_flutter/ably_flutter.dart';
 
+Map<String, dynamic> encodeMessageExtras(MessageExtras extras) => extras == null
```
ben-xD, line 3: Should this be a method on `MessageExtras` instead of a function?, so usage will be `message.extras.toMap()`. Somewhat related to renaming `message.extras.extras` into something more readable. Then in the tests, no extra functions are needed

test_integration/lib/test/realtime/realtime_subscribe.dart
```diff
@@ -1,26 +1,30 @@
...
   );
   await realtime.connect();
+  return realtime;
+}
 
+Future<List<Map<String, dynamic>>> _getAllMessages(
+  String apiKey,
+  String channelName, {
+  String name,
+  List<String> names,
```
ben-xD, line 24: Whats the difference between `channelName`, `name` and `names`? I think the parameter names (and also the arguments used) are a bit confusing. I guess my confusion stems from `channel.subscribe(name: name, names: names)`.

ben-xD (commented):
I spotted a few things that slowed me down when reading the code, so it might help the readability/ maintainability of the code if those points are addressed, but I did not check test coverage, conformance to the spec/ implementation correctness though.

I won't approve/ disapprove because of my limited understanding of the codebase so far, I leave that @QuintinWillison

android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
```diff
@@ -444,6 +450,13 @@ private Object decodeChannelMessageData(Map<String, Object> jsonMap) {
     return decodeMessageData(jsonMap.get(PlatformConstants.TxMessage.data));
   }
 
+  private MessageExtras decodeChannelMessageExtras(Map<String, Object> jsonMap) {
+    if (jsonMap == null) return null;
+    Object extras = jsonMap.get(PlatformConstants.TxMessageExtras.extras);
+    final Gson gson = new Gson();
```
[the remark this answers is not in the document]
QuintinWillison: Good spot. Example of how we do it in `ably-java` [here](https://github.com/ably/ably-java/blob/68477899d2b67f564517c9368bec4e71c2f03547/lib/src/main/java/io/ably/lib/types/Capability.java#L187).

lib/src/spec/message.dart
```diff
@@ -49,38 +50,71 @@ class MessageData<T> {
...
-      return MessageExtras<Map>(value);
-    } else {
-      throw AssertionError(
-        'Message extras must be either `Map`, or `null`.'
-        ' Does not support $value ("${value.runtimeType}")',
-      );
-    }
+  static MessageExtras fromMap(Map<String, dynamic> value) {
+    if (value == null) return null;
+    value = Map<String, dynamic>.from(value);
```
QuintinWillison, line 100: I think this is a shallow copy. Potentially this could be quite a deep structure, so that could be problematic.

lib/src/spec/message.dart
```diff
@@ -49,38 +50,71 @@ class MessageData<T> {
...
+    final deltaMap = value.remove('delta') as Map;
+    final extras = MessageExtras(value)
+      .._delta = DeltaExtras._fromMap(deltaMap);
```
QuintinWillison, line 103: I'm not convinced this is appropriate use of cascades, as it makes this code very confusing to the casual reader, especially given you're mutating a member with an underscore prefix.

lib/src/spec/message.dart
```diff
@@ -49,38 +50,71 @@ class MessageData<T> {
...
+  String toString() => {'extras': extras, 'delta': delta}.toString();
+
+  @override
+  bool operator ==(Object other) =>
+      other is MessageExtras &&
+      const MapEquality().equals(other.extras, extras) &&
+      other.delta == delta;
+
+  @override
+  int get hashCode => '$extras:${delta.hashCode}'.hashCode;
```
QuintinWillison, line 117: This is a very odd pattern (triggering potentially many `toString()` implementations), and I doubt particularly efficient. Can't you leverage the `hasCode` implementation from the `extras` map and combine that with `delta.hashCode`?

lib/src/spec/message.dart
```diff
@@ -49,38 +50,71 @@ class MessageData<T> {
...
+  /// json-encodable map of extras
+  final Map<String, dynamic> extras;
+
+  /// configuration for delta compression extension
+  DeltaExtras _delta;
 
-  /// Only Map and List types are supported
-  MessageExtras(this._extras) : assert(T == Map);
+  /// delta configuration received from channel message
+  DeltaExtras get delta => _delta;
```
QuintinWillison, line 90: This can be `null` - are we not able to annotate that on the interface?

lib/src/spec/message.dart
```diff
@@ -49,38 +50,71 @@ class MessageData<T> {
...
 
-  /// retrieve extras
-  T get extras => _extras;
+  /// extras is a mandatory json-encodable argument while delta is an extension
```
QuintinWillison, line 92: The name `extras` and the term 'mandatory' are not applicable to a class-level commentary.

See [here](https://sdk.ably.com/builds/ably/ably-flutter/pull/123/ably_flutter/MessageExtras-class.html) for rendered context.

QuintinWillison (changes requested):
Also, out of interest, did you consider using [json_serializable](https://pub.dev/packages/json_serializable) here?

lib/src/spec/message.dart
tiholic: We can do that with `nullsafety` migration

lib/src/spec/message.dart
tiholic: @QuintinWillison what do you suggest is better here? Our lint rules enforce cascading wherever applicable.

One option is to return it without assigning to a variable `extras` which will then look like:
```dart
return MessageExtras(value).._delta = DeltaExtras._fromMap(deltaMap);
```

test_integration/lib/test/realtime/realtime_subscribe.dart
tiholic: Anyhow to avoid confusion, I'm going to rename these arguments as `messageName` and `messageNames`

test_integration/lib/utils/encoders.dart
tiholic: Not sure we should do that if there is no spec item for that! 🤔 

@QuintinWillison what's your opinion?

lib/src/spec/message.dart
```diff
@@ -49,38 +50,71 @@ class MessageData<T> {
...
+  final String format;
+
+  /// initializes with from and format
+  const DeltaExtras({this.from, this.format});
+
+  /// create instance from a map
+  static DeltaExtras _fromMap(Map value) {
+    if (value == null) return null;
+    return DeltaExtras(
+      from: value['from'] as String,
```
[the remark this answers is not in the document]
tiholic: The generated constants are only used for encoding/decoding w.r.t. platform channel communication. I feel it is best to keep their usage isolated only for that.

test_integration/lib/test/realtime/realtime_subscribe.dart
ben-xD: Would a user ever need to use both `messageName` and `messageNames` simultaneously? If not, can you also limit it to `names`, or split them into 2 functions?

In TS, they have union types, so it makes sense in `messageCallback`. However, taking 2 forms of `messageName` simultaneously is a bit complicated.

test_integration/lib/utils/encoders.dart
ben-xD: It's not a feature spec point, its an implementation detail related to clean code/ encapsulation.

lib/src/spec/message.dart
ben-xD: Why not make it optional argument on the `MessageExtras` initialiser?

ios/Classes/codec/AblyPlatformConstants.m
```diff
@@ -70,6 +70,14 @@
 NSString *const TxMessageData_data = @"data";
 NSString *const TxMessageData_type = @"type";
 
+// key constants for DeltaExtras
+NSString *const TxDeltaExtras_format = @"format";
```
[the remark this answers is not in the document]
tiholic: Uniqueness should not be a concern from what I can see!


In case you haven't looked at the usage in depth, here's how the story goes... :

`dart` platform message encoder encodes `DeltaExtras` as a json by putting value for `format` against `TxDeltaExtras.format` key. Now, `cocoa` or `java` extracts that value using `TxDeltaExtras_format` or `TxDeltaExtras.format` respectively. And vice versa is also true, i.e. when the encoding happens on platform side and decoded on dart side.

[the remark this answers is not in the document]
tiholic: In case you haven't noticed, these constants are generated.
You can read more about them here https://github.com/ably/ably-flutter/tree/main/bin#code-generation-to-keep-platform-constants-in-sync

android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
```diff
@@ -444,6 +450,13 @@ private Object decodeChannelMessageData(Map<String, Object> jsonMap) {
...
+    return new MessageExtras(gson.fromJson(gson.toJson(extras, Map.class), JsonObject.class));
```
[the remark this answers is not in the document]
tiholic: I'm using that to convert a `Map` to `JsonObject` where `Map` need not only be a `Map<String, String>`

lib/src/spec/message.dart
QuintinWillison: see: https://github.com/ably/ably-flutter/issues/82

test_integration/lib/test/realtime/realtime_subscribe.dart
tiholic: Limiting it to `names` sounds better, but that's against the spec.

Dart doesn't allow overloading: https://dart.dev/guides/language/effective-dart/design#avoid-using-runtime-type-tests-to-fake-overloading

And union types are in consideration: https://github.com/dart-lang/language/issues/1222 & https://github.com/dart-lang/language/issues/83

android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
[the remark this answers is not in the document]
ben-xD: So you're using `toJson` to remove non `String` keys. Is it just in case a user adds a non-string key on the dart-side? That's quite subtle, but also tackling the problem deeper than where it happens? Why not just maintain the invariant (extras don't have non-string keys) when a user creates them?

test_integration/lib/test/realtime/realtime_subscribe.dart
ben-xD: > Limiting it to names sounds better, but that's against the spec.

Do you have the feature spec point that says that? 

> Dart doesn't allow overloading
Yup, but that doesn't mean you can't have 2 functions, and to quote your first link:

> If callers usually know which type they have and which specific operation they want, it’s better to define separate methods with different names to let callers select the right operation. 

They don't mention making one function which takes all (both) the types a user might want to use.

> And union types are in consideration

Yea, I meant you're taking inspiration from typescript's union type, and unfortunately it doesn't work in Dart.

QuintinWillison: I see in the root readme you state:

> Use `channel.subscribe(name: "event1")` or `channel.subscribe(names: ["event1", "event2"])` to listen to specific named messages.

I also note that this means of defining method signature also applies to publish.

Regardless, @ben-xD, I'm not keen to be changing the plugin public API at this stage. Though I would appreciate @tiholic validating that those examples in the readme both work as I do find the presence of multiple arguments on `subscribe` (`name` and `names`) confusing. Especially bearing in mind the API commentary for that method doesn't mention this in any way.

Also, for what it's worth, I don't think that JavaScript interfaces are a good source of inspiration for Dart interfaces... I would have preferred you looked to Java or Cocoa for that. 🤔

ios/Classes/codec/AblyPlatformConstants.m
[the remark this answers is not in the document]
ben-xD: I haven't seen that before or noticed those were generated. So what you're saying is `format` and `from` are already scoped to `DeltaExtras`, there is no point prepending `"DeltaExtras"` to it. If there are no chances of collision, then I'm happy for you to resolve this conversation. Thanks

However, could we update the readme there to say that `codegencontext.dart` needs to be updated with new values, the `generating files` section is not just as simple as `cd bin` and `dart codegen.dart`.  found the readme a bit hard to understand, and had to use this PR to piece it together. Perhaps a `# Usage` section would be good for fresh eyes.

lib/src/spec/message.dart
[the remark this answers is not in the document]
ben-xD: Are you saying even if we change the values (`from`, `format`) in `codegencontext.dart`, this code will keep working?

lib/src/spec/message.dart
```diff
@@ -49,38 +50,71 @@ class MessageData<T> {
...
-      return value;
-    } else if (value is Map) {
-      return MessageExtras<Map>(value);
-    } else {
-      throw AssertionError(
-        'Message extras must be either `Map`, or `null`.'
-        ' Does not support $value ("${value.runtimeType}")',
-      );
-    }
+  static MessageExtras fromMap(Map<String, dynamic> value) {
```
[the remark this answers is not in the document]
ben-xD: Actually this is fine, or alternatively use `map` instead of `value`, because that is more readable in the body of the method. I don't mind either way, you can resolve this whenever.

test_integration/lib/utils/encoders.dart
QuintinWillison: Perhaps a compromise would be to make this an [extension method](https://dart.dev/guides/language/extension-methods), defined by the test suite. We would not want to be exposing this as public API on the SDK, especially if it's not defined in the feature spec..

test/models/message_test.dart
```diff
@@ -55,7 +55,7 @@ void main() {
       });
 
       test('#extras retrieves extras', () {
-        expect(message.extras, extras);
+        expect(message.extras.extras, extras);
```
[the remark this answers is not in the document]
ben-xD: Again, this is just a test so feel free to not do this, especially if you have other things to do.

ios/Classes/codec/AblyPlatformConstants.m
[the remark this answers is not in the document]
tiholic: We do have a section describing, briefly, the same: https://github.com/ably/ably-flutter/tree/main/bin#when-would-i-need-to-run-code-generation

We can re-organize if necessary. Or, would it be easier of there are more details with steps on how to create new constants? We can move this discussion to an issue, if that is the case

lib/src/codec.dart
```diff
@@ -792,19 +808,30 @@ class Codec extends StandardMessageCodec {
...
+    if (jsonMap == null) return null;
+    return MessageExtras.fromMap(jsonMap);
+  }
+
   /// Decodes value [jsonMap] to [Message]
   /// returns null if [jsonMap] is null
   Message _decodeChannelMessage(Map<String, dynamic> jsonMap) {
     if (jsonMap == null) return null;
     final timestamp = _readFromJson<int>(jsonMap, TxMessage.timestamp);
+    var extras = _readFromJson<Object>(jsonMap, TxMessage.extras);
```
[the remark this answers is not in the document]
tiholic: Yes, there is a change of it not being a `MessageExtras` object on decode as, unlike java, the cocoa library doesn't have a dedicated model for `MessageExtras` (and `DeltaExtras`)

test_integration/lib/test/realtime/realtime_subscribe.dart
tiholic: Java supports both `name` and `names` with help of overloading, but looks like cocoa supports only `name`.

Created an issue to track this: https://github.com/ably/ably-flutter/issues/131

android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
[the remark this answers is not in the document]
tiholic: 1f0c17c1093de7f146b4b03af202a398a43150c8

android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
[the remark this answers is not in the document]
tiholic: 1f0c17c1093de7f146b4b03af202a398a43150c8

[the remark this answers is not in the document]
tiholic: Improved readability: 1f0c17c1093de7f146b4b03af202a398a43150c8

lib/src/codec.dart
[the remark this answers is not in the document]
tiholic: added commentary clarifying the same: 8bac9b5a6b0563c0a80d8ba92b7a6bbf95ff492d

lib/src/spec/message.dart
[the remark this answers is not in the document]
tiholic: renamed here: 16b10b18550155c054f65d2d912ba1051e14ac5c

test/models/message_test.dart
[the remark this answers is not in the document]
tiholic: This is a definition from the library itself, so your suggestion stands valid :)
addressed here: f46faaca29d3f75254f06f420fae043b8e5c63f2

test_integration/lib/utils/encoders.dart
tiholic: I'd say we leave it at its existing implementation for now. Please create an issue if this change could have a positive impact on performance or readability.

lib/src/spec/message.dart
tiholic: b1165e542c36e69d3b0e020c515dbc2c42b45b13

lib/src/spec/message.dart
tiholic: Considering inputs from both: d30ed0ea880c155e6f5ff7290c4974c592b80e1b

> You may wonder why a private named constructor instead of adding to default constructor:
_Deltas is not supposed to be assigned by SDK users in any way as it is read only by definition.
However, it still stands as an exposed API via `MessageExtras.fromMap` due to technical limitations of ability to configure a package private constructor. So, just trying to reduce the exposure as much as we can._

lib/src/spec/message.dart
tiholic: 9818326e39dad8cac0b12b87f333aac1a919c1f9

lib/src/spec/message.dart
tiholic: 675bd272850444a0e9bdc1058f54ea9574b27707

lib/src/spec/message.dart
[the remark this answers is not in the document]
tiholic: Well, that's a good point. Addressed it here: 66a6a29acc1e442484d7679e014f9becd70401ff

android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
```diff
@@ -181,9 +188,9 @@ protected void writeValue(ByteArrayOutputStream stream, Object value) {
 
   private void WriteJsonElement(ByteArrayOutputStream stream, JsonElement value) {
     if (value instanceof JsonObject) {
-      super.writeValue(stream, (new Gson()).fromJson(value, Map.class));
+      super.writeValue(stream, (gson).fromJson(value, Map.class));
```
ben-xD, line 191: Is there a need for the paranthesis now?

android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
```diff
@@ -640,6 +652,20 @@ private String encodeChannelEvent(ChannelEvent event) {
     return jsonMap;
   }
 
+  private Map<String, Object> encodeChannelMessageExtras(MessageExtras c) {
```
ben-xD, line 655: What does `c` and `v` stand for?

ben-xD (approved):
Just a couple of readability comments left from me. Thanks

android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
tiholic: shorthand references were used in this class for all encoder/decoders to keep the code compact. As it is an internal detail much attention was not given to why that specific character was chosen. They were all just duplicated and modified to suit necessary encoder.

android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
tiholic: removed: f7be2e8595b6e13b5e448f9030390e10247cfb84

lib/src/spec/message.dart
QuintinWillison: Hmm, still not ideal as by "combine" I wasn't describing still using string representation to mix them together, however I recognise I wasn't clear on that. We need to get this PR landed so I'm not going to insist on this changing now but I have created an issue to improve in due course: https://github.com/ably/ably-flutter/issues/132

lib/src/spec/message.dart
QuintinWillison: Yup, that'll work, but as has been mentioned elsewhere this is a bit hammer to crack a nut (serialize in order to immediately deserialize). We could improve this in future, for which I've created https://github.com/ably/ably-flutter/issues/133.

QuintinWillison (approved):

PR #136, opened 2021-06-24
Migrate to null-safety

tiholic:
- `lib`
- `test`
- `example`
- `test_integration`


ben-xD:
Hi @tiholic :)

Could you explain the process that you took to migrate to null safety? That would be interesting and useful to know. 

I notice there a lot of usages of optionals, and I wanted to see if they're necessary (though it will require some work to find out which fields **need** to be optional). For example, all the fields of `Message` are optional, but isn't `id` and `timestamp` always known? An example of useful information for investigation is [TM2](https://docs.ably.io/client-lib-development-guide/features/#TM2).

- There are many states that e.g. `Message` can be in if we make all the fields optional. This makes it complex to construct and handle.
- Reducing the usage of optionals is useful to avoid having to unwrap optional. In fact, it might make code more verbose because optional cases need to be handled even if they do not happen in reality. This tempts the developer to use `!` and just implicitly unwrap it dangerously.

Another example is `AblyMessage`, which is used to communicate between platform and dart. If we require `handle` and `type`, that would make the usage of this stricter, and so on the other side of the message, we can guarantee the data to be there.

**TLDR:** Though we don't have compilation failures and we used the migration tool, there could be a lot of improvements because our models are too flexible, we may be abusing nullability by using `?` too much.

tiholic:
Hey @ben-xD!

I followed this guide first to auto migrate https://dart.dev/null-safety/migration-guide, then looked into each of the implemented APIs (channel publish, subscribe, history, paginated results, etc). While I may also have touched few areas that are not implemented, we need to ensure they are appropriate while implementing them as per the spec.

> For example, all the fields of Message are optional, but isn't id and timestamp (and others) always known?

A message can be received (read) or published (created). So fields like `id` and `timestamp` are not always expected to be present. An alternative would be to have different models for read and write (deviates from spec).

As you may have understood by now that `AblyMessage.type ` only comes into play when we have generics involved, and `handle` is not always necessary (refer to usages).


As many languages do not have support for null safety, it might not have been considered while building the SDK developer docs to flag which property of a model is `nullable` and which is not.

QuintinWillison (approved):
This is very difficult to review line by line, but looks sensible to me.

test_integration/test_driver/tests_abstract.dart
```diff
@@ -6,37 +6,38 @@ import 'tests_config.dart';
...
 
   tearDownAll(() async {
-    if (driver != null) {
-      const message = TestControlMessage(TestName.getFlutterErrors);
-      final flutterErrors = await getTestResponse(driver, message);
-      print('Flutter errors: ${flutterErrors.payload}');
-      final _ = driver.close();
-    }
+    // if (driver != null) {
+    // TODO can driver be null? :thinking:
```
ben-xD, line 27: Can these comments be removed? Since driver is never null.

example/lib/main.dart
```diff
@@ -480,11 +481,11 @@ class _MyAppState extends State<MyApp> {
...
-    String name,
-    bool enabled,
-    ably.PaginatedResult<T> page,
-    Future<ably.PaginatedResult<T>> Function() query,
-    Function(ably.PaginatedResult<T> result) onUpdate,
+    required bool enabled,
+    String? name,
+    ably.PaginatedResult<T>? page,
+    Future<ably.PaginatedResult<T>> Function()? query,
+    Function(ably.PaginatedResult<T> result)? onUpdate,
```
ben-xD, line 488: These 2 parameters (query, onUpdate) are always provided in the usages of `getPageNavigator`. They are also required in the logic of this method. Why not make them required instead of optional? Then we can remove the `!` from onUpdate and query.

example/lib/nested_realtime_events.dart
```diff
@@ -21,7 +21,7 @@ void listenRealtimeConnection(ably.Realtime realtime) {
 
   //DISPOSE ON CONNECTED
   final stream = realtime.connection.on();
-  StreamSubscription omegaSubscription;
+  late StreamSubscription omegaSubscription;
   omegaSubscription = stream.listen((stateChange) async {
```
ben-xD, line 25: This can be `final omegaSubscription = ...`, and the line above can be removed? Same for the other subscriptions.

Interestingly this file is not being used (as mentioned at the top of the file). Is it still valid to keep?

test/ably_event_listener_test.dart
```diff
@@ -25,15 +25,15 @@ class MockEmitter {
   int streamsCount;
   Map<int, int> indexes = {};
   List<int> injectables;
-  List<Stream<int>> streams;
+  late List<Stream<int>> streams;
```
ben-xD, line 28: This doesn't need to be late, since we can just use `List<Stream<int>> streams = [];`

ben-xD (approved):
LGTM with comments

example/lib/nested_realtime_events.dart
tiholic: We're using `await omegaSubscription.cancel();` inside the `onData` callback of `listen` which won't recognize the `omegaSubscription` if it is not already defined/declared.

I think we can remove this once we have a more modular example app.

test_integration/test_driver/tests_abstract.dart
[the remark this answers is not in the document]
tiholic: 2d69bc1784966c6d42b1179bf95267e100afd28b

example/lib/main.dart
tiholic: ff2a1ce4cc50d6cb75664e8256a083a28ae3445c

test/ably_event_listener_test.dart
tiholic: It can also be marked final.

ff2a1ce4cc50d6cb75664e8256a083a28ae3445c

ben-xD (approved):

ben-xD:
The failing tests have been disabled and are successfully passing (on most occasions) in https://github.com/ably/ably-flutter/pull/144. I'm merging this in now even though the tests fail, because the test disabling PR depends on this one.

PR #137, opened 2021-06-25
Add instructions for M1 macs and fixing ruby fatal error for example apps

ben-xD:


.ruby-version
```diff
@@ -0,0 +1 @@
+3.0.1
```
tiholic, line 1: Why do we need ruby related configuration here?

ben-xD: The reason is Ruby is needed, is to install cocoapods and ffi, used for iOS. More information: because `flutter run` can run in the root directory, the ruby version should be specified wherever this can be run, not just the iOS directory. We actually already have it, but in an uncommon format: it's in `.tool-versions`. 

However, if want to de-duplicate `3.0.1`, we should delete `tool-versions` because that is a niche format only used by `asdf`, where as `asdf` can also use `ruby-version`.

tiholic: Pretty detailed! Thanks :)

tiholic (approved):

README.md
```diff
@@ -57,6 +57,15 @@ Features that we do not currently support, but we do plan to add in the future:
 - cd to `example` folder
 - run `flutter pub get` to install dependencies
 - `flutter run` will start the application on connected android / iOS device
+- Note for M1 Macs:
```
QuintinWillison, line 60: I would suggest starting a new section here:

```md
## Troubleshooting the example
```

Because hopefully most users will not hit these issues, so that keeps "Running the example" as pure "golden path" instructions.

QuintinWillison (changes requested):

README.md
ben-xD: I've added this and also added `## Example App` and `### Running example app`. 9764e94

.ruby-version
QuintinWillison: @ben-xD It was already specified in [.tool-versions](https://github.com/ably/ably-flutter/blob/main/.tool-versions). It's not ideal to maintain this in two places, but if you must then please update both.

ben-xD: `.tool-version` was already deleted because ASDF can read `.ruby-version`, so there is no duplication. That file was deleted in 5bae664

QuintinWillison: ok, sorry, I looked at the wrong branch. I also didn't realise that ASDF would read `.ruby-version`. I can think of at least [one other place](https://github.com/ably/ably-asset-tracking-swift/blob/main/.tool-versions) we could make this switch.

QuintinWillison (approved):

.ruby-version
ben-xD: Okay good idea, I made that change: https://github.com/ably/ably-asset-tracking-swift/pull/175

Issue #107, opened 2021-04-29
Push Notifications Device Registration (activation) and device subscription


QuintinWillison:
The absence of this feature is a known limitation of this plugin.

See:

- Docs: [Push Notifications - Device activation and subscription: Subscribe for push notifications](https://ably.com/documentation/general/push/activate-subscribe#subscribing)
- [RSH7](https://docs.ably.io/client-lib-development-guide/features/#RSH7): The `push` field on `RestChannel` and `RealtimeChannel`, with methods for `subscribeDevice`, `subscribeClient`, `unsubscribeDevice`, `unsubscribeClient` and `listSubscriptions`
- [RSH8](https://docs.ably.io/client-lib-development-guide/features/#RSH8): `LocalDevice` class

Issue #109, opened 2021-04-29
Push Notifications Administration (Admin)


QuintinWillison:
The absence of this feature is a known limitation of this plugin.

See:

- Docs: [Push Notifications - Admin](https://ably.com/documentation/general/push/admin)
- [RSH1](https://docs.ably.io/client-lib-development-guide/features/#RSH1): `admin` interface on `Push`, including:
    - [RSH1a](https://docs.ably.io/client-lib-development-guide/features/#RSH1a): `publish`
    - [RSH1b](https://docs.ably.io/client-lib-development-guide/features/#RSH1b): `deviceRegistrations` including `get`, `list`, `save`, `remove` and `removeWhere`
    - [RSH1c](https://docs.ably.io/client-lib-development-guide/features/#RSH1c): `channelSubscriptions` including `list`, `listChannels`, `save`, `remove` and `removeWhere`
- [RSH6](https://docs.ably.io/client-lib-development-guide/features/#RSH6): Push device authentication
- [PCD1](https://docs.ably.io/client-lib-development-guide/features/#PCD1) thru [PCD7](https://docs.ably.io/client-lib-development-guide/features/#PCD7): `DeviceDetails`

ben-xD:
From the [the Push Admin docs](https://ably.com/documentation/general/push/admin), 
> The client libraries provide a Push Admin API that is intended to be used on a customer’s servers

Since ably-flutter only runs in Flutter/ mobile applications and cannot be run on "customer's servers", the value of this feature is quite low. Having said that, a customer may wish to implement an Admin flutter application. This feature would be more useful if we build a fully dart SDK which could run on dart runtime on servers (without flutter).

PR #140, opened 2021-07-22
Push notifications (activation and subscription)

ben-xD:
Push notifications are received on Android and iOS. Push notifications are not available on the dart-side (see `Future work in a separate PR` below again). Push notifications can be sent either directly from the ably.com dashboard (notification tab), through a push admin, or on the example app which will publish to a test channel.
- User interface in example app, including separate files/ widgets to make code manageable/ readable.
- Added 2 methods to read and request user permission for notifications (Only useful on iOS, since no permission needed on Android).
- Some readme additions to help with configuring native applications for push notifications.
- And also updated the ably.com documentation to be clearer: https://github.com/ably/docs/pull/1151 and https://github.com/ably/docs/pull/1154
- Wrote push payload documentation. Crafting the correct push payload and understanding how messages are received by each platform is quite complicated, and since we are trying to abstract away android/ ios into a flutter package, we may have to handle these cases and document them to users correctly. If the payload is not constructed properly, messages will not get received properly, and users will ask us.

## Future work in a separate PR:
- **Listening/ reacting to push notifications**.  See https://github.com/ably/ably-flutter/issues/141
- ~Handle users opening/ tapping/ interacting with the notification (On Android the notification data is given in intent extras to the launcher activity. In iOS, this is a delegate method, `didReceiveNotificationResponse`).~ This doesn't need to be handled in Ably-flutter. Once the message listener in the dart side is implemented, users can listen for this message, create a notification with a callback using `flutter_local_notifications` package.
- ~Allow users to add buttons/ text entry/ customise the notification/ `UNNotificationContentExtension` so their notifications are more useful to users.~ Users can do this with `awesome_notifications` in the message listener.
- Push Admin. This is lower priority because the push admin API is documented to be for server side use. https://github.com/ably/ably-flutter/issues/109
- ~Investigate compatibility with other push notification libraries (e.g. awesome_notifications). Since Ably-java and ably-cocoa doesn't intervene with receiving push notifications, users may just prefer to activate the device with Ably but then use another library for creating alert notifications or handling background notifications.~ Awesome_notifications and flutter_local_notifications will be compatible, but firebase_messaging is not because it filters specifically for FCM messages, including on iOS. Ably does not use FCM on iOS.
- ~Investigate compatibility with other Firebase libraries (e.g. to avoid breaking apps using firebase auth), except firebase_messaging.~ It should be compatible as we do not currently change the UNNotificationCenter delegate ourselves, but ask the user to do so (e.g. it is done in the example app).
- Write tutorial and validate the correctness for the documentation/ install guide for push notifications (added in readme)

Closes https://github.com/ably/ably-flutter/issues/107


android/src/main/java/io/ably/flutter/plugin/AblyFlutterPlugin.java
```diff
@@ -28,22 +34,23 @@ public void onAttachedToEngine(@NonNull FlutterPluginBinding flutterPluginBindin
...
+        streamsChannel.setStreamHandlerFactory(arguments -> new AblyEventStreamHandler(applicationContext));
 
         final MethodChannel channel = new MethodChannel(messenger, "io.ably.flutter.plugin", codec);
         AblyMethodCallHandler methodCallHandler = AblyMethodCallHandler.getInstance(
                 channel,
                 // Streams channel will be reset on `register` method call
                 // and also on every hot-reload
-                streamsChannel::reset
+                streamsChannel::reset,
+            applicationContext
```
QuintinWillison, line 53: Indentation mismatch with lines above.

android/src/main/java/io/ably/flutter/plugin/AblyLibrary.java
```diff
@@ -1,31 +1,37 @@
...
 import io.ably.lib.realtime.AblyRealtime;
+import io.ably.lib.rest.AblyBase;
 import io.ably.lib.rest.AblyRest;
 import io.ably.lib.types.AblyException;
 import io.ably.lib.types.AsyncPaginatedResult;
 import io.ably.lib.types.ClientOptions;
+import io.ably.lib.types.ErrorInfo;
 
-class AblyLibrary {
+public class AblyLibrary {
```
QuintinWillison, line 14: I'm curious as to why package scope wasn't sufficient.

android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
```diff
@@ -652,6 +671,74 @@ private String encodeChannelEvent(ChannelEvent event) {
...
+
+    switch (state) {
+      case ACTIVE:
+        return PlatformConstants.TxDevicePushStateEnum.active;
+      case FAILING:
+        return PlatformConstants.TxDevicePushStateEnum.failing;
+      case FAILED:
+        return PlatformConstants.TxDevicePushStateEnum.failed;
+      default:
+        return null;
```
QuintinWillison, line 710: Is this ever a valid situation? I would rather you failed and threw an exception if this is indicative of an issue elsewhere.

example/android/app/build.gradle
```diff
@@ -63,4 +64,7 @@ dependencies {
     testImplementation 'junit:junit:4.12'
     androidTestImplementation 'androidx.test:runner:1.3.0'
     androidTestImplementation 'androidx.test.espresso:espresso-core:3.3.0'
+
+    // Push Notifications (you'll need a Firebase project and google-services.json)
+    implementation platform('com.google.firebase:firebase-bom:28.2.0')
```
QuintinWillison, line 69: Out of interest, can the example app be built and run without `google-services.json` being present?

example/android/app/src/main/AndroidManifest.xml
```diff
@@ -21,6 +21,12 @@
                 <category android:name="android.intent.category.LAUNCHER"/>
             </intent-filter>
         </activity>
+<!--        TODO Find a way to declare this service in a different manifest?/ provided by plugin -->
```
QuintinWillison, line 24: If you haven't already then could you please create an issue to track this. Otherwise, `TODO` statements in the code like this become a parallel backlog.

example/lib/main.dart
```diff
@@ -12,20 +15,14 @@ class MyApp extends StatefulWidget {
...
+const defaultChannel = Constants.channelName;
 
 class _MyAppState extends State<MyApp> {
   String _platformVersion = 'Unknown';
   String _ablyVersion = 'Unknown';
-  OpState _provisioningState = OpState.notStarted;
-  provisioning.AppKey? _appKey;
+
+  // final String _apiKey = const String.fromEnvironment(Constants.ablyApiKey);
+  String _apiKey = const String.fromEnvironment(Constants.ablyApiKey);
```
QuintinWillison, line 25: I know we discussed this elsewhere but is it now documented in the readme?

example/lib/main.dart
```diff
@@ -122,48 +121,22 @@ class _MyAppState extends State<MyApp> {
...
-    final clientOptions = ably.ClientOptions()
-          // .fromKey(_appKey.toString())
-          ..environment = 'sandbox'
+    final clientOptions = ably.ClientOptions.fromKey(_apiKey)
           ..logLevel = ably.LogLevel.verbose
           ..logHandler = ({msg, exception}) {
             print('Custom logger :: $msg $exception');
           }
-          ..tokenDetails = ably.TokenDetails.fromMap(
+/*          ..tokenDetails = ably.TokenDetails.fromMap(
```
QuintinWillison, line 134: The example code needs to be authoritative so commented out sections generally aren't helpful, especially if they've not got some kind of meta-comment next to them that explains _why_ they're commented out or what re-introducing them might do.

example/lib/ui/push_notifications/push_notifications_sliver.dart
```diff
@@ -0,0 +1,86 @@
...
+                'Push Notifications',
+                style: TextStyle(fontSize: 24, fontWeight: FontWeight.bold),
+              ),
+            ),
+            buildCreateAblyClientText(),
+            buildSummaryText(),
+            PushNotificationsActivationSliver(_pushNotificationService),
+            PushNotificationsSubscriptionsSliver(_pushNotificationService),
+            PushNotificationsPublishingSliver(_pushNotificationService),
+            // TODO Implement Push Admin API and then use the following:
```
QuintinWillison, line 54: Is there an issue to track this?

lib/src/platform/src/push_native.dart
```diff
@@ -0,0 +1,47 @@
...
+          {bool provisionalPermissionRequest = false}) async =>
+      invokeRequest<bool>(PlatformMethod.pushRequestNotificationPermission, {
+        TxPushRequestNotificationPermission.provisionalPermissionRequest:
+            provisionalPermissionRequest
+      });
+
+  @override
+  Future<void> deactivate() => invoke(PlatformMethod.pushDeactivate);
+
+  // TODO Consider implementing the Push Admin API (it is designed for servers)
```
QuintinWillison, line 39: Is there an issue to track this?

QuintinWillison (changes requested):
I've done a first-pass, relatively-low-quality review. A few things to mention. And also worth noting that I didn't check whether generated code committed back was up-to-date or not.

android/build.gradle
```diff
@@ -23,6 +24,10 @@ apply plugin: 'com.android.library'
 
 dependencies {
     implementation 'io.ably:ably-android:1.2.6'
+    implementation 'com.google.firebase:firebase-messaging:22.0.0'
+    // firebase-iid is a temporary workaround because `:firebase-messaging:22.0.0` removed FirebaseInstanceId
+    // TODO Remove firebase-iid dependency once [PR](https://github.com/ably/ably-java/pull/687) is released
```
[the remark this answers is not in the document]
ben-xD: Updated to the new version and removed the workaround in c2d015f25e999b67f7fc6bcded9a14d50e2a55a3

README.md
```diff
@@ -450,6 +456,76 @@ channel
...
+            - Place `google-services.json` in `example/android/app/`. We have `gitignore`d this file since it is associated with our firebase project, but it is [not sensitive](https://stackoverflow.com/questions/37358340/should-i-add-the-google-services-json-from-firebase-to-my-repository), so you can commit it to share it with other developers/ colleagues.
+            - Update your build.gradle files according to the [Set up the SDK section](https://firebase.google.com/docs/cloud-messaging/android/client#set_up_the_sdk) of the firebase guide.
+            - Declare the AblyPushNotificationService: In `AndroidManifest.xml`, declare a service already implemented in ably-flutter by adding the following between the `<application>` and `</application>` tags:
+```xml
+<service android:name="io.ably.flutter.plugin.push.AblyPushNotificationService">
+    <intent-filter>
+        <action android:name="com.google.firebase.MESSAGING_EVENT" />
+    </intent-filter>
+</service>
+```
```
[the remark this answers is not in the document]
ben-xD: Fixed in 8917549ea328a880479f8d4021168341e2f091e4

android/src/main/java/io/ably/flutter/plugin/AblyLibrary.java
ben-xD: When I added `public` to AblyLibrary, it temporarily tricked Android Studio to resolving the `AblyLibrary` symbol in other files. Annoyingly, this fix is not permanent, so there is no reason to keep this. Reverted in 2000557a1a21ed289e33820de5b93e601165e43f

example/android/app/src/main/AndroidManifest.xml
ben-xD: Comment removed in 41641c582c0fa4d23929718e9feda1772ea7aa38

There is no issue needed because it is not possible. Other libraries are making flutter developers modify the Android Manifest to add the service. For example, [`flutter_local_notifications` example `AndroidManifest.xml`](https://github.com/MaikuB/flutter_local_notifications/blob/master/flutter_local_notifications/example/android/app/src/main/AndroidManifest.xml).

example/lib/ui/push_notifications/push_notifications_sliver.dart
ben-xD: Yes, it's here https://github.com/ably/ably-flutter/issues/109

I left some related files because that is where they would go if they would be implemented.

lib/src/platform/src/push_native.dart
ben-xD: Same answer as https://github.com/ably/ably-flutter/pull/140#discussion_r685236518

example/lib/main.dart
ben-xD: Thanks for catching that. I left it in that state when I wasn't sure if we should remove the provisioning code from the example app. We decided that it was okay because users would never need to provision an Ably account themselves. Commented code removed in 7fd639f7ae5392e00babb7148cae9f16bd86a5fa

example/lib/main.dart
ben-xD: Yup, this is documented in the [README's Running the example
 section](https://github.com/ably/ably-flutter/tree/feature/push-notifications#running-the-example), for example: 
```
If starting the app using the command line, run `flutter run --dart-define=ABLY_API_KEY=put_your_api_key_here`
```

example/android/app/build.gradle
ben-xD: From previous experience not having the file, I expected the build succeeds, but it immediately crashes at runtime. However, after experimenting, this file isn't even necessary to receive push notifications. I've tested it by creating a new Android emulator image and running the app without the file.

The BoM is the "bill of materials", a nice way Firebase uses to make sure all the dependencies are versioned/ work together.

More importantly, we don't need to use the BoM because I explicitly provided 22.0.0 for firebase messaging, our only dependency from Firebase. Comment removed in 92ebd73756d9ff34ccfce6e8d54d32b1f4be06bc

android/src/main/java/io/ably/flutter/plugin/AblyEventMessage.java
```diff
@@ -0,0 +1,15 @@
+package io.ably.flutter.plugin;
+
+class AblyEventMessage<T> {
+    final String eventName;
+    final T message;
+
+    AblyEventMessage(final String eventName, final T message) {
+        if (null == eventName) {
+            throw new NullPointerException("message cannot be null.");
```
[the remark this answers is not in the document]
ben-xD: Good spot. I didn't write this class, just moved it across from another file. It should've been done as part of the refactoring PR but was done later in this PR instead. I'm not sure what the intention is. @tiholic could you help?

android/src/main/java/io/ably/flutter/plugin/AblyFlutterPlugin.java
ben-xD: Fixed in 1cd73046d6db15fb8afad4d5882ba62469fa5c6a

android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
ben-xD: That's a good point, and I would prefer it, but I was just following the rest of the file's style, `return null` occurs 42 times in that file (trying to be consistent).

example/android/app/build.gradle
ben-xD: An example of the error someone might face when they don't add the file is on this [stack overflow question](https://stackoverflow.com/questions/32072568/how-to-add-google-services-json-in-android).

ben-xD: After investigation, I found that the Firebase notification composer messages do not arrive if
Screenshot 2021-08-09 at 16 30 14
 this file is not missing. Ably push notifications are received as normal even when this file is not added. 

I would recommend keeping the instruction to add the file to the README. And perhaps even adding some guidance about firebase notification composer messages not being received.

.run/example.run.xml
```diff
@@ -1,6 +1,6 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="example" type="FlutterRunConfigurationType" factoryName="Flutter">
-    <option name="additionalArgs" value="--observatory-port 8888 --disable-service-auth-codes" />
+    <option name="additionalArgs" value="--observatory-port 8888 --disable-service-auth-codes --dart-define=ABLY_API_KEY=" />
```
tiholic, line 3: This 👏  🥇

android/build.gradle
```diff
@@ -9,6 +9,7 @@ buildscript {
 
     dependencies {
         classpath 'com.android.tools.build:gradle:4.1.1'
+        classpath 'com.google.gms:google-services:4.3.8'
```
tiholic, line 12: Looks like we have a new release recently: https://mvnrepository.com/artifact/com.google.gms/google-services/4.3.9

android/src/main/java/io/ably/flutter/plugin/AblyEventMessage.java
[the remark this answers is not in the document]
tiholic: My bad. It should have been `"eventName cannot be null"`. I must have copied the check from `AblyMessage` class and missed updating the message. Can you add a commit fixing the message @ben-xD

android/src/main/java/io/ably/flutter/plugin/AblyEventMessage.java
```diff
@@ -0,0 +1,15 @@
...
+        }
+        this.eventName = eventName;
+        this.message = message;
+    }
+
+}
```
tiholic, line 15: Missing EOF

android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
tiholic: I think we can track this in a separate issue to throw a platform exception instead of returning null.

example/android/app/src/main/AndroidManifest.xml
tiholic: Here's something on resolution of `intent-filter` between manifests https://developer.android.com/studio/build/manifest-merge

>  _The `intent-filter` element is never matched between manifests. Each is treated as unique and is added to the common parent element in the merged manifest_

android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java
```diff
@@ -583,6 +601,132 @@ private void closeRealtime(@NonNull MethodCall call, @NonNull MethodChannel.Resu
...
+        if (realtime != null) {
+          result.success(realtime.device());
+          return;
+        }
+
+        AblyRest rest = ablyLibrary.getRest(ablyClientHandle);
+        if (rest != null) {
+          result.success(rest.device());
+          return;
+        }
```
tiholic, line 721: Alternatively we can make use of the newly created `getAblyClient` method. Might save some reading effort?

```java
AblyBase client = ablyLibrary.getAblyClient(ablyClientHandle);
if (client instanceof AblyRealtime) {
  result.success(((AblyRealtime) client).device());
} else if (client!=null) {
  result.success(((AblyRest) client).device());
}
```

example/android/.gitignore
```diff
@@ -5,3 +5,5 @@ gradle-wrapper.jar
 /gradlew.bat
 /local.properties
 GeneratedPluginRegistrant.java
+
+app/google-services.json
```
tiholic, line 9: missing EOF

tiholic (changes requested):
I've gone through a first round of review covering Android and generated code related changes.

example/android/.gitignore
ben-xD: caf41a9b5163f68edefa655ed96f20e3e6417591

android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java
ben-xD: I wanted to maintain the type safety and I think this is also more readable. Casting and checking `instanceof` is less safe and readable. Additionally, to be safe using the code snippet you provide, we would have to have 3 if/else statements, and in the third one, throw an error. 

This is particularly more confusing because `AblyRealtime` is a `AblyRest` too, so this pattern may be copied in the future, with isinstance AblyRest first and things would break. 

My principle is I only cast as a last resort. WDYT?

android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java
```diff
@@ -583,6 +601,132 @@ private void closeRealtime(@NonNull MethodCall call, @NonNull MethodChannel.Resu
...
+        client.push.activate();
+        result.success(null);
+      } catch (AblyException e) {
+        handleAblyException(result, e);
+      }
+    });
+  }
+
+  private void pushRequestNotificationPermission(@NonNull MethodCall call, @NonNull MethodChannel.Result result) {
+    result.success(true);
```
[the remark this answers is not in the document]
ben-xD: fba3a5deb098fe4c8093a95a553474e88d6a3937

android/src/main/java/io/ably/flutter/plugin/AblyEventMessage.java
ben-xD: c585daabfbf0eec1269813774fb37bcf41f0b6f7 contains new extra line and log message fix

example/lib/ui/push_notifications/push_notifications_sliver.dart
ben-xD: I thought about it, and there is no point keeping a few fields for a feature that is not yet implemented (since it is a "server side" feature as described in the docs). It will also go out of sync with decisions.

8cd1585b6f3d9bcacff42eb536c05d30fe42f7be removed the fields and comments.

Please see and use https://github.com/ably/ably-flutter/issues/109 to track work on this.

android/src/main/java/io/ably/flutter/plugin/AblyEventMessage.java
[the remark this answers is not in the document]
ben-xD: c585daabfbf0eec1269813774fb37bcf41f0b6f7

android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
ben-xD: Okay, I created https://github.com/ably/ably-flutter/issues/149

example/android/app/src/main/AndroidManifest.xml
ben-xD: Thanks, I did not know about [that page](https://developer.android.com/studio/build/manifest-merge), I was reading [another page](https://android-doc.github.io/tools/building/manifest-merge.html#merge-rules). I learnt about the Merged Manifest tab in Android Studio. I've successfully moved the declaration to the plugin AndroidManifest, out of the example. 945d57beb5eb4a2afdf909b03264d78a198fdae8

android/build.gradle
ben-xD: Use latest version of google-services android package. Also remove unnecessary dependency in plugin project.	110a4a7	Ben Butterworth <24711048+ben-xD@users.noreply.github.com>	10 Aug 2021 at 11:23

android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java
```diff
@@ -583,6 +601,132 @@ private void closeRealtime(@NonNull MethodCall call, @NonNull MethodChannel.Resu
...
   }
 
+  private void pushActivate(@NonNull MethodCall call, @NonNull MethodChannel.Result result) {
+    final AblyFlutterMessage message = (AblyFlutterMessage) call.arguments;
+    this.<Integer>ablyDo(message, (ablyLibrary, ablyClientHandle) -> {
+      // Launch a Service of FirebaseMessagingService. Or should this be done during subscribe?
+
+      try {
+        AblyBase client = ablyLibrary.getAblyClient(ablyClientHandle);
+        client.push.activate();
```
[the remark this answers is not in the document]
ben-xD: I think this one is subjective. I think currently it's more readable:
- We know that we are using an `AblyBase`, and 
- `getAblyClient` is a method call I want to separate from the subsequent method call, `activate`.

test_integration/lib/provisioning.dart
```diff
@@ -94,36 +81,9 @@ Future<Map<String, dynamic>> getTokenRequest() async {
...
     delayFactor: Duration(seconds: 2),
-  ).retry(() => http.get(Uri.parse(authURL)));
+  ).retry(() => http.get(Uri.parse(_authURL)));
   print('tokenRequest from tokenRequest server: ${r.body}');
   return Map.castFrom<dynamic, dynamic, String, dynamic>(
     jsonDecode(r.body) as Map,
   );
 }
-
-Future<Map<String, dynamic>> getTokenDetails(
```
tiholic, line 104: We may need this in `test_integration` when we write integration tests covering `authCallback`

test_integration/lib/provisioning.dart
```diff
@@ -94,36 +81,9 @@ Future<Map<String, dynamic>> getTokenRequest() async {
...
-        body: json.encode({
-          'keyName': keyName,
-          'timestamp': DateTime.now().millisecondsSinceEpoch,
-        }),
-      ));
-  print('tokenDetails from server: ${r.body}');
-  return Map.castFrom<dynamic, dynamic, String, dynamic>(
-    jsonDecode(r.body) as Map,
-  );
-}
```
tiholic, line 129: Now that this content is moved into `test_integration`, we can remove `ably_flutter_example` from `test_integration/pubspec.yaml`

tiholic (changes requested):
I see integration tests are missing. If it is not possible to do a complete testing of newly implemented features (i.e. to check we receive push notifications), we can assert basic functionality by checking the method calls execute successfully.

Also unit tests for models that are worked on as a part of this PR are missing. See `tests/models` for existing unit tests.

android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java
tiholic: Makes sense.

android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java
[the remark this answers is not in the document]
tiholic: Alright

example/lib/main.dart
tiholic: I think with the update to `.run/example.run.xml` the `--dart-defined` is already included for Android studio run configuration. We can just mention to set the API key at end of `Additional run args`

ben-xD: The instruction for that is also in the [section](https://github.com/ably/ably-flutter/tree/feature/push-notifications#running-the-example) 
> - If starting the app using Android Studio:
>   - Edit the example app run configuration add the --dart-define flag to additional run args

I took this opportunity to add more detail and improve the language in 6a4ca945e8ff5bc100abd7a7346337271c2f25b2

[Click for Preview](https://github.com/ably/ably-flutter/blob/5c128027bcb8ba91e4ec8b1df25d968c93334e57/README.md)

android/src/main/java/io/ably/flutter/plugin/push/AblyPushNotificationService.java
```diff
@@ -0,0 +1,39 @@
...
+    // One approach: Listen for `PUSH_NOTIFICATION_ACTION` using a broadcast receiver and there, call the Ably dart side.
+  }
+
+  @Override
+  public void onNewToken(@NonNull String s) {
+    ActivationContext.getActivationContext(this).onNewRegistrationToken(RegistrationToken.Type.FCM, s);
+    super.onNewToken(s);
+  }
+}
+
```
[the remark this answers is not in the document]
ben-xD: c5f19a48ace6cce798eddc8543f25928df89d1c3

lib/src/platform/src/codec.dart
```diff
@@ -645,6 +662,104 @@ class Codec extends StandardMessageCodec {
...
+      case TxDevicePushStateEnum.failing:
+        return DevicePushState.failing;
+      case TxDevicePushStateEnum.failed:
+        return DevicePushState.failed;
+      default:
+        return null;
+    }
+    // throw AblyException(
+    //   'Platform communication error. DevicePushState is invalid: $enumValue',
+    // );
```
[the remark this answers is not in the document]
ben-xD: It should never be null, but it is (there is a bug in ably-cocoa/ ably-java, actually they just forgot to implement that field completely 🐛 ). See https://github.com/ably/ably-flutter/issues/146

I've removed the comments since it can be null. 768d1f3b44e3d4e0fe0d02c3a3e99ec96f680bdb

test_integration/lib/provisioning.dart
ben-xD: Awesome, thanks for catching that. 92b0809afeebef2a6f0876bc2b1ba06cc918c76b

lib/src/platform/src/push_native.dart
```diff
@@ -0,0 +1,47 @@
...
+    }
+  }
+
+  @override
+  Future<void> activate() => invoke(PlatformMethod.pushActivate);
+
+  @override
+  Future<bool> requestNotificationPermission(
+          {bool provisionalPermissionRequest = false}) async =>
+      invokeRequest<bool>(PlatformMethod.pushRequestNotificationPermission, {
```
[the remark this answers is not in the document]
ben-xD: Good idea. 24bb5e0762ca0f807cb7be02ce97ad2d602772e9 and doc comment fixed in 535cb7e4f04309aa7e109748a37ec41484e8428c

test_integration/lib/provisioning.dart
ben-xD: Thanks for catching that. I have put it back in with a8beb04fc7c83f66c07e1fccf255b3276586e53b

lib/src/platform/src/push_channel_native.dart
```diff
@@ -0,0 +1,83 @@
...
+          code: 40000,
+          href: 'https://help.ably.io/error/40000',
+          message:
+              "expected parameter 'deviceId', 'clientId', 'deviceClientId', and/or 'channel'");
+
+      throw AblyException(
+          '40000',
+          "expected parameter 'deviceId', 'clientId', 'deviceClientId', and/or 'channel'",
+          errorInfo);
+    }
```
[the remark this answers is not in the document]
ben-xD: Good idea, done in e53a5f915e139d1fb09c21323ae41157600c434d

ben-xD:
While the platform code was still in my head, I thought it would be a good idea to encapsulate and refactor...


# Example

## Initial code:
```objective-C
static const FlutterHandler _pushSubscribeDevice = ^void(AblyFlutterPlugin *const plugin, FlutterMethodCall *const call, const FlutterResult result) {
    _ablyClientReceiver(plugin, call, result,
                       ^void(ARTRealtime *const realtime,
                             FlutterMethodCall *const call,
                             const FlutterResult result,
                             NSString *const channelName) {
        ARTRealtimeChannel *const channel = [realtime.channels get: channelName];
        [channel.push subscribeDevice];
        result(nil);
    },
                       ^void(ARTRest *const rest,
                             FlutterMethodCall *const call,
                             const FlutterResult result,
                             NSString *const _Nullable channelName){
        [[rest.channels get:channelName].push subscribeDevice];
        result(nil);
    }
   );
};

static const FlutterHandler _pushUnsubscribeDevice = ^void(AblyFlutterPlugin *const plugin, FlutterMethodCall *const call, const FlutterResult result) {
    _ablyClientReceiver(plugin, call, result,
                       ^void(ARTRealtime *const realtime,
                             FlutterMethodCall *const call,
                             const FlutterResult result,
                             NSString *const _Nullable channelName) {
        [[realtime.channels get: channelName].push unsubscribeDevice];
        result(nil);
    },
                       ^void(ARTRest *const rest,
                             FlutterMethodCall *const call,
                             const FlutterResult result,
                             NSString *const _Nullable channelName){
        [[rest.channels get:channelName].push unsubscribeDevice];
        result(nil);
    }
   );
};
```

## After Swift refactor
A new file is created, `PushNotificationHandlers`, where the handlers go.
```swift
    @objc
    public static let subscribeDevice: FlutterHandler = { plugin, call, result in
        callbackWithAblyPushChannel(plugin: plugin, call: call, result: result) { pushChannel, call, result in
            pushChannel.subscribeDevice()
            result(nil)
        }
    }
    
    @objc
    public static let unsubscribeDevice: FlutterHandler = { plugin, call, result in
        callbackWithAblyPushChannel(plugin: plugin, call: call, result: result) { pushChannel, call, result in
            pushChannel.unsubscribeDevice()
            result(nil)
        }
    }
```

# Example 2:
TBC

lib/src/platform/src/push_channel_native.dart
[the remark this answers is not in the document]
tiholic: href should be dynamic; based on the error code, I believe. Can you check the error code href documentation for better impelementation?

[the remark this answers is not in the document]
tiholic: href should be dynamic; based on the error code, I believe. Can you check the error code href documentation for better implementation?

tiholic:
Nice refactoring there! Looks like Obj-C will now be easier to maintain and extend for new APIs :)

lib/src/platform/src/push_channel_native.dart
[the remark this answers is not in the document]
ben-xD: Oh i see, and also, the AblyException code is different to the ErrorInfo code. Im going to leave the AblyException `null` because they are different, WDYT? d0fdbf92e3e8c47c8922cfc8ab3713b61534c145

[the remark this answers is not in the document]
ben-xD: Actually, after discussing on slack, that extra convenience constructor was not necessary. Removed in 89a9176cf1ee33a88048cd45771995ddc6278357

tiholic (approved):
Approving without tests.

Track @  https://github.com/ably/ably-flutter/issues/153

README.md
```diff
@@ -27,7 +27,7 @@ We handle the complexity of realtime messaging so you can focus on your code.
```
ben-xD, line 27: @marklewin this may be a good starting point, but `Push Notifications.md` is the majority of the documentation.

Push Notifications.md
```diff
@@ -0,0 +1,285 @@
+# Push Notifications
```
ben-xD, line 1: @marklewin this is the main documentation added for this work.

QuintinWillison (approved):

PR #144, opened 2021-08-05
Skip 1 failing test, fix 1 failing test and schedule tests on main twice a day

ben-xD:
Schedule integration test workflow to run at 4.39am and 15:39pm daily. The times chosen are pretty much random, because [Github documentation on scheduled workflows](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events) state:
> Note: The schedule event can be delayed during periods of high loads of GitHub Actions workflow runs. High load times include the start of every hour. To decrease the chance of delay, schedule your workflow to run at a different time of the hour.

I've added this because we shouldn't have to discover that integration tests are failing due to server-side Ably changes only once our PRs are ready to be merged. This will increase visibility of the behaviour changes in the dependencies (e.g. realtime). If a test fails and code was not changed, we can be sure to investigate a certain type of behaviour (breaking changes from platform as opposed to code changes in the PR)

This PR also includes a file name change, an integration test code simplication, and method rename.


.github/workflows/check.yaml
```diff
@@ -3,6 +3,8 @@ on:
   push:
     branches:
       - main
+  schedule:
+    - cron: '39 4,15 * * *'
```
QuintinWillison, line 7: I think the [cron encoding](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#schedule) is sufficiently opaque that it's worth adding a comment above this line to explain what it's scheduling in a human-readable form, please.

QuintinWillison (changes requested):

.github/workflows/check.yaml
ben-xD: Good idea, thanks. e8b721e

test_integration/lib/utils/encoders.dart
```diff
@@ -15,7 +15,7 @@ Map<String, dynamic>? encodeMessageExtras(MessageExtras? extras) =>
 
 Map<String, dynamic> encodeMessage(Message message) => {
       'id': message.id,
-      'timestamp': message.timestamp!.toIso8601String(),
+      'timestamp': message.timestamp?.toIso8601String(),
```
tiholic, line 18: Why are we expecting timestamp to be null? A message in history should always bear a timestamp, right?

test_integration/README.md
```diff
@@ -1,8 +1,22 @@
...
+ - run `flutter drive` which runs them all (one and only, `main.dart`).
+
+To run a specific test:
+- In [tests_config.dart](test_driver/tests_config.dart), comment out the key: value pairs which you don't want to run.
+- Launch both integration test app and driver:
+  - Launch `Integration Test` compound run configuration in Android, or
+  - Launch `Integration Test App` and `Integration Test Driver` run configurations, or
+  - Launch each application manually:
+    - Launch the integration test app manually
+    - Launch the driver: `dart test_integration/test_driver/runner.dart`
```
tiholic, line 19: There is a way to specify module(s) to runner using the `-m` or `--modules` option

`-m moduleName`, or `-m module1,module2,module3` or `-m module1 -m module2`

ex:
`VM_SERVICE_URL=http://127.0.0.1:8888/ $FLUTTER_HOME/bin/dart test_driver/runner.dart -m rest,realtime`

tiholic (changes requested):

test_integration/README.md
tiholic: I think, the above deserves a mention in README.

ben-xD: Okay good point, I found that it ran too many tests still so I didn't not use it, but it's good to document it 😅

test_integration/lib/utils/encoders.dart
ben-xD: True, but this was causing the failing/ flaky test. I didnt want to investigate the cause of the null in the timestamp since it's not an issue on the SDK side, but I really should report it. I'll take a look at this again next week

tiholic: Let's have an issue for this before landing this PR?

ben-xD: Yup, here it is: https://github.com/ably/ably-flutter/issues/151

test_integration/README.md
ben-xD: 90c2333cf6ccb059d0bef80ad7f9fcad0c50deba

tiholic (approved):

PR #163, opened 2021-09-01
Use image URLs instead of relative paths, and add note to `CONTRIBUTING.md`

ben-xD:
When referencing images in markdown files, using a local path such as `images/android.png`, for example `An android device running on API level 30` will result in the image missing on pub.dev README preview. 

Image not rendering in pub.dev readme

I've suggested an improvement https://github.com/dart-lang/pub-dev/issues/5068, but in the mean time it would be good to fix the images showing on pub.dev

**Edit:** The reason why this happens is pub.dev converts the image to the `master` branch URL, e.g. `https://github.com/ably/ably-flutter/raw/master/images/run-configuration-1.png`, assuming the default branch is master. See https://github.com/dart-lang/pub-dev/issues/5068


owenpearson (approved):
LGTM, although I think we should change back when pub.dev resolves that issue because I assume this will make the images not load if someone tries to preview the markdown locally while offline.

QuintinWillison (approved):
Nice. Thanks. 😄

Issue #156, opened 2021-08-25
Token Authentication with authCallback in Android: `java.lang.Exception: Invalid authCallback response`


ben-xD:
When using token authentication by setting the authCallback in ably-flutter running on Android devices, an exception is always thrown. Token authentication still works because this method is called until this method does return the value.

More detail: The ably-java expects a synchronous return value when it calls `Object authCallbackResponse = tokenOptions.authCallback.getTokenRequest(params);` to call the `authCallback` the user has set. In normal Android this returns a value immediately, but in Ably-flutter this is actually an asynchronous request to the dart side so it doesn’t return anything (null). Therefore, in `Auth.java` in ably-java, `authCallbackResponse` is null, and `throw AblyException.fromErrorInfo(new ErrorInfo("Invalid authCallback response", 400, 40000));` is called. 

https://github.com/ably/ably-flutter/blob/36b604fdaed87342edc2fd0c9ad94c34d362148d/android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java#L425-L452

The relevant ably-java [code](https://github.com/ably/ably-java/blob/e73a3bdb205f188baea891003cbc1af058e0db73/lib/src/main/java/io/ably/lib/rest/Auth.java#L579-L589) is:

```java
            try {
                /* the callback can return either a signed token request, or a TokenDetails */
                Object authCallbackResponse = tokenOptions.authCallback.getTokenRequest(params);
                if(authCallbackResponse instanceof String)
                    return new TokenDetails((String)authCallbackResponse);
                if(authCallbackResponse instanceof TokenDetails)
                    return (TokenDetails)authCallbackResponse;
                if(authCallbackResponse instanceof TokenRequest)
                    signedTokenRequest = (TokenRequest)authCallbackResponse;
                else
                    throw AblyException.fromErrorInfo(new ErrorInfo("Invalid authCallback response", 400, 40000)); // <-- This gets called because `authCallbackResponse` is null
```

Token authentication still works, but unfortunately the exception is not nice to see/ stack trace is scary.

I first came across this error in May 2021, but misdiagnosed it to be about "not handling clientIds properly". So I will close https://github.com/ably/ably-flutter/issues/122

To reproduce this, I have a demo application which uses token auth: https://github.com/ben-xD/ably-flutter-token or you can just try to use token auth with your own token auth server.

ben-xD:
[...]

PR #164, opened 2021-09-01
Fix `invalid authCallback response` when using token authentication

ben-xD:
Use latch to turn asynchronous method channel call synchronous. Please see https://github.com/ably/ably-flutter/issues/156 to see the exception which is thrown at **every launch of the app** when token authentication callback is first called by ably-java.

Closes https://github.com/ably/ably-flutter/issues/156


QuintinWillison (approved):
Looks like a pragmatic approach. And, presumably, ties in with the work you've done in https://github.com/ably/ably-flutter/pull/172.

Issue #141, opened 2021-07-22
Implement Push Notifications listener


ben-xD:
In ably-cocoa and ably-java, we do not provide a listener for users to react to push notification messages received in their app. However, in ably-flutter, it becomes complex to listen/ react to these messages, and users need to implement platform specific code (e.g. implementing a Service for Android, and implementing `UNUserNotificationCenterDelegate` on iOS). We should pick the best way of doing this to reduce the amount of work the user has to do on each platform, though configuration will still be required through Xcode for iOS and modifying `build.gradle` files, creating a firebase project and adding `google-services.json` for Android.

This work will continue after this PR is merged: https://github.com/ably/ably-flutter/pull/138

Issue #165, opened 2021-09-10
Use consistent indentation across all files in Java


ben-xD:
Unfortunately, some files use 2 spaces for indentation (e.g. `AblyEventStreamHandler.java`), and some use 4 spaces (e.g. `AblyFlutterPlugin.java`).

ikbalkaya:
For this one, we could create an .editorconfig for autoformatting

PR #166, opened 2021-09-10
Handle Push Notifications in Dart, including background messages, notification taps

ben-xD:
Definitely not ready for review, but I wanted to show some progress. 

The next most important step (Android only) would be to define a new **entrypoint** (not the user's main.dart/ application entry point) that would be ran when the Background Isolate is launched, and then when the entrypoint finishes running/ initializing, call the user-defined dart callback with the RemoteMessage inside the Isolate when it is ready. 

Closes https://github.com/ably/ably-flutter/issues/141


ben-xD:
I have had to add a new import (in 5f19a3b92cb52c23151c2dcd18766bd8b2eef9f4) because of a breaking change/ deprecation of a method used in testing, because upgrading my machine's Flutter version to the latest stable release (2.5.0) affected the local and CI builds.

We should migrate from this deprecated API, see: https://github.com/ably/ably-flutter/issues/173

CONTRIBUTING.md
```diff
@@ -36,6 +36,38 @@ For the moment it would not seem appropriate for our plugin to throw instances c
...
+- **Entrypoint:** In Firebase messaging, they explicitly define a new DartEntrypoint (e.g. A Callback which is launched in it's own isolate) which will be launched by the Dart VM. We do not launch our custom isolate / entrypoint, we re-use the users default entry point (their application). In Ably-Flutter, we use re-use the users main application. Therefore, we provide the same environment for message handling on both Android and iOS: their application **is running** when we handle the message.
+- **Resource consumption tradeoffs:** Firebase launches an isolate capable of only handling messages at app launch, even if they don't need it. They keep this isolate launching throughout the app, and have a queue process to queue messages. To do this, they maintain an Android Service. This allows them 10 minutes of execution time, where as on iOS the same library only has 30 seconds of execution time. Instead, we launch a new isolate on every message. Instead of creating a service and queueing work for it for each message. FCM will hold the messages until we are ready anyway (they are our queue). A new message will spawn a new engine.
```
ben-xD, line 66: Sorry, I should have said 

> Instead, we launch a new isolate on every message if the application is **not** yet running.

CONTRIBUTING.md
```diff
@@ -36,6 +36,38 @@ For the moment it would not seem appropriate for our plugin to throw instances c
...
+
+#### Push Notifications Background Message Handling
+
+Background processing in Flutter is a complicated subject that has not been explained very much in public. It involves creating Flutter Engines/isolates manually (for Android), passing messages back and forth, etc.
+
+Differences between Ably-Flutter and Firebase Messaging implementation (Android only):
+- **Entrypoint:** In Firebase messaging, they explicitly define a new DartEntrypoint (e.g. A Callback which is launched in it's own isolate) which will be launched by the Dart VM. We do not launch our custom isolate / entrypoint, we re-use the users default entry point (their application). In Ably-Flutter, we use re-use the users main application. Therefore, we provide the same environment for message handling on both Android and iOS: their application **is running** when we handle the message.
```
ben-xD, line 65: Actually, my language is incorrect. Callbacks != Entrypoints. We use the default entrypoint, where as Flutter uses a callback. They are for the same purpose: starting an isolate with specified dart code. Callback is a dart function, entrypoint is a top level main function.

android/src/main/java/io/ably/flutter/plugin/push/FirebaseMessagingReceiver.java
```diff
@@ -0,0 +1,89 @@
...
+
+import java.util.List;
+
+import io.ably.flutter.plugin.AblyFlutterPlugin;
+
+public class FirebaseMessagingReceiver extends BroadcastReceiver {
+  private static final String TAG = FirebaseMessagingReceiver.class.getName();
+
+  @Nullable
+  private static PendingResult asyncCompletionHandlerPendingResult = null;
```
ikbalkaya, line 25: Hmm. Why is this a private static ? Looks confusing to me.

android/src/main/java/io/ably/flutter/plugin/push/FirebaseMessagingReceiver.java
```diff
@@ -0,0 +1,89 @@
...
+      if (process.importance == ActivityManager.RunningAppProcessInfo.IMPORTANCE_FOREGROUND) {
+        // App is in the foreground
+        return true;
+      }
+    }
+
+    return false;
+  }
+
+  static boolean finish() {
```
ikbalkaya, line 82: Also this static method with finish() needs an explanation IMO

android/src/main/java/io/ably/flutter/plugin/util/NotificationUtilities.java
```diff
@@ -0,0 +1,191 @@
...
+
+  private interface DownloadImageCallback {
+    void finish(Bitmap bitmap);
+  }
+
+  private static void downloadImage(Uri imageUri, DownloadImageCallback callback) {
+    try {
+      InputStream in = new URL(imageUri.toString()).openStream();
+      Bitmap bitmap = BitmapFactory.decodeStream(in);
+      callback.finish(bitmap);
```
ikbalkaya, line 112: I think you should also close the stream here.

android/src/main/java/io/ably/flutter/plugin/util/NotificationUtilities.java
```diff
@@ -0,0 +1,191 @@
...
+    }
+    return builder.build();
+  }
+
+
+  private interface DownloadImageCallback {
+    void finish(Bitmap bitmap);
+  }
+
+  private static void downloadImage(Uri imageUri, DownloadImageCallback callback) {
```
ikbalkaya, line 108: In Android you normally cannot perform a network download in the main thread. This method probably should be commented to warn users about its usage

ikbalkaya: Also should move to another class as it is not directly related to notifications

android/src/main/java/io/ably/flutter/plugin/util/NotificationUtilities.java
```diff
@@ -0,0 +1,191 @@
...
+              // user visible name of the channel
+              "Miscellaneous",
+              // shows everywhere, makes noise, but does not visually intrude.
+              NotificationManager.IMPORTANCE_DEFAULT));
+    }
+
+    return FCM_FALLBACK_NOTIFICATION_CHANNEL;
+  }
+
+  private static Bundle getManifestMetadata(PackageManager packageManager, String packageName) {
```
ikbalkaya, line 179: This one also should move to another class. Not directly related to notifications

example/lib/push_notifications/push_notification_handlers.dart
```diff
@@ -1,20 +1,63 @@
...
+      print('Notification was tapped: $remoteMessage');
+    });
+  }
+
+  /// You can get the notification which launched the app by a user tapping it.
+  static void getLaunchMessage() {
+    notificationEvents.notificationTapLaunchedAppFromTerminated
+        .then((remoteMessage) {
+      if (remoteMessage != null) {
+        print('The app was launched by the user by tapping the notification');
```
QuintinWillison, line 51: I notice that here, and in `_backgroundMessageHandler`, you use multiple `print` statements... I would tend to avoid this approach when they represent the same event as there's a chance they will be separated downstream. Can you not use a single call to `print`, perhaps with a format string or something like that?

ikbalkaya (commented):
I have made some comments on particularly Java/Android code and my comments are not very comprehensive. I just commented on the obvious areas. One of my observation is that, there are lots of static methods, calls which I did not understand the full reason why. I would use static methods only for utilities and helper classes. There are also some long methods that could be shortened and some method names could benefit from changing.

Also the variables are not final where they have been declared. Not very obvious in Java but it's a best practice to declare variables final if you don't have a strong reason not to (such as runtime initialisation, assignment)
I think this code might benefit from some refactoring in the future.

QuintinWillison (changes requested):
Wow... I didn't expect this pull request to be as big as this. It's taken me several hours to review and I know I've still not reviewed it at anywhere near the depth I would like to. Much of my comments are nit picking and/or pattern spotting.

In terms of the overall architecture and design of your implementation, I'll be honest that I've not really been able to build a good mental model of that. This is, perhaps, why you were hinting in this morning's standup that a 'code review call' might be useful. I'm not sure how this pull request could have been broken into smaller chunks, but that might have helped. I'm trying to spend a fraction of a day to understand something that you've spent weeks curating - I was always going to be on a losing streak there!

Something I've noticed across a number of the Java files is that we appear to have indentation at 2 spaces, not 4 spaces. Were we to introduce checkstyle to this codebase, specifically [IndentationCheck](https://checkstyle.sourceforge.io/apidocs/com/puppycrawl/tools/checkstyle/checks/indentation/IndentationCheck.html), then this would fail - the standard in the Java world is 4 spaces.

android/src/main/java/io/ably/flutter/plugin/AblyFlutterPlugin.java
```diff
@@ -65,4 +82,55 @@ public void onDetachedFromEngine(@NonNull FlutterPluginBinding binding) {
...
+    @Override
+    public boolean onNewIntent(Intent intent) {
+        mainActivity.setIntent(intent);
+        doIfIntentContainsRemoteMessage(intent, (message) -> {
+            methodChannel.invokeMethod(PlatformConstants.PlatformMethod.pushOnNotificationTap, message);
+        });
+        return false;
+    }
+
+    private Boolean doIfIntentContainsRemoteMessage(Intent intent, CompletionHandlerWithRemoteMessage completionHandler) {
```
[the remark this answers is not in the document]
ben-xD: I was conflicted about this name as well, but settled on it because I didn't want to write `if (extras == null)` in multiple locations and wanted to encapsulate the condition. Do you have a suggestion for the name?

ben-xD:
@QuintinWillison , you may have noticed some bad formatting and inconsistent indentation in the Java code. Unfortunately some files are indented with 2 spaces, and others use 4 spaces. I created an issue here: https://github.com/ably/ably-flutter/issues/165 Because of this, Java formatting is currently in a bad state, because I have go into settings and change the value to match the file, otherwise the whole file is changed due to auto format. Java formatting will be fixed for all files in that issue.

lib/src/platform/src/push_notification_events_native.dart
```diff
@@ -0,0 +1,79 @@
...
+
+  /// Implementation of setOnBackgroundMessage. For more documentation,
+  /// see [PushNotificationEvents.setOnBackgroundMessage]
+  void setOnBackgroundMessage(BackgroundMessageHandler handler) async {
+    _onBackgroundMessage = handler;
+    if (io.Platform.isAndroid) {
+      try {
+        await BackgroundIsolateAndroidPlatform.methodChannel
+            .invokeMethod(PlatformMethod.pushSetOnBackgroundMessage);
+      } on MissingPluginException {
```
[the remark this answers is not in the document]
ben-xD: The android side is currently only listening for this method when the isolate was launched **manually** by the Android side. So this code throwing an exception is actually the norm when an app starts, but happens once per launch. The only time this method channel works is when the Activity is not running. So currently, it is intentional that user does not see this error when the app launches normally. Unfortunately, if something breaks in the case where the app is not launched normally (isolate launched manually), we won't know about it. I initially opted to not added extra logic code to handle this on the platform side, since it can be kept entirely in the dart side.

An easy solution to this is to add a listener on the android side to listen for this when the isolate was launched with the Activity (a normal app/ **not manually**), and do nothing as a result of this method call, but this would add complexity/ code. To avoid swallowing errors, I'll implement this.

ios/Classes/PushNotificationEventHandlers.swift
```diff
@@ -0,0 +1,75 @@
...
+    @objc(initWithDelegate:andMethodChannel:) public init(_ delegate: UNUserNotificationCenterDelegate, _ methodChannel: FlutterMethodChannel) {
+        self.delegate = delegate;
+        self.methodChannel = methodChannel;
+    }
+    
+    public func userNotificationCenter(_ center: UNUserNotificationCenter, willPresent notification: UNNotification, withCompletionHandler completionHandler: @escaping (UNNotificationPresentationOptions) -> Void) {
+        let message = RemoteMessage.fromNotificationContent(content:notification.request.content);
+        methodChannel.invokeMethod(AblyPlatformMethod_pushOnShowNotificationInForeground, arguments: message) { result in
+            if let result = result as? NSNumber, result == NSNumber(value: true) {
+                if #available(iOS 14.0, *) {
```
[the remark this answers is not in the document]
ben-xD: There's only one method being called, and one enum case being used. A developer can check the definition of the method or enum and see that there are iOS version restrictions. In this case, looking at `.banner`, we will see that it requires iOS 14+: https://developer.apple.com/documentation/usernotifications/unnotificationpresentationoptions/3564812-banner. It's quite a common attribute in iOS apps, I've seen it many times and think it should be assumed knowledge.

CleanShot 2021-09-20 at 23 31 24

There are more usages of `#available` in the code and potentially more in the future, and it is the first keyword in the keyword list: https://docs.swift.org/swift-book/ReferenceManual/Attributes.html

android/src/main/AndroidManifest.xml
```diff
@@ -1,3 +1,25 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
   package="io.ably.flutter.plugin">
+    <application>
+        <!-- Receive new tokens if firebase_messaging flutter library is not installed-->
+        <service android:name="io.ably.flutter.plugin.push.PushMessagingService"
+            android:exported="false">
+            <!-- We want a lower priority than firebase_messaging firebase library (priority 0),
+            since it will send a intent broadcast when it receives a new token which we can listen
+            to it. This service should only run on an app without firebase_messaging flutter
+            library. -->
```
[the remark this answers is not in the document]
ben-xD: Thanks for mentioning, its a completely useless comment. I had not thoroughly gone through the changes in this PR before taking it out of draft. It's been completely removed now in 21d94b4a7080344161b06ab0ffb1d7b53ad3ca2e

android/src/main/AndroidManifest.xml
```diff
@@ -1,3 +1,25 @@
...
+            <intent-filter android:priority="-1">
+                <action android:name="com.google.firebase.MESSAGING_EVENT"/>
+            </intent-filter>
+        </service>
+        <!-- TODO register a runtime broadcast receiver for io.flutter.plugins.firebase.messaging.TOKEN. -->
```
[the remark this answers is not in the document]
ben-xD: This has been removed in 21d94b4a7080344161b06ab0ffb1d7b53ad3ca2e. Thanks for catching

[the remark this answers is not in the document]
ben-xD: Sorry, it's outdated. Removed in 21d94b4a7080344161b06ab0ffb1d7b53ad3ca2e

android/src/main/java/io/ably/flutter/plugin/AblyFlutterPlugin.java
```diff
@@ -65,4 +82,55 @@ public void onDetachedFromEngine(@NonNull FlutterPluginBinding binding) {
...
+        Bundle extras = intent.getExtras();
+        if (extras == null) {
+            return false;
+        }
+        RemoteMessage message = new RemoteMessage(intent.getExtras());
+        // TODO should i sent a RemoteMessage if only notificaiton is populated
```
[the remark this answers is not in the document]
ben-xD: I've manually unindented those lines, since autoformatting the file changed every line in the file 😅

The todo was actually still relevant, but I've fixed it now. I now only send a message to the dart side if the dictionary is not empty. Notifications only messages are not sent to the dart side, to be consistent with iOS. 

55f111fd8fa15a4307b2ed26b494311bea508137

ios/Classes/AblyFlutterPlugin.m
```diff
@@ -667,8 +662,18 @@ -(void)registerWithCompletionHandler:(const FlutterResult)completionHandler {
...
     [[UIApplication sharedApplication] registerForRemoteNotifications];
+    // Check if application was launched from a notification tap.
+    
+    // https://stackoverflow.com/a/21611009/7365866
+    NSDictionary *notification = [launchOptions objectForKey:UIApplicationLaunchOptionsRemoteNotificationKey];
+    if (notification) {
+        PushHandlers.pushNotificationTapLaunchedAppFromTerminatedData = notification;
+    }
+    
+    NSLog(@"didFinishLaunchingWithOptions: %@, \n %@", launchOptions, launchOptions.allKeys, launchOptions.allValues);
```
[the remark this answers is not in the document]
ben-xD: That's interesting, thanks. This log line was added just for debugging, and it's been removed now: 4ce28aef27a72a3f72f5e7e8293493d4df44390b

ios/Classes/PushNotificationEventHandlers.swift
```diff
@@ -0,0 +1,75 @@
...
+            }
+        }
+    }
+    
+    public func userNotificationCenter(_ center: UNUserNotificationCenter, didReceive response: UNNotificationResponse, withCompletionHandler completionHandler: @escaping () -> Void) {
+        let remoteMessage = RemoteMessage.fromNotificationContent(content: response.notification.request.content)
+        methodChannel.invokeMethod(AblyPlatformMethod_pushOnNotificationTap, arguments: remoteMessage, result: nil)
+        
+        if let delegate = delegate {
+            // TODO Call dart side to say error, we did not call this because the delegate was already set.
```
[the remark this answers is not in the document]
ben-xD: I updated this file earlier today, and that todo comment has been removed. Additional bug fixes were added to this file: 43eb6e9bf9ce11c09e437e5dca94184211f50d9a

ios/Classes/PushNotificationEventHandlers.swift
```diff
@@ -0,0 +1,75 @@
...
+    @available(iOS 12.0, *)
+    public func userNotificationCenter(_ center: UNUserNotificationCenter, openSettingsFor notification: UNNotification?) {
+        if let delegate = delegate {
+            delegate.userNotificationCenter?(center, openSettingsFor: notification)
+        }
+    }
+    
+    @objc
+    public func application(_ application: UIApplication, didReceiveRemoteNotification userInfo: [AnyHashable : Any], fetchCompletionHandler completionHandler: @escaping (UIBackgroundFetchResult) -> Void) {
+        // TODO check if foreground, background or inactive? See https://developer.apple.com/documentation/uikit/uiapplicationstate?language=objc
```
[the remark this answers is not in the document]
ben-xD: Repeating a comment I said at https://github.com/ably/ably-flutter/pull/166#discussion_r712564915
> I updated this file earlier today, and that todo comment has been removed. Additional bug fixes were added to this file: 43eb6e9bf9ce11c09e437e5dca94184211f50d9a

example/lib/push_notifications/push_notification_handlers.dart
ben-xD: Thanks. I've refactored this one to print just one line, but also to show a toast when the error happens for the user, since this is in the example app: 32d4d2aa7c6e44ad2afd4ded1c10660c157bee84

ios/Classes/AblyFlutterPlugin.m
```diff
@@ -685,8 +690,10 @@ - (void)application:(UIApplication *)application didFailToRegisterForRemoteNotif
 }
 
 - (BOOL)application:(UIApplication *)application didReceiveRemoteNotification:(NSDictionary *)userInfo fetchCompletionHandler:(void (^)(UIBackgroundFetchResult))completionHandler {
-    // This needs to return NO allow the message to arrive to the users application AppDelegate.
-    return NO;
+    [_pushNotificationEventHandlers application:application didReceiveRemoteNotification:userInfo fetchCompletionHandler:completionHandler];
+    // TODO try this, to give other plugins and AppDelegate a chance to receive the remote notification as well. If NO/ false is returned, the user's didReceiveRemoteNotification delegate method will be called
```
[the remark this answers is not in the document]
ben-xD: Oh this is one I left recently because it requires some thought. In firebase_messaging, if the APNs message has `userInfo[@"gcm.message_id"]`, then it returns `YES`, and the users app delegate method **won't** be informed of the remote notification. However, we want to allow customers to implement app delegate methods (handle messages natively) if they choose to, even if our plugin is installed. So we would have to return `NO`, that our plugin did not handle the message. 

I've done that here: d562fd4c1ef9f257412a16033acb17ec928fabc2

CONTRIBUTING.md
ben-xD: 8a790f2eb1990a0599d25c5aa614458708c96566

CONTRIBUTING.md
ben-xD: 215976dcaffff7f9a2e07ff5d1d7439b939c5dd7

PushNotifications.md
```diff
@@ -236,15 +234,82 @@ final message = ably.Message(
...
+```dart
+ably.Push.notificationEvents.onNotificationTap.listen((remoteMessage) {
+  print('Notification was tapped: $remoteMessage');
+});
+```
 
 #### Data Message / Background Notification
 
-ably-flutter currently does not pass the messages to the Flutter application/ Dart-side, so users will need to listen to messages in each platform. See [Implement Push Notifications listener](https://github.com/ably/ably-flutter/issues/141) for more information.
+- When the app is in the foreground, you can listen to messages using:
```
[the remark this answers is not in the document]
ben-xD: This was done among other improvements in e2a1958402cc43cb7e7a049b4fb1c0dcce9c8ba4 earlier today. I just didn't push it until more recently

PushNotifications.md
```diff
@@ -347,6 +412,12 @@ This means your registration token is invalid. Ably is may not have your device'
 ActivationContext.getActivationContext(this).onNewRegistrationToken(RegistrationToken.Type.FCM, registrationToken);
 ```
 
+### My notifications do not make a sound when they are shown on the device.
```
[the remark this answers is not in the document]
ben-xD: 53b14ed4ab49a35e6143482893c098fd66706a2b

👍

This is the troubleshooting section which is styled as an FAQ in many ways

android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
```diff
@@ -797,4 +802,20 @@ private String encodePresenceAction(PresenceMessage.Action action) {
     return jsonMap;
   }
 
+  private Map<String, Object> encodeRemoteMessage(RemoteMessage message) {
```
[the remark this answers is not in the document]
ben-xD: I did create an issue to address this in the future: https://github.com/ably/ably-flutter/issues/174 This requires some thought to design a class which contains data from iOS and Android. Also, it might be challenging (but possible) to get some fields on both platforms, whereas some are impossible (e.g. priority on Android vs. iOS).

I didn't document that we are only passing certain fields of an **FCM** RemoteMessage, primarily because on iOS there is **very little** to give anyway: its just an NSDictionary. An ably remote message is not the same as an FCM RemoteMessage, so I didn't think it was worth mentioning we lack some FCM RemoteMessage fields.

IMHO the MVP would be this RemoteMessage, the notification and the data map.

android/src/main/java/io/ably/flutter/plugin/push/PushBackgroundIsolateRunner.java
```diff
@@ -0,0 +1,101 @@
...
+            }
+          }
+      );
+    } else {
+      result.notImplemented();
+    }
+  }
+
+  private void finish() {
+    assert flutterEngine != null;
```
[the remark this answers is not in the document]
ben-xD: I've used it because it is an assumption in that the FlutterEngine was running when you call finish, and if the assumption breaks that is a SDK developer mistake. If I just fail early, I'm telling users of this function that it's okay to mutate `flutterEngine`, close it early, or just set it to `null`. I don't want this, because `finish()` is the only thing that should nullify `flutterEngine`. TBH, it doesn't matter because finish also calls `FirebaseMessagingReceiver.finish();`, which kills the entire Receiver. Most importantly, what happens if `flutterEngine` was set to `null`, but was not `destroy`'d? The assert enforces a contract, a user should not be mutating `flutterEngine`.

It also only happens in debug mode: https://stackoverflow.com/a/6176529/7365866 and https://stackoverflow.com/a/2758645/7365866

android/src/main/java/io/ably/flutter/plugin/push/FirebaseMessagingReceiver.java
```diff
@@ -0,0 +1,89 @@
...
+      // No existing Flutter Activity is running, create a FlutterEngine and pass it the RemoteMessage
+      new PushBackgroundIsolateRunner(context, this, message);
+    }
+  }
+
+  Boolean isApplicationInForeground(Context context) {
+    ActivityManager activityManager = (ActivityManager) context.getSystemService(Context.ACTIVITY_SERVICE);
+    List<ActivityManager.RunningAppProcessInfo> appProcesses = activityManager.getRunningAppProcesses();
+    // This only shows processes for the current android app.
+    assert (appProcesses.size() == 1); // We have not tested multiple processes running on 1 app.
```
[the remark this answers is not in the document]
ben-xD: It's an assumption I'm making. Investigating cases where appProcesses might be not equal to 1 is complicated, hence I stuck an assert there to tell developers I have not tested/ know of ways to get multiple processes.

I've asked this question on Stack Overflow 10 days ago, with no responses: https://stackoverflow.com/questions/69131027/in-what-cases-is-there-no-process-on-android-when-calling-activitymanagers-getr

It also only happens in debug mode: https://stackoverflow.com/a/6176529/7365866 and https://stackoverflow.com/a/2758645/7365866

CONTRIBUTING.md
```diff
@@ -36,6 +36,38 @@ For the moment it would not seem appropriate for our plugin to throw instances c
...
+- **Resource consumption tradeoffs:** Firebase launches an isolate capable of only handling messages at app launch, even if they don't need it. They keep this isolate launching throughout the app, and have a queue process to queue messages. To do this, they maintain an Android Service. This allows them 10 minutes of execution time, where as on iOS the same library only has 30 seconds of execution time. Instead, we launch a new isolate on every message. Instead of creating a service and queueing work for it for each message. FCM will hold the messages until we are ready anyway (they are our queue). A new message will spawn a new engine.
+- **Execution / wall clock time:** We have an execution time of 30 seconds on both Android and iOS for each message. Firebase messaging has an execution time of 10 minutes for all consecutive messages on Android between the service runs out of execution time, and 30 seconds on iOS. If users want more execution time, we recommend [package:workmanager](https://pub.dev/packages/workmanager).
```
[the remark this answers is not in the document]
ben-xD: I need to test/ investigate this limitation further, by experimenting with Firebase messaging. What I mean is: Firebase messaging launches an Android Service (`FlutterFirebaseMessagingBackgroundService` which is a [JobIntentService](https://developer.android.com/reference/androidx/core/app/JobIntentService)) which hosts an Isolate. However, these services can't live forever, and have been documented to live around **10 minutes**:
https://stackoverflow.com/questions/48630606/how-long-is-the-jobservice-execution-time-limit-mentioned-in-androids-jobinte

Once the service is stopped for going past its execution time limit, its not known if the service will be launched by Android immediately, or some time in the future. 

I didn't want to investigate Firebase messaging/ find a bug in it. I have elaborated that section in e5ac8443118b0000d936df8f70ed469138742b0a

android/src/main/java/io/ably/flutter/plugin/AblyFlutterPlugin.java
```diff
@@ -1,28 +1,43 @@
...
 import io.flutter.plugin.common.BinaryMessenger;
 import io.flutter.plugin.common.MethodChannel;
 import io.flutter.plugin.common.MethodCodec;
+import io.flutter.plugin.common.PluginRegistry;
 import io.flutter.plugin.common.PluginRegistry.Registrar;
 import io.flutter.plugin.common.StandardMethodCodec;
 
-public class AblyFlutterPlugin implements FlutterPlugin {
+public class AblyFlutterPlugin implements FlutterPlugin, ActivityAware, PluginRegistry.NewIntentListener {
+    public static Boolean isActivityRunning = false;
```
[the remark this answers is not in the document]
ben-xD: From reading the code on Github, it may not be clear, but in the code, `isActivityRunning` (now `isMainActivityRunning`) is used in `FirebaseMessagingReceiver`, which is a statically declared (launched even when the app is off) broadcast receiver. Because there is no guarantee that MainActivity/ FlutterActivity is running when this broadcast receiver is launched (and by extension the user's Dart code), the broadcast receiver **needs to know if it should launch the user's dart application**. This is documented in `FirebaseMessagingReceiver`'s `onReceive` method:
```java
    } else if (AblyFlutterPlugin.isMainActivityRunning) {
      // Flutter is already running, just send a background message to it.
      Intent onMessageReceivedIntent = new Intent(PUSH_ON_BACKGROUND_MESSAGE_RECEIVED);
      onMessageReceivedIntent.putExtras(intent.getExtras());
      LocalBroadcastManager.getInstance(context).sendBroadcast(onMessageReceivedIntent);
    } else {
```

# Another problem
In theory though, there can be many AblyFlutterPlugins running, so **just** **a boolean** is not good here. This case only happens when a user creates another FlutterActivity or FlutterFragment which (automatically if the user doesn't configure it) registers new plugin instances, including a new AblyFlutterPlugin. This would happen if a user is using Flutter in their Android app ("add-to-app" use cases).

Instead, we'd need to keep track of the counts of FlutterEngines. If it is 0, create one, otherwise do not. Each AblyFlutterPlugin will increment/ decrement this counter inside the `ActivityAware` interface methods, specifically: 

Or more simply to remove this static field, always create a new FlutterEngine if the app is the background, even if the app is already running. This would be slightly wasteful. 

TLDR: WDYT? 
- Add complexity to add support for add-to-app (multiple FlutterEngines)
- Remove complexity whilst adding support by always launching a FlutterEngine, even when not needed
- Keep this static field, and do not support add-to-app

android/src/main/AndroidManifest.xml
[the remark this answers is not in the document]
ben-xD: I need to check if it's actually useless, it might be important 😅

android/src/main/java/io/ably/flutter/plugin/AblyFlutterPlugin.java
```diff
@@ -1,28 +1,43 @@
...
 
     @Override
     public void onAttachedToEngine(@NonNull FlutterPluginBinding flutterPluginBinding) {
         applicationContext = flutterPluginBinding.getApplicationContext();
-        MethodChannel methodChannel = setupChannels(flutterPluginBinding.getBinaryMessenger(), applicationContext);
-        PushActivationEventHandlers.instantiate(applicationContext, methodChannel);
+        setupChannels(flutterPluginBinding.getBinaryMessenger(), applicationContext);
     }
 
     // This static function is optional and equivalent to onAttachedToEngine. It supports the old
```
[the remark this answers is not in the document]
ben-xD: This is meant to be a static function. It is a function used by a generated class for old Flutter Android apps. It's unlikely many users still need this support, given that the v2 embeddings API was released 3 years ago: https://github.com/flutter/engine/blame/233a6e01fbb0f3cb407da7bee6817ccdcb6a9822/shell/platform/android/io/flutter/embedding/android/FlutterActivity.java#L4

I have fixed it in a6691ca42d4d0d6826a6ab0e89ad851d12069f05, to keep the support anyway, but it is untested. I suggest we remove it completely: https://github.com/ably/ably-flutter/issues/178

android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java
```diff
@@ -36,26 +39,23 @@
...
-  static synchronized AblyMethodCallHandler getInstance(final MethodChannel channel, final OnHotRestart listener, final Context applicationContext) {
-    if (null == _instance) {
-      _instance = new AblyMethodCallHandler(channel, listener, applicationContext);
-    }
-    return _instance;
-  }
-
   private final MethodChannel channel;
   private final OnHotRestart onHotRestartListener;
   private final Map<String, BiConsumer<MethodCall, MethodChannel.Result>> _map;
```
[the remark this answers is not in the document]
ben-xD: _map is a method of method strings to implementations. If the java side receives this string when method is invoked from the Dart side, the implementation is called. 

The reason why its `_map` (quite short), is because it used very often, where each line is also quite long, e.g.: `_map.put(PlatformConstants.PlatformMethod.getPlatformVersion, this::getPlatformVersion);`. 

It's also possible to remove the map completely, and just use a switch statement instead of having map accesses. This is many plugins implement it. It looks much clearer, but the lines will be double as long (because switch statements take up more space 😅).

WDYT, should it be short or longer?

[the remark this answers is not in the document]
ben-xD: https://github.com/ably/ably-flutter/issues/180

android/src/main/java/io/ably/flutter/plugin/push/CompletionHandlerWithRemoteMessage.java
```diff
@@ -0,0 +1,7 @@
+package io.ably.flutter.plugin.push;
+
+import com.google.firebase.messaging.RemoteMessage;
+
+public interface CompletionHandlerWithRemoteMessage {
```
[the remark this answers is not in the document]
ben-xD: They are quite separate use cases (We discussed it in a video call), one is just a callback that never expects `RemoteMessage`, the other one always expects it.

android/src/main/java/io/ably/flutter/plugin/push/FirebaseMessagingReceiver.java
```diff
@@ -0,0 +1,89 @@
...
+
+public class FirebaseMessagingReceiver extends BroadcastReceiver {
+  private static final String TAG = FirebaseMessagingReceiver.class.getName();
+
+  @Nullable
+  private static PendingResult asyncCompletionHandlerPendingResult = null;
+
+  @Override
+  public void onReceive(Context context, Intent intent) {
+    Boolean isApplicationInForeground = isApplicationInForeground(context);
```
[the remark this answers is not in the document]
ben-xD: This was done in 537d28454445ac51df1190db3833f65e7084e47a, thanks

android/src/main/java/io/ably/flutter/plugin/AblyFlutterPlugin.java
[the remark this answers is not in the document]
ben-xD: This was renamed to `handleRemoteMessageIntent`, in 1ddee9b3edde22bc1c28451cfced568f49b8a2e1.

android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java
```diff
@@ -36,26 +39,23 @@
...
 
-  static synchronized AblyMethodCallHandler getInstance(final MethodChannel channel, final OnHotRestart listener, final Context applicationContext) {
-    if (null == _instance) {
-      _instance = new AblyMethodCallHandler(channel, listener, applicationContext);
-    }
-    return _instance;
-  }
-
   private final MethodChannel channel;
   private final OnHotRestart onHotRestartListener;
```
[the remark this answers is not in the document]
ben-xD: This was done in f86061da5d8019ea3cc3ee4ff3c2f80dc868863a

android/src/main/java/io/ably/flutter/plugin/push/FirebaseMessagingReceiver.java
ben-xD: It's not static anymore, thanks for catching that: 3d93df4513c4ca7127184bcddb9e6940c6286acc

android/src/main/java/io/ably/flutter/plugin/push/FirebaseMessagingReceiver.java
ben-xD: It's not static anymore, thanks for catching that: 3d93df4

CONTRIBUTING.md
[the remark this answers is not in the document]
ben-xD: dec7a62642434d8eb3d79a2f424269e76335aceb. Some issues were fixed by changes/improvements.

android/src/main/java/io/ably/flutter/plugin/push/CompletionHandler.java
```diff
@@ -0,0 +1,6 @@
+package io.ably.flutter.plugin.push;
+
+interface CompletionHandler {
```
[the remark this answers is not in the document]
ben-xD: Thanks, good idea:

android/src/main/java/io/ably/flutter/plugin/AblyFlutterPlugin.java
```diff
@@ -65,4 +82,55 @@ public void onDetachedFromEngine(@NonNull FlutterPluginBinding binding) {
     private static MethodCodec createCodec() {
         return new StandardMethodCodec(new AblyMessageCodec());
     }
+
+    @Override
```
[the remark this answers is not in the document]
ben-xD: Done in 6910e8b23d9474561b582e58f498cc04f936c333

lib/src/platform/src/push_notification_events_native.dart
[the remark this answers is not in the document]
ben-xD: I've made the above changes in: 70251990f2c6e96cecf11bbe74096cf84929c712, "Replace exception catching with background method call handler which ignores the method call".

android/src/main/java/io/ably/flutter/plugin/util/NotificationUtilities.java
ben-xD: Most of this file contents is taken from Firebase, and I wanted to put in the same file. It's a private method.

android/src/main/java/io/ably/flutter/plugin/util/NotificationUtilities.java
ben-xD: Most of this file contents is taken from Firebase, and I wanted to put in the same file. It's a private method. Now, I'm even thinking about removing this entire file actually before this PR merges.

android/src/main/java/io/ably/flutter/plugin/util/NotificationUtilities.java
ben-xD: This file was deleted, because the entire functionality of this file IMHO should not be in the project. The functionality: Firebase Android SDK does not show notifications when the app is in the foreground, but on iOS it is customisable. I copied some code from Firebase and modified it to allow users to configure it. 

I have removed this now: c5070b3c4e0afc964ac5a30b851b85d2145e7015

android/src/main/java/io/ably/flutter/plugin/util/NotificationUtilities.java
ben-xD: Repeating a comment from https://github.com/ably/ably-flutter/pull/166#discussion_r716573968:
> This file was deleted, because the entire functionality of this file IMHO should not be in the project. The functionality: Firebase Android SDK does not show notifications when the app is in the foreground, but on iOS it is customisable. I copied some code from Firebase and modified it to allow users to configure it.
>
> I have removed this now: c5070b3

android/src/main/java/io/ably/flutter/plugin/util/NotificationUtilities.java
ben-xD: Repeating a comment from https://github.com/ably/ably-flutter/pull/166#discussion_r716573968:
> This file was deleted, because the entire functionality of this file IMHO should not be in the project. The functionality: Firebase Android SDK does not show notifications when the app is in the foreground, but on iOS it is customisable. I copied some code from Firebase and modified it to allow users to configure it.
>
> I have removed this now: c5070b3

android/src/main/java/io/ably/flutter/plugin/push/PushBackgroundIsolateRunner.java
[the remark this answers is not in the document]
ikbalkaya: I don't know much about the mechanism of Flutter engine, but why do you allow users to mutate this? Should this not be a part of some sort of lifecycle?

android/src/main/java/io/ably/flutter/plugin/push/FirebaseMessagingReceiver.java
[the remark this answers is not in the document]
ikbalkaya: I also don't think that asserts should be in production code unlesss it is deifinitely necessary. There might be a better way to check whether the application is running 
https://stackoverflow.com/a/5862048
This may not be very relevant to notifications. I think something we can disccuss

android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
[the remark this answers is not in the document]
ikbalkaya: Again, not completely sure about the details but if there is an exception / exclusion from spec, it should be documented. Maybe we can quickly discuss the design of this one as well?

ikbalkaya (changes requested):
I have gone through the comments and resolved most of them. However I think some still a bit more attention. Probably worth us a call to quickly discuss

android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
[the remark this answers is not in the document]
ben-xD: I don't think it's related to the feature spec or exceptions in this case though 😅 :)

android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java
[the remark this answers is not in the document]
ikbalkaya: bf5e52b3ee634bb708ae759b8cdb27b707c7c7c6

android/src/main/java/io/ably/flutter/plugin/push/CompletionHandlerWithRemoteMessage.java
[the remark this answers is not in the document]
ikbalkaya: 413732afe02405be9fc58493852a7d7d95a5050f

android/src/main/java/io/ably/flutter/plugin/push/CompletionHandler.java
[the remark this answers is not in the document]
ikbalkaya: I have not used Runnable except for in context of running new threads. Not sure if it's a good idea to use it in this context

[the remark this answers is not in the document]
ikbalkaya: b31593c1027c2b30f11bf84516e801c86b917a68

android/src/main/java/io/ably/flutter/plugin/push/FirebaseMessagingReceiver.java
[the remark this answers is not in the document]
ikbalkaya: c32ce6296f4fbf16271efb6bacbe194087987a6c

android/src/main/java/io/ably/flutter/plugin/push/PushBackgroundIsolateRunner.java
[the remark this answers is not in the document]
ikbalkaya: 26a1862b2a7c44a75574d84f5a326fd3dce5a1db

ikbalkaya (approved):
My comments were addressed properly. This approval does not reflect Quintin's change requests

lib/src/platform/src/push_notification_events_native.dart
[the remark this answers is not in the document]
QuintinWillison: Thanks. I can see that required a fair bit of engineering to work around, but hopefully worth it. 🤞

ios/Classes/PushNotificationEventHandlers.swift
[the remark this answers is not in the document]
QuintinWillison: It was only a suggestion and, honestly, the time you've spent explaining this to me in a response to my comment could perhaps have been spent adding commentary next to the code in question. 🤷‍♂️

example/lib/push_notifications/push_notification_handlers.dart
QuintinWillison: Nice. 👍

android/src/main/java/io/ably/flutter/plugin/push/PushBackgroundIsolateRunner.java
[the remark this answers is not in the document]
QuintinWillison: Thanks, both. 😄

android/src/main/java/io/ably/flutter/plugin/push/CompletionHandler.java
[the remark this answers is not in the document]
QuintinWillison: Sounds like this one went full circle. 😆 I'm guessing you two discussed so I shall defer to your combined judgement.

_exits stage left..._

android/src/main/java/io/ably/flutter/plugin/AblyFlutterPlugin.java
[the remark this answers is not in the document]
QuintinWillison: If I'm reading your responses correctly then I think:

1. Yes, "Keep this static field, and do not support add-to-app"
2. Also, "create an issue to add support for Add-to-app (very low priority)"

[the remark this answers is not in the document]
QuintinWillison: @ben-xD feel free to resolve this conversation on my behalf once you're satisfied that it's settled - ideally adding a link to the issue you create here for the record.

CONTRIBUTING.md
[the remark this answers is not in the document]
QuintinWillison: Nice. Thanks. 😄

QuintinWillison (approved):

CONTRIBUTING.md
```diff
@@ -36,6 +36,38 @@ For the moment it would not seem appropriate for our plugin to throw instances c
 of the [Error class](https://api.dart.dev/stable/2.4.0/dart-core/Error-class.html), as this is perhaps more
 appropriate for programmer's failures on the Dart side of things. But time will tell.
 
+### Push Notifications 
+
+#### Debugging the app from launch
```
marklewin, line 41: It would be really nice to have screenshots for both iOS and Android here!

CONTRIBUTING.md
```diff
@@ -36,6 +36,38 @@ For the moment it would not seem appropriate for our plugin to throw instances c
...
+- iOS: In Xcode, build the Xcode app scheme with launch option: "Wait for the executable to be launched" instead of "Automatically". You need to repeat this for every launch. Then "Attach Flutter" in Android Studio or CLI if you want to, as well. 
+
+#### Push Notifications activation / deactivation
+
+The platform SDKs ([ably-android](https://github.com/ably/ably-java), [ably-cocoa](https://github.com/ably/ably-cocoa)) provide ways to check if device activation, deactivation or registration update fails. On Android, these errors are sent in Intents which you should register for at runtime. In Cocoa, your errors are provided through `ARTPushRegistererDelegate` methods. However, this error does not get returned immediately/quickly in all cases. For example, if there was no internet connection, then `Push.activate()` will not throw an error, it will just block forever, because errors are not provided by the SDKs. Once an internet connection is made, the intent will be sent / delegate methods will be called.
```
marklewin, line 48:
```suggestion
The platform SDKs ([ably-android](https://github.com/ably/ably-java) and [ably-cocoa](https://github.com/ably/ably-cocoa)) enable you to check if device activation, deactivation, or registration update fails. On Android, these errors are sent in "intents" which you should register for at runtime. In Cocoa, your errors are provided through `ARTPushRegistererDelegate` methods. However, this error does not get returned immediately/quickly in all cases. For example, if there was no internet connection, then `Push.activate()` will not throw an error, it will just block forever, because errors are not provided by the SDKs. Once an internet connection is made, the intent will be sent / delegate methods will be called.
```

marklewin: Also, it's not clear (at least to me) whether the text that starts "However, this error does not get returned..." applies to both Android and Cocoa, or just the latter.

CONTRIBUTING.md
```diff
@@ -36,6 +36,38 @@ For the moment it would not seem appropriate for our plugin to throw instances c
...
+
+Implementation wise, we do this by passing a reference to the `FlutterResult` we can use to pass back the result, when. This makes it convenient for users: they can `await push.activate()`. However, they should not rely on this Future completing, in the case of network issues.
```
marklewin, line 50: "Implementation wise" should be hyphenated, but I suggest not using it because it might not be clear to non-native English speakers.

marklewin: Take care to capitalize only proper nouns (names of products, etc). E.g. "Future" and (above) "Intents"

CONTRIBUTING.md
```diff
@@ -36,6 +36,38 @@ For the moment it would not seem appropriate for our plugin to throw instances c
...
+
+#### Notification tap handling
+
+Android's firebase-messaging / FCM library allows users to select the Intent action used when the automatically generated notification is tapped by the user. You can do this by setting `fcm.notification.click_action` in Ably's push payload. However, for this to work, users would need to declare the intent action within their `AndroidManifest.xml`. Therefore, we don't really tell users they can modify `click_action` and configure it. However, they can do so if they wish.
```
marklewin, line 54: allows -> enables ("Allows" implies permission)

automatically generated -> automatically-generated

CONTRIBUTING.md
```diff
@@ -36,6 +36,38 @@ For the moment it would not seem appropriate for our plugin to throw instances c
...
+
+#### Notifications generation for Foreground Apps
+
+iOS allows users to show the notification even if the app is in the foreground, by calling a delegate method ([`userNotificationCenter(_:willPresent:withCompletionHandler:)`](https://developer.apple.com/documentation/usernotifications/unusernotificationcenterdelegate/1649518-usernotificationcenter)) which the user can choose to show the message to the user, based on the notification content. FCM / Android does not provide this functionality. Users can only configure this behaviour for iOS, by using `PushNotificationEvents#setOnShowNotificationInForeground`.
```
marklewin, line 58: allows -> enables

CONTRIBUTING.md
```diff
@@ -36,6 +36,38 @@ For the moment it would not seem appropriate for our plugin to throw instances c
...
+
+#### Notifications generation for Foreground Apps
+
+iOS allows users to show the notification even if the app is in the foreground, by calling a delegate method ([`userNotificationCenter(_:willPresent:withCompletionHandler:)`](https://developer.apple.com/documentation/usernotifications/unusernotificationcenterdelegate/1649518-usernotificationcenter)) which the user can choose to show the message to the user, based on the notification content. FCM / Android does not provide this functionality. Users can only configure this behaviour for iOS, by using `PushNotificationEvents#setOnShowNotificationInForeground`. 
+
+#### Push Notifications Background Message Handling
+
+Background processing in Flutter is a complicated subject that has not been explained very much in public. It involves creating Flutter Engines/isolates manually (for Android), passing messages back and forth, etc.
```
marklewin, line 62: in public?

CONTRIBUTING.md
```diff
@@ -36,6 +36,38 @@ For the moment it would not seem appropriate for our plugin to throw instances c
...
+
+Differences between ably-Flutter and Firebase Messaging implementation (Android only):
+- **Isolate code:** In Firebase Messaging, they explicitly define a callback which is launched in it's own isolate. We do not launch a custom entrypoint/ dart code, but instead re-use the user's default entrypoint (their app code), by using `DartExecutor.DartEntrypoint.createDefault()`. Therefore, we provide the same environment for message handling on both Android and iOS: their application **is running** when we handle the message.
```
marklewin, line 65: Who is "they"?

it's -> its

Careful with spacing around "/" ... I've seen " / " and "/ ". Unless you can replace whatever "/" stands for with the word "or" then you really shouldn't be using it.

CONTRIBUTING.md
```diff
@@ -36,6 +36,38 @@ For the moment it would not seem appropriate for our plugin to throw instances c
...
+- **Resource consumption tradeoffs:** Firebase launches an isolate capable of only handling messages at app launch, even if they don't need it. They keep this isolate launching throughout the app, and have a queue process to queue messages. To do this, they maintain an Android Service. This allows them 10 minutes of execution time, where as on iOS the same library only has 30 seconds of execution time. Instead, we launch a new isolate on every message if the application is not yet running. Instead of creating a service and queueing work for it for each message. FCM will hold the messages until we are ready anyway (they are our queue). A new message will spawn a new engine.
```
marklewin, line 66: Again, we need to be clear about who "they" and "them" refers to.

And who is "we/they" here? Ably? "FCM will hold the messages until we are ready anyway (they are our queue)." Let's be specific about which parties we are talking about here.

CONTRIBUTING.md
```diff
@@ -36,6 +36,38 @@ For the moment it would not seem appropriate for our plugin to throw instances c
...
+  
+Because of this architectural simplicity, we do not need to use [`PluginUtilies`](https://stackoverflow.com/questions/69208164/what-does-pluginutilities-do-in-flutters-dartui-library-and-how-do-i-use/69208165#69208165), pass references of 2 methods between Dart and platform side, or save / load these methods in `SharedPreferences`. We avoid conflicts between the default `FlutterEngine` launched in an `Activity` and by ourselves, by using separate method channels (unique channel names).
```
marklewin, line 69: 1-10 should be spelled out, e.g. "two methods", not "2 methods"

"Dart and platform side" -> Dart and the platform

"We avoid conflicts between the default `FlutterEngine` launched in an `Activity` and by ourselves, by using separate method channels (unique channel names).". -> I'm not sure what the distinction between "we" and "ourselves" is in this sentence.

PushNotifications.md
```diff
@@ -80,11 +77,18 @@ To get push notifications setup in your own app, read [Setting up your own app](
 
 Devices need to be [activated](#device-activation) with Ably once. Once activated, you can use their device ID, client ID or push token (APNs device token/ FCM registration token) to push messages to them using the Ably dashboard or a [Push Admin](https://ably.com/documentation/general/push/admin) (SDKs which provide push admin functionality, such as [Ably-java](https://github.com/ably/ably-java), [Ably-js](https://github.com/ably/ably-js), etc.). However, to send push notifications through Ably channels, devices need to [subscribe to a channel for push notifications](#subscribing-to-channels-for-push-notifications). Once subscribed, messages on that channel with a [push payload](#sending-messages) will be sent to devices which are subscribed to that channel.
 
+### Ensure `WidgetsBinding` is initialized
+
+If calling any methods from the `ably_flutter` package in your `main` function, call `WidgetsFlutterBinding.ensureInitialized();`.
```
marklewin, line 82:
```suggestion
If you invoke any methods from the `ably_flutter` package in your `main` function, then ensure that you also call `WidgetsFlutterBinding.ensureInitialized();`. 
```

Should the `WidgetsBinding` in the header read `WidgetsFlutterBinding`? Perhaps the heading could be more descriptive?

PushNotifications.md
```diff
@@ -80,11 +77,18 @@ To get push notifications setup in your own app, read [Setting up your own app](
 
 Devices need to be [activated](#device-activation) with Ably once. Once activated, you can use their device ID, client ID or push token (APNs device token/ FCM registration token) to push messages to them using the Ably dashboard or a [Push Admin](https://ably.com/documentation/general/push/admin) (SDKs which provide push admin functionality, such as [Ably-java](https://github.com/ably/ably-java), [Ably-js](https://github.com/ably/ably-js), etc.). However, to send push notifications through Ably channels, devices need to [subscribe to a channel for push notifications](#subscribing-to-channels-for-push-notifications). Once subscribed, messages on that channel with a [push payload](#sending-messages) will be sent to devices which are subscribed to that channel.
 
+### Ensure `WidgetsBinding` is initialized
+
+If calling any methods from the `ably_flutter` package in your `main` function, call `WidgetsFlutterBinding.ensureInitialized();`. 
+You only need to do this if calling the functions before calling `runApp()`.
```
marklewin, line 83: Perhaps combine the two sentences for clarity, e.g.

"If you invoke any methods from the `ably_flutter` package in your `main` function before the call to `runApp()`, then ensure that you also call `WidgetsFlutterBinding.ensureInitialized();`.

PushNotifications.md
```diff
@@ -80,11 +77,18 @@ To get push notifications setup in your own app, read [Setting up your own app](
...
+This is done to ensure all platform methods will be successfully received by the native Ably plugin on the host platform.
```
marklewin, line 84:
```suggestion
This ensures that all platform methods will be successfully received by the native Ably plugin on the host platform.
```

PushNotifications.md
```diff
@@ -80,11 +77,18 @@ To get push notifications setup in your own app, read [Setting up your own app](
...
+
 ### Device activation
 
 - Create a rest or realtime client: e.g. `final realtime = ably.Realtime(options: clientOptions);`
 - Activate the device for push notifications with Ably: `ablyClient.push.activate();`. This only
   needs to be done once, and will be used across all future app launches, as long as the app is not deactivated. This method will throw an AblyException if it fails.
+- The future returned by `activate` is not guaranteed to complete quickly. For example, in the cases where there is no internet connection, `activate` will wait until internet connection. Therefore, any code waiting for the future to complete should be not important, e.g. logging code.
```
marklewin, line 91:
```suggestion
- The future returned by `activate` is not guaranteed to complete quickly. For example, if there is no internet connection, `activate` will wait until one is available. Therefore, code waiting for the future to complete should not be important, e.g. logging code.
```

marklewin: It's unclear to me whether you are saying that logging code _is_ important and therefore _not_ a good candidate for this. I assume not?

PushNotifications.md
```diff
@@ -236,17 +234,93 @@ final message = ably.Message(
 
 ### Receiving Messages
 
+To see an example of receiving messages and handling other notification related behaviour, look at [push_notification_handlers](example/lib/push_notifications/push_notification_handlers.dart) in the example app.
```
marklewin, line 237: notification related behavior -> notification-related behavior (note US spelling)

But perhaps just: "For examples of handling incoming messages and dealing with notifications, see:"

PushNotifications.md
```diff
@@ -236,17 +234,93 @@ final message = ably.Message(
...
+
 #### Notification Message / Alert Push Notification
 
-**Android**: You cannot handle notification messages as they are shown to the user without calling any methods in your application. To create notifications which launch the application to a certain page (notifications which contain deep links or app links), or notifications which contain buttons/actions, images, and inline replies, you should send a data message and create a notification when the message is received.
+**Android**: In the case where the app is in the background / terminated, you cannot configure / disable notification messages as they are automatically shown to the user by Firebase Messaging Android SDK. To create notifications which launch the application to a certain page (notifications which contain deep links or app links), or notifications which contain buttons / actions, images, and inline replies, you should send a data message and create a notification when the message is received.
```
marklewin, line 241: In the case where -> When

PushNotifications.md
```diff
@@ -236,17 +234,93 @@ final message = ably.Message(
...
+
 #### Notification Message / Alert Push Notification
 
-**Android**: You cannot handle notification messages as they are shown to the user without calling any methods in your application. To create notifications which launch the application to a certain page (notifications which contain deep links or app links), or notifications which contain buttons/actions, images, and inline replies, you should send a data message and create a notification when the message is received.
+**Android**: In the case where the app is in the background / terminated, you cannot configure / disable notification messages as they are automatically shown to the user by Firebase Messaging Android SDK. To create notifications which launch the application to a certain page (notifications which contain deep links or app links), or notifications which contain buttons / actions, images, and inline replies, you should send a data message and create a notification when the message is received. 
+
+**iOS**: In the case where the app is in the background / terminated, ably-flutter doesn't provide the functionality to configure / extend alert notifications on iOS, and these will automatically be shown to the user.
```
marklewin, line 243: In the case where -> When

Library names should be in code font throughout (`ably-flutter`)

PushNotifications.md
```diff
@@ -236,17 +234,93 @@ final message = ably.Message(
...
+
+To create richer (more content or more interactive) local notifications once a data message is received on both Android and iOS, take a look at [awesome_notifications](https://pub.dev/packages/awesome_notifications).
```
marklewin, line 245:
```suggestion
To create richer and more interactive local notifications when a data message is received on both Android and iOS, take a look at [awesome_notifications](https://pub.dev/packages/awesome_notifications).
```

PushNotifications.md
```diff
@@ -236,17 +234,93 @@ final message = ably.Message(
...
+  - Advanced usage: 
+    - To do this natively on Android instead, you could use [`notificationManager.notify`](https://developer.android.com/reference/android/app/NotificationManager#notify(int,%20android.app.Notification)). 
+    - To do this natively on iOS, you can send a background message and follow [Scheduling a Notification Locally from Your App](https://developer.apple.com/documentation/usernotifications/scheduling_a_notification_locally_from_your_app). However, on iOS, you could also [customize the appearance of an alert notification](https://developer.apple.com/documentation/usernotificationsui/customizing_the_appearance_of_notifications), by registering and implementing a [`UNNotificationContentExtension`](https://developer.apple.com/documentation/usernotificationsui/unnotificationcontentextension).
```
marklewin, line 248:
```suggestion
    - To do this natively on iOS, you can send a background message and follow [Scheduling a Notification Locally from Your App](https://developer.apple.com/documentation/usernotifications/scheduling_a_notification_locally_from_your_app). On iOS, you can also [customize the appearance of an alert notification](https://developer.apple.com/documentation/usernotificationsui/customizing_the_appearance_of_notifications), by registering and implementing a [`UNNotificationContentExtension`](https://developer.apple.com/documentation/usernotificationsui/unnotificationcontentextension).
```

PushNotifications.md
```diff
@@ -236,17 +234,93 @@ final message = ably.Message(
...
+
+#### Foreground notifications
 
-**iOS**: You can send a background message and follow [Scheduling a Notification Locally from Your App](https://developer.apple.com/documentation/usernotifications/scheduling_a_notification_locally_from_your_app). However, on iOS, you could also [customize the appearance of an alert notification](https://developer.apple.com/documentation/usernotificationsui/customizing_the_appearance_of_notifications), by registering and implementing a [`UNNotificationContentExtension`](https://developer.apple.com/documentation/usernotificationsui/unnotificationcontentextension).
+On both Android and iOS, ably-flutter allows you to configure notification sent directly from Ably to be shown when the app is in foreground. To configure if a notification is shown in the foreground:
```
marklewin, line 252:
```suggestion
On both Android and iOS, `ably-flutter` enables you to show Ably notifications when the app is in the foreground. To configure this foreground display:
```

PushNotifications.md
```diff
@@ -236,17 +234,93 @@ final message = ably.Message(
...
+  print('Opting to show the notification when the app is in the foreground.');
+  return true;
+});
+```
+
+Internally: On Android, the SDK does this by generating the same notification Firebase creates when the app is in the background. On iOS, the SDK does this by implementing [`userNotificationCenter(_:willPresent:withCompletionHandler:)`](https://developer.apple.com/documentation/usernotifications/unusernotificationcenterdelegate/1649518-usernotificationcenter).
+
+#### Notification taps
+
+- To be informed when a notification tap launches the app:
```
marklewin, line 266:
```suggestion
- To be informed when the user taps on a notification to launch the app:
```

PushNotifications.md
```diff
@@ -236,17 +234,93 @@ final message = ably.Message(
...
+ably.Push.notificationEvents.notificationTapLaunchedAppFromTerminated
+    .then((remoteMessage) {
+  if (remoteMessage != null) {
+    print('The app was launched by the user by tapping the notification');
+    print(remoteMessage.data);
+  }
+});
+```
+  
+- To be informed when a notification is tapped by the user whilst the app is in the foreground:
```
marklewin, line 278:
```suggestion
- To be informed when the user taps on a notification while the app is in the foreground:
```

PushNotifications.md
```diff
@@ -236,17 +234,93 @@ final message = ably.Message(
...
+ably.Push.notificationEvents.onNotificationTap.listen((remoteMessage) {
+  print('Notification was tapped: $remoteMessage');
+});
+```
 
 #### Data Message / Background Notification
 
-ably-flutter currently does not pass the messages to the Flutter application/ Dart-side, so users will need to listen to messages in each platform. See [Implement Push Notifications listener](https://github.com/ably/ably-flutter/issues/141) for more information.
+- On iOS, if a notification is also present, it will be shown before `onMessage` or the callback passed to `ably.Push.notificationEvents.setOnBackgroundMessage` is called.
+- On Android, if a notification is also present, it will be shown after these methods finish (or it's Future completes).
```
marklewin, line 289: it's -> its

PushNotifications.md
```diff
@@ -236,17 +234,93 @@ final message = ably.Message(
...
+  print('RemoteMessage.Data: ${message.data}');
+}
+ably.Push.notificationEvents.setOnBackgroundMessage(_backgroundMessageHandler);
+```
+  - This method can be synchronous or `async`.
+
+##### Advanced: native message handling
 
-**Android**: This means implementing [`FirebaseMessageService`](https://firebase.google.com/docs/cloud-messaging/android/receive) and overriding `onMessageReceived` method. You must also declare the Service in your `AndroidManifest.xml`. Once you receive your message, you could create a richer notification, by following [Create a Notification](https://developer.android.com/training/notify-user/build-notification).
+Users can listen to messages in each platform using the native message listeners instead of listening to messages on the Dart side. This is not recommended unless you want to **avoid** using other plugins, such as [awesome_notifications](https://pub.dev/packages/awesome_notifications) and [flutter_local_notifications](https://pub.dev/packages/flutter_local_notifications).
```
marklewin, line 314:
```suggestion
Users can listen to messages in each platform using the native message listeners instead of Dart listeners. This is not recommended unless you want to **avoid** using other plugins, such as [awesome_notifications](https://pub.dev/packages/awesome_notifications) and [flutter_local_notifications](https://pub.dev/packages/flutter_local_notifications).
```

PushNotifications.md
```diff
@@ -236,17 +234,93 @@ final message = ably.Message(
...
+
+**Android**: This means implementing [`FirebaseMessageService`](https://firebase.google.com/docs/cloud-messaging/android/receive) and overriding `onMessageReceived` method. You must also declare the `Service` in your `AndroidManifest.xml`. Once you receive your message, you could create a richer notification, by following [Create a Notification](https://developer.android.com/training/notify-user/build-notification). As declaring this service would override the service used internal by Ably Flutter, be sure to provide Ably Flutter with your registration token. Implement the following `onNewToken` method in `FirebaseMessagingService`:
```
marklewin, line 316:
```suggestion
**Android**: This requires you to implement [`FirebaseMessageService`](https://firebase.google.com/docs/cloud-messaging/android/receive) and override the `onMessageReceived` method. You must also declare the `Service` in your `AndroidManifest.xml`. Once you receive your message, you could [create a richer notification](https://developer.android.com/training/notify-user/build-notification). As declaring this service would override the service used internally by Ably Flutter, be sure to provide Ably Flutter with your registration token. Implement the following `onNewToken` method in `FirebaseMessagingService`:
```

PushNotifications.md
```diff
@@ -347,6 +421,12 @@ This means your registration token is invalid. Ably is may not have your device'
 ActivationContext.getActivationContext(this).onNewRegistrationToken(RegistrationToken.Type.FCM, registrationToken);
 ```
 
+### My notifications do not make a sound when they are shown on the device
+
+Android: Ensure you have set up a Notification channel (which displays as notification "categories" to Android users within your application settings). The example app shows how to do this by using [flutter_local_notifications](https://pub.dev/packages/flutter_local_notifications) to configure `AndroidNotificationChannel`.
```
marklewin, line 426: Could bold Android and iOS here, as you have done elsewhere.

marklewin: I don't think Notification channel needs to be capitalized here.

marklewin (changes requested):

android/src/main/java/io/ably/flutter/plugin/AblyFlutterPlugin.java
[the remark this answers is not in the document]
ben-xD: Cool, here it is: https://github.com/ably/ably-flutter/issues/183

CONTRIBUTING.md
ben-xD: 3765af68a76229b6c66afe1e975c815d598df9a4

CONTRIBUTING.md
ben-xD: 642a0a5f5cada12e98ca0a52c809db26ad200bf5

ben-xD: 642a0a5f5cada12e98ca0a52c809db26ad200bf5

CONTRIBUTING.md
ben-xD: 11fd07566c967a2d64102a7d4dce52ad3ec587e5

CONTRIBUTING.md
ben-xD: 51b737b98bad1ab4b394566a49b487f10fbb5a21

CONTRIBUTING.md
ben-xD: Not sure what you mean. This is a contributing.md file.

PushNotifications.md
ben-xD: 407807f53a9d46b692c2cca30492e33d6f441bdd

CONTRIBUTING.md
ben-xD: 598aca720af6c4956619fb5b817d16b2af4bef40

CONTRIBUTING.md
ben-xD: a6e241b12db9adc72b53311cc4073b95d3bcc298

PushNotifications.md
ben-xD: e6f9a88cbd0b50150943aa94cf10affc98595e1f

PushNotifications.md
ben-xD: This part of the docs was outdated, I've had to update it: a4bd017f4f70b3f013bb3999ca07793c84a58968

PushNotifications.md
ben-xD: Clarified in fd50412e271b63b00b16edb9c852c0033606678e

PushNotifications.md
ben-xD: e6f9a88cbd0b50150943aa94cf10affc98595e1f

PushNotifications.md
ben-xD: I initially chose the term "richer", but I don't think it's very good. I've removed it from the docs and clarified this sentence: 3e229b82b03847083da389b6ac73e779202b4375

Issue #168, opened 2021-09-13
Replace `package:pedantic` with `package:flutter_lints`


ben-xD:
https://pub.dev/packages/pedantic has been deprecated. There is a new package for flutter linting (as opposed to just dart linting), and an existing package recommended for **Flutter apps, packages, and plugins**
- https://flutter.dev/docs/release/breaking-changes/flutter-lints-package

We should switch to this. New apps flutter apps already use `flutter-lints` by default as of Flutter 2.3.0. We are at 2.5.0 now. Instructions on installing this can be found https://pub.dev/packages/flutter_lints#usage

Issue #85, opened 2021-02-15
Lint rule to enforce strict raw types


tiholic:
https://github.com/ably/ably-flutter/blob/fd15cba20d3ec4f9768df14b10b1ad88e4a4c174/analysis_options.yaml#L18

Issue #169, opened 2021-09-15
Remove `prefer_relative_imports` and use absolute package imports


ben-xD:
In `analysis_options.yaml`, we have 
```
    # personal preference this conflicts with prefer_relative_imports
    # - always_use_package_imports
```
and 
```
prefer_relative_imports # personal preference
```

-  This reduces developer experience, because the IDE plugin (Flutter-Intellij, which is used in Android Studio too) auto-imports using absolute package paths, not relative paths. This rule goes **against the default**, and reduces efficiency with development because they have to explicitly **fix** the imports to be relative again: annoying/ waste of time. Therefore, it's not really personal preference.
- Current problem: Because of the plugin, these imports are **already** inconsistent. Some imports are relative, some are absolute, because again, the IDE automatically creates absolute ones.

TLDR: In this issue, we should remove the above code snippets, opting for `always_use_package_imports` instead. Then, we would update the paths. This would allow new developers to be productive and avoid linter warnings when auto importing anything.

PR #170, opened 2021-09-15
Replace deprecated `package:pedantic` with `package:flutter_lints` and update `analysis_options.yaml`

ben-xD:
I removed deprecated [pedantic](https://pub.dev/packages/pedantic) package, added new [flutter_lints](https://pub.dev/packages/flutter_lints) package, and simplified `analysis_options.yaml` to be more maintainable and readable.

Closes https://github.com/ably/ably-flutter/issues/168
Also, closes https://github.com/ably/ably-flutter/issues/85

--- 

After doing the dependency changes, I checked which rules are already being added "included" by using the new lint package, `include: package:flutter_lints/flutter.yaml`, and removed the lint rules which are **already enabled by default**. This reduces the content of `analysis_options.yaml` **without** changing any dart analysis warnings. This a combination of 3 files, flutter, recommended and core: 
- https://github.com/flutter/packages/blob/master/packages/flutter_lints/lib/flutter.yaml
- https://github.com/dart-lang/lints/blob/main/lib/recommended.yaml
- https://github.com/dart-lang/lints/blob/main/lib/core.yaml

This existing shopping-list/ kitchen-sink style of `analysis_options.dart` is common when a lint package is not used/ `include`'d. However, when we do use a convenient package like flutter-lints, it doesn't make sense to override them to be the same values.

From this, I realised that the annoying issue I've been facing over the past 2 months leading to inconsistent styling in the code is down to this a lint rule that was turned on: https://github.com/ably/ably-flutter/issues/169, so I created that issue. To avoid noise in this PR, that will be a separate PR in the future.

---

I also had to format `test/rest/channels_test.dart`. It seems like the Dart analyzer was not linting/ checking formatting in that file previously.


tiholic (approved):
I hope many rules that are removed from `analysis_options.yaml` because they are not supported?

QuintinWillison (approved):
It would be impractical for me to comb this line by line. I trust that you've made the right choices. Thanks for updating this. 🚀

Issue #171, opened 2021-09-16
Remove timeouts for platform method calls


ben-xD:
We have timeouts for futures when making method calls to the platform side:

https://github.com/ably/ably-flutter/blob/main/lib/src/platform/src/platform_object.dart#L40-L48

and

https://github.com/ably/ably-flutter/blob/main/lib/src/platform/src/platform.dart#L32-L38

This is timeout is not helpful, and in fact hurts developer experience (including users). They are not able to debug/ place use breakpoints properly, because the platform method times out. Additionally, slower devices platform method calls will naturally be slower. This has caused poor developer experience for me, and users in the past, for example: https://github.com/ably/ably-flutter/issues/72

- If a method channel or `WidgetsBinding` is not initialised early enough, you will get `No implementation found for method MethodName`, instantaneously.
- If the Dart VM is not initialized, then the entire app would not be running. This code would not be running either. 
- If something catastrophic was happening, then that is the problem. 

There is no reason to support having this timeout. I would call it a bug, because users cannot debug their applications properly or run this app on a slow device (in theory).

In https://github.com/ably/ably-flutter/issues/72#issuecomment-741922513, Quintin asked
> What do other libraries do to solve this?

The answer is that other libraries do not have timeouts at all, there is no reason to arbitrarily kill a platform method after any amount of seconds.

I have discussed this a bit with @tiholic, thank you :).

ben-xD:
https://github.com/ably/ably-flutter/blob/main/lib/src/platform/src/platform.dart#L32-L38

PR #172, opened 2021-09-17
Remove timeouts for platform methods

ben-xD:
I've had this code for development for the past 2 weeks, to avoid debugging/ development pain. It doesn't seem to be a problem, and I cannot see a reason to keep it, as per https://github.com/ably/ably-flutter/issues/171

Closes https://github.com/ably/ably-flutter/issues/171


QuintinWillison (approved):
I notice checks failing, but otherwise looks like a positive step forward. Thanks!

PR #190, opened 2021-10-29
Fix RealtimeChannelOptions to avoid forcing cipher argument

### LINKED ISSUE #182: Can not set RealTimeChannelOptions params
Because cipher is not nullable I'm unable to pass `{'rewind':'1'}` to `RealtimeChannelOptions`  without using cipher.

```javascript
RealtimeChannelOptions RealtimeChannelOptions(
  Object cipher, {
  Map<String, String>? params,
  List<ChannelMode>? modes,
})
```

Otherwise I could not find a reference within the flutter example how to use one...

Upon further investigation, `CipherParams` seems to be abstract. Crypto interface doesn't seem to be implemented so I couldn't use `getDefaultParams` using my server-side generated key. So I tried to manually pass `cipherParams` as a Map but that threw a methodChannel error as well.

Basically locked out of symmetric encryption as well as channel options...

ben-xD:
Fixes #182


ikbalkaya (approved):
LGTM

pubspec.lock
```diff
@@ -222,7 +222,7 @@ packages:
     source: hosted
     version: "1.8.0"
   pedantic:
-    dependency: "direct main"
+    dependency: transitive
```
QuintinWillison, line 225: How does this change relate to the issue being fixed?

lib/src/realtime/src/realtime_channel_options.dart
```diff
@@ -12,6 +14,6 @@ class RealtimeChannelOptions extends RestChannelOptions {
   final List<ChannelMode>? modes;
 
   /// create channel options with a cipher, params and modes
-  RealtimeChannelOptions(Object cipher, {this.params, this.modes})
-      : super(cipher);
+  RealtimeChannelOptions({this.params, this.modes, CipherParams? cipher})
```
QuintinWillison, line 17: Does this represent a breaking API change?

QuintinWillison (commented):
A couple of questions before I can feel able to approve this.

lib/src/realtime/src/realtime_channel_options.dart
ben-xD: Yes, it's a breaking change. 

- It moves the interface from an API which was bad (forcing the user to set a cipher if they wanted to set params or modes) to one where all these options are optional. 
- It was also quite unclear, what types can Cipher be?. I also change it from an `Object` which could be anything, to be more specific: CipherParams. 

**In practice, I would not worry about this breaking change.** This constructor isn't widely used right now by customers. On Android, modes are ignored, and params was used as CipherParams. In iOS, a crash occurs. It is a completely untested / unvalidated piece of functionality in the app. 

These bugs are fixed in https://github.com/ably/ably-flutter/pull/191/files
```
readValueFromJson(jsonMap, PlatformConstants.TxRealtimeChannelOptions.params, v -> o.cipherParams = (Map<String, String>) v);
```

Notice the params passed in from the dart side is being set as cipherParams... 🙈

pubspec.lock
ben-xD: It doesn't, but unfortunately some of these files have a life of their own (I don't truly understand what causes these to change). 

I manually changed it back in

PR #191, opened 2021-10-29
Fix 2 Channel options bugs

ben-xD:
There are 2 bug fixes in here:

**1 in Ably-java**: It was wrongly assumed that Ably-java does not support channel modes. It does, and here I add support for it.

**1 in Ably-cocoa:** The wrong Ably-cocoa method was used, and in this case, causes a **runtime crash** whenever a user **sets options on a realtime channel**. This has happened because code was copy pasted on not ran or validated.

There is one enhancement regarding the cipher passed to channel options. You can see this in `createChannelOptions(@Nullable Object cipher)`. Previously, in Ably-java, we assume the cipher to be a String, causing a crash if it wasn't. Now, I add support for byte array, as well as logging information out if the cipher is not a string or a byte array.


android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
```diff
@@ -355,26 +358,73 @@ private ChannelOptions decodeRestChannelOptions(Map<String, Object> jsonMap) {
...
+
+  private ChannelMode[] createChannelModesArray(ArrayList<String> modesString) {
+    ChannelMode[] modes = new ChannelMode[modesString.size()];
+    for (int i = 0; i < modesString.size(); i++) {
+      modes[i] = decodeChannelOptionsMode(modesString.get(i));
+    }
+    return modes;
+  }
+
+  private ChannelMode decodeChannelOptionsMode(String string) {
```
ikbalkaya, line 415: Can you change the name of the param from string to something else please? Can be something like `modeString` I also think you could benefit from creating a static map which maps the mode strings to modes and you could just read it from the map.

ikbalkaya (changes requested):
I made a  slight readability improvement request

android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
ikbalkaya: For static map, if you haven't done it before you can use `static` bloc in the class https://stackoverflow.com/a/507658

ben-xD: I've renamed it in 746100d9eb496bc30c7ec43eb12e9bfe57a450ae

However, what are the benefits of creating the static map? To me, the existing switch statement is simpler and shorter.

QuintinWillison (approved):

android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
ikbalkaya: As we spoke on the phone, it's a map of some constant values and applies nicely to this situation. It is common to map strings to some typed values, especially when translating from an external source using a defined protocol. The end goal is to have a map which you can use around app and that is a reference for that protocol. Anyway, in this case this function is also acceptable but I'd encourage you to refactor this for later.

ikbalkaya (approved):

Issue #104, opened 2021-04-29
Symmetric Encryption


QuintinWillison:
The absence of this feature is a known limitation of this plugin.

See:
- Docs: [Ably Realtime: Encryption](https://ably.com/documentation/realtime/encryption)
- [RSL5](https://docs.ably.io/client-lib-development-guide/features/#RSL5): Message payload encryption
- [RSE1](https://docs.ably.io/client-lib-development-guide/features/#RSE1): `Crypto` function `getDefaultParams`
- [RSE2](https://docs.ably.io/client-lib-development-guide/features/#RSE2): `Crypto` function `generateRandomKey`
- [TM3](https://docs.ably.io/client-lib-development-guide/features/#TM3): `Message` methods `fromEncoded` and `fromEncodedArray`
- [TP4](https://docs.ably.io/client-lib-development-guide/features/#TP4): `PresenceMessage` methods `fromEncoded` and `fromEncodedArray`
- [TB2b](https://docs.ably.io/client-lib-development-guide/features/#TB2b): `ChannelOptions` attribute `cipher`
- [TB3](https://docs.ably.io/client-lib-development-guide/features/#TB3): Optional `ChannelOptions` constructor accepting a cipher key
- [TZ1](https://docs.ably.io/client-lib-development-guide/features/#TZ1) and [TZ2](https://docs.ably.io/client-lib-development-guide/features/#TZ2): `CipherParams`

Issue #180, opened 2021-09-21
Simplify and refactor `AblyMethodCallHandler.java` and `AblyMessageCodec.java`


ben-xD:
- This file is 770 lines long. We should group method call handlers into their separate files.
- We could replace the map with a switch statement, like other Flutter packages. It will avoid having complexity like: `final BiConsumer<MethodCall, MethodChannel.Result> handler = _map.get(call.method);`. See https://github.com/ably/ably-flutter/pull/166#discussion_r712299430 for a small discussion. Or alternatively, just rename the `_map`.

ben-xD:
We should also do this in `AblyMessageCodec.java`

PR #192, opened 2021-10-29
Symmetric encryption

ben-xD:
This PR closes https://github.com/ably/ably-flutter/issues/104

It adds symmetric encryption, please read the new [documentation added to the README](https://github.com/ably/ably-flutter/tree/feature/symmetric-encryption#symmetric-encryption).

This PR will not yet work on iOS because it does not yet use the bug fix from Ably Cocoa: https://github.com/ably/ably-cocoa/pull/1208 (I've tested this locally). A **new version of Ably Cocoa needs to be released**, and then Ably Flutter upgraded to the latest version (Ably Cocoa has a couple of small breaking changes fixable in less than 1 minute). I have confirmed that it works in Ably Flutter with this bug fix.

This PR does not yet include decryption of messages which are passed to `PresenceMessage.fromEncoded` and `Message.fromEncoded`, as per [TP4](https://docs.ably.com/client-lib-development-guide/features/#TP4) and [TM3](https://docs.ably.com/client-lib-development-guide/features/#TM3). This will be done in a separate PR https://github.com/ably/ably-flutter/issues/207, to keep this PR smaller.


ikbalkaya:
> This PR will not yet work on iOS because it does not yet use the bug fix from Ably Cocoa: [ably/ably-cocoa#1208](https://github.com/ably/ably-cocoa/pull/1208) (I've tested this locally). A **new version of Ably Cocoa needs to be released**, and then Ably Flutter upgraded to the latest version (Ably Cocoa has a couple of small breaking changes fixable in less than 1 minute).
> 
> This PR does not yet include decryption of messages which are passed to `PresenceMessage.fromEncoded` and `Message.fromEncoded`, as per [TP4](https://docs.ably.com/client-lib-development-guide/features/#TP4) and [TM3](https://docs.ably.com/client-lib-development-guide/features/#TM3). This will be done in a separate PR, to keep this PR smaller.

@ben-xD It would be useful to add what this PR is doing as well :)

ben-xD:
> > This PR will not yet work on iOS because it does not yet use the bug fix from Ably Cocoa: [ably/ably-cocoa#1208](https://github.com/ably/ably-cocoa/pull/1208) (I've tested this locally). A **new version of Ably Cocoa needs to be released**, and then Ably Flutter upgraded to the latest version (Ably Cocoa has a couple of small breaking changes fixable in less than 1 minute).
> > This PR does not yet include decryption of messages which are passed to `PresenceMessage.fromEncoded` and `Message.fromEncoded`, as per [TP4](https://docs.ably.com/client-lib-development-guide/features/#TP4) and [TM3](https://docs.ably.com/client-lib-development-guide/features/#TM3). This will be done in a separate PR, to keep this PR smaller.
> 
> @ben-xD It would be useful to add what this PR is doing as well :)

Thanks for commenting, I just realised I did not write documentation for this yet. I will write this now, and then it should make sense.

example/lib/encrypted_messaging_service.dart
```diff
@@ -0,0 +1,90 @@
...
+  ably.RealtimeChannel? _realtimeChannel;
+  ably.RestChannel? _restChannel;
+
+  final BehaviorSubject<List<ably.Message>> messageHistoryBehaviorSubject =
+      BehaviorSubject<List<ably.Message>>.seeded([]);
+
+  ValueStream<List<ably.Message>> get messageHistoryStream =>
+      messageHistoryBehaviorSubject.stream;
+
+  static const examplePassword = 'password-to-encrypt-and-decrypt-text';
```
ikbalkaya, line 22: Just wondering, if this a test class?

ikbalkaya (changes requested):
Light review warning -> Made some comments that will likely end up in very minor or no changes

android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java
```diff
@@ -102,6 +107,14 @@ public AblyMethodCallHandler(final MethodChannel channel, final HotRestartCallba
     // paginated results
     _map.put(PlatformConstants.PlatformMethod.nextPage, this::getNextPage);
     _map.put(PlatformConstants.PlatformMethod.firstPage, this::getFirstPage);
+
+    // Encryption
```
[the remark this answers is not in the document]
ben-xD: It's just a way to keep this file tidy, following the existing contents of the file as convention. https://github.com/ably/ably-flutter/blob/6171f2fbd23f85274b3ec865ba624c7f57b00f01/android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java#L65-L104

android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
```diff
@@ -356,25 +369,31 @@ private TokenDetails decodeTokenDetails(Map<String, Object> jsonMap) {
...
   private ChannelOptions decodeRestChannelOptions(Map<String, Object> jsonMap) {
     if (jsonMap == null) return null;
-    final Object cipher = jsonMap.get(PlatformConstants.TxRealtimeChannelOptions.cipher);
-    try {
-      return createChannelOptions(cipher);
-    } catch (AblyException e) {
-      System.out.println("Exception while decoding RestChannelOptions: " + e);
-      return null;
+    ChannelOptions options = new ChannelOptions();
+    options.cipherParams = decodeCipherParams((Map<String, Object>) jsonMap.get(PlatformConstants.TxRestChannelOptions.cipherParams));
```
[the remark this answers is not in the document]
ben-xD: Very safe because I wrote the code on the other side. 
https://github.com/ably/ably-flutter/blob/0898c07cf4487b8b8614260ad28daa11e64f336d/lib/src/platform/src/codec.dart#L379-L390

This is the way Flutter plugins pass information back and forth from Dart and Java side. You can see this pattern all over Ably Flutter, but also every package plugin (A flutter plugin which contains iOS or Android code).

For example, look at this implementation in the Flutterfire firebase_messaging library: https://github.com/FirebaseExtended/flutterfire/blob/a91fa459add243b37a7da9a67fbe94f2116bb3ca/packages/cloud_firestore/cloud_firestore/android/src/main/java/io/flutter/plugins/firebase/firestore/FlutterFirebaseFirestorePlugin.java#L271-L289

Although we cannot **statically** guarantee the type is correct before casting, **checking** the type at runtime and throwing an error or logging if it is not as expected adds verbosity, complexity with no additional benefits IMHO.

android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
```diff
@@ -356,25 +369,31 @@ private TokenDetails decodeTokenDetails(Map<String, Object> jsonMap) {
...
-    final Object cipher = jsonMap.get(PlatformConstants.TxRealtimeChannelOptions.cipher);
-
-    ChannelOptions options;
-    try {
-      options = createChannelOptions(cipher);
-    } catch (AblyException e) {
-      System.out.println("Exception while decoding RealtimeChannelOptions: " + e);
-      return null;
+    ChannelOptions options = new ChannelOptions();
+    options.cipherParams = decodeCipherParams((Map<String, Object>) jsonMap.get(PlatformConstants.TxRealtimeChannelOptions.cipherParams));
```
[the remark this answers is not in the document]
ben-xD: This is the same cast as https://github.com/ably/ably-flutter/pull/192#discussion_r744586883

android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
```diff
@@ -189,6 +197,11 @@ private Byte getType(Object value) {
       return PlatformConstants.CodecTypes.pushChannelSubscription;
     } else if (value instanceof RemoteMessage) {
       return PlatformConstants.CodecTypes.remoteMessage;
+    } else if (value instanceof Crypto.CipherParams) {
+      return PlatformConstants.CodecTypes.cipherParams;
+    } else if (value instanceof ChannelOptions) {
```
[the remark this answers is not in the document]
ben-xD: Where did you find `RealtimeChannelOptions` in Ably Java? I could not find such symbol, and concluded `ChannelOptions` is shared between Realtime and Rest channels.

However, because you mentioned it, I had a deeper read into Ably Java and found a different issue: https://github.com/ably/ably-java/issues/730

android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
```diff
@@ -82,9 +86,11 @@ T decode(Map<String, Object> jsonMap) {
 
   private Map<Byte, CodecPair> codecMap;
   private static final Gson gson = new Gson();
+  private final CipherParamsStorage cipherParamsStorage;
 
-  public AblyMessageCodec() {
+  public AblyMessageCodec(CipherParamsStorage cipherParamsStorage) {
     final AblyMessageCodec self = this;
+    this.cipherParamsStorage = cipherParamsStorage;
     codecMap = new HashMap<Byte, CodecPair>() {
```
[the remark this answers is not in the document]
ben-xD: This can also be done in `AblyMethodCallHandler.java`, doing this in this PR would add a lot of unrelated changes. Having said that, I don't want to create an issue for this yet because I am not sure of the solution: It could just be a switch statement inside `onMethodCall`, instead of using this map at all: 

https://github.com/ably/ably-flutter/blob/0898c07cf4487b8b8614260ad28daa11e64f336d/android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java#L153-L165

There are more places where things can go wrong in the current implementation, and reducing code in this hot path would be useful.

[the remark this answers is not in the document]
ben-xD: Oh, please see https://github.com/ably/ably-flutter/issues/180, I made one in September 😁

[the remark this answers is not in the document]
ikbalkaya: I made this comment only because the constructor code was massive and extracting the building to a method should not make any change to the code. It wasn't a comment on your implementation

android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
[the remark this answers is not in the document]
ikbalkaya: I have not found `RealtimeChannelOptions` anywhere. As I'm doing light review only, I'm skimming through your code. Since you are concluding that the result should translate into `realtimeChannelOptions` but only evaluating a base class might be risky. So in this case, if you are sure thhat this will only translate to realtimeChannelOptions and not `restChannelOptions`(I don't know if this exist as well) it's fine.

android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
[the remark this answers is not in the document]
ben-xD: I agree it is big. However, extracting it would be quite a large change in the diff and would like to keep it for a separate PR. If I extracted it, it would be less clear that I added:
```
put(PlatformConstants.CodecTypes.cipherParams,
            new CodecPair<>(self::encodeCipherParams, self::decodeCipherParams));
```

android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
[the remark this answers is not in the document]
ikbalkaya: Thanks for clarification 👍

android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java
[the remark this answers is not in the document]
ikbalkaya: Ah ok understood. that looked like a regular comment to me. For the future reference, you could specify code regions using
//region Description
//endregion

This will help IntelliJ based IDEs pick that and filter it :)

android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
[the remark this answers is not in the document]
ikbalkaya: Ok then, I'll leave this with you 👍

README.md
```diff
@@ -518,6 +517,33 @@ channel
 );
 ```
 
+### Symmetric Encryption
+
+
+Ably client libraries support built-in symmetric encryption of message content, making it easier to build apps that encrypt content fully end-to-end. Whilst TLS is enabled by default and ensures that data is securely sent to and received from Ably, messages are not encrypted within the Ably system. Using the encryption feature of our client libraries ensures that message payloads are opaque, can never be decrypted by Ably, and can only be decrypted by other clients that share your secret key.
+
+- Get or create a key, either by
```
QuintinWillison, line 525: This set of outer bullets looks like it's a bunch of steps to be executed in sequence, so I would make it an ordered, numbered list.

README.md
```diff
@@ -518,6 +517,33 @@ channel
...
+    - creating a random one using: `ably.Crypto.generateRandomKey();`. You must to distribute this specific key to other clients, since calling `generateRandomKey` on another device will generate a different key.
+- Create a `CipherParams` instance by passing a key to `final cipherParams = await ably.Crypto.getDefaultParams(key: key);`. The key can be a Base64-encoded `String`, or a `Uint8List`.
+- Create a `RealtimeChannelOptions` or `RestChannelOptions` from this key: e.g. `final channelOptions = ably.RealtimeChannelOptions(cipher: cipherParams);`
+  - Alternatively, if you are only setting CipherParams on ChannelOptions, you could skip creating the `CipherParams` instance: `ably.RestChannelOptions.withCipherKey(cipherKey)` or `ably.RealtimeChannelOptions.withCipherKey(cipherKey)`.
+- Set this options on your channel: `realtimeClient.channels.get(channelName).setOptions(channelOptions);`
+- Use your channel as normal, such as by publishing messages or subscribing for messages.
+
+Overall, it would like this:
+```dart
+final key = ...; // Get a key from your server, create a key from password or create a random key.
```
QuintinWillison, line 538: Comment's too long for GitHub's default rendering. Perhaps:

```suggestion
final key = ...; // from your server, from password or create random
```

android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
```diff
@@ -356,25 +369,31 @@ private TokenDetails decodeTokenDetails(Map<String, Object> jsonMap) {
...
-      return null;
+    ChannelOptions options = new ChannelOptions();
+    options.cipherParams = decodeCipherParams((Map<String, Object>) jsonMap.get(PlatformConstants.TxRestChannelOptions.cipherParams));
+    if (options.cipherParams != null) {
+      options.encrypted = true;
     }
+    return options;
+  }
+
+  /// RealtimeChannelOptions (the inheriting class) is used for both RealtimeChannelOptions and
```
QuintinWillison, line 380: I'm not sure why this line starts with `///` and the next with `//`.

android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
```diff
@@ -356,25 +369,31 @@ private TokenDetails decodeTokenDetails(Map<String, Object> jsonMap) {
...
+  // RestChannelOptions, since Ably Java has only one type: ChannelOptions
+  private Map<String, Object> encodeChannelOptions(final ChannelOptions c) {
+    if (c == null) return null;
```
QuintinWillison, line 383: I'm seen this pattern before in this codebase, and may have commented on it then, but not sure if that was similar context. How come this contract can be so loose? Surely `null` here _should_ throw [NPE](https://docs.oracle.com/javase/7/docs/api/java/lang/NullPointerException.html) or otherwise fail hard?

android/src/main/java/io/ably/flutter/plugin/util/CipherParamsStorage.java
```diff
@@ -0,0 +1,39 @@
+package io.ably.flutter.plugin.util;
+
+import java.util.HashMap;
+
+import io.ably.lib.util.Crypto;
+
+/**
+ * Stores and gets CipherParams by an Integer handle. On Android, Ably-Flutter will use this handle
```
QuintinWillison, line 8:
```suggestion
 * Stores and gets CipherParams by an Integer handle. On Android, ably-flutter will use this handle
```

android/src/main/java/io/ably/flutter/plugin/util/CipherParamsStorage.java
```diff
@@ -0,0 +1,39 @@
...
+ * to represent CipherParams on the Dart-side. This is done because CipherParams constructor is
```
QuintinWillison, line 9:
```suggestion
 * to represent CipherParams on the Dart-side. This is done because the CipherParams constructor is
```

android/src/main/java/io/ably/flutter/plugin/util/CipherParamsStorage.java
```diff
@@ -0,0 +1,39 @@
...
+ * private, and only Ably-Java can create them. When a method is calld with CipherParams as an
```
QuintinWillison, line 10:
```suggestion
 * private, and only ably-java can create them. When a method is called with CipherParams as an
```

android/src/main/java/io/ably/flutter/plugin/util/CipherParamsStorage.java
```diff
@@ -0,0 +1,39 @@
...
+ * argument, this Integer handle will be turned back into the CipherParams using
+ * [CipherParamsStorage.from]. This only works when both methods are called on the same instance
+ * of CipherParamsStorage, and not between app restarts.
+ */
+public class CipherParamsStorage {
+  private Integer _nextHandle = 1;
+  HashMap<Integer, Crypto.CipherParams> cipherParamsByHandle = new HashMap();
+
+  /**
+   * Store
```
QuintinWillison, line 20: Is this commentary a WIP?

android/src/main/java/io/ably/flutter/plugin/util/CipherParamsStorage.java
```diff
@@ -0,0 +1,39 @@
...
+   * @param params Instance of cipherParams to be stored at runtime.
+   * @return a handle representing the CipherParams which lasts as long as CipherParamsStorage is
+   * instantiated. It doesn't not persist across app restarts.
```
QuintinWillison, line 23: Double negative "doesn't not". 😕

android/src/main/java/io/ably/flutter/plugin/util/CipherParamsStorage.java
```diff
@@ -0,0 +1,39 @@
...
+   */
+  public int getHandle(Crypto.CipherParams params) {
+    cipherParamsByHandle.put(_nextHandle, params);
+    return _nextHandle++;
+  }
+
+  /**
+   * Gets the [CipherParams] instance from the Integer handle. Use this to get a CipherParams
+   * when the Dart side provides only an Integer representing the CipherParams.
```
QuintinWillison, line 32: You used "Dart-side" in an earlier commentary, but it's not hyphenated here.

lib/src/crypto/src/cipher_params.dart
```diff
@@ -0,0 +1,14 @@
...
+
+/// params to configure encryption for a channel
+///
+/// Pass this as the cipher constructor argument of  [RestChannelOptions] or
+/// [RealtimeChannelOptions]. Do not construct [CipherParams] yourself, as it
+/// internally refers to an instance of [CipherParams] on the platform side
+/// (Android / iOS).
+///
+/// This class does not expose algorithm, keyLength or mode because these fields
+/// are private in Ably Java's Crypto.CipherParams.
```
QuintinWillison, line 11:
```suggestion
/// are private in ably-java's Crypto.CipherParams.
```

lib/src/crypto/src/cipher_params_native.dart
```diff
@@ -0,0 +1,60 @@
+import 'dart:io';
+import 'dart:typed_data';
+
+import '../../error/src/ably_exception.dart';
+import 'cipher_params.dart';
+
+/// The internal (non-public) representation of CipherParams.
+///
+/// On Android, most fields are either private or package-private, and cannot
+/// be accessed. Therefore, Ably Flutter stores the instance of CipherParams
```
QuintinWillison, line 10:
```suggestion
/// be accessed. Therefore, ably-flutter stores the instance of CipherParams
```

lib/src/crypto/src/cipher_params_native.dart
```diff
@@ -0,0 +1,60 @@
...
+/// and only gives a references to the dart side. When another method want
```
QuintinWillison, line 11:
```suggestion
/// and only gives a references to the Dart-side. When another method wants
```

lib/src/crypto/src/cipher_params_native.dart
```diff
@@ -0,0 +1,60 @@
...
+/// to use this instance, we use the reference to get back the instance.
+///
+/// On iOS, ARTCipherParams fields are public. Therefore, Ably Flutter can
```
QuintinWillison, line 14:
```suggestion
/// On iOS, ARTCipherParams fields are public. Therefore, ably-flutter can
```

lib/src/crypto/src/cipher_params_native.dart
```diff
@@ -0,0 +1,60 @@
...
+/// to the dart side.
+class CipherParamsNative implements CipherParams {
+  int? androidHandle;
+  Uint8List? key;
+
+  String? algorithm;
+
+  String? mode = "cbc";
+
+  /// Create a Dart side representation of CipherParams
```
QuintinWillison, line 26:
```suggestion
  /// Create a Dart-side representation of CipherParams
```

lib/src/rest/src/rest_channels.dart
```diff
@@ -0,0 +1,14 @@
+import '../../common/common.dart';
+import '../rest.dart';
+
+/// A collection of rest channel objects
```
QuintinWillison, line 4:
```suggestion
/// A collection of REST channel objects
```

lib/src/rest/src/rest_channel_options.dart
```diff
@@ -1,12 +1,21 @@
...
-  /// create channel options with a cipher
-  RestChannelOptions({this.cipher});
+  /// Create a [RestChannelOptions] directly from a CipherKey. This is
+  /// equivalent to calling the constructor.
+  static Future<RestChannelOptions> withCipherKey(key) async {
+    final cipherParams = await Crypto.getDefaultParams(key: key);
+    return RestChannelOptions(cipherParams: cipherParams);
+  }
+
+  /// create channel options with a cipher.
```
QuintinWillison, line 18:
```suggestion
  /// Create channel options with a cipher.
```

lib/src/crypto/src/crypto.dart
```diff
@@ -0,0 +1,48 @@
+import 'dart:typed_data';
+
+import '../../error/error.dart';
+import '../../generated/platform_constants.dart';
+import '../../platform/platform.dart';
+import '../crypto.dart';
+
+/// Crypto related methods related to creating keys ([generateRandomKey]) and configuration
```
QuintinWillison, line 8: Given the name of this class is literally `Crypto`, I think this commentary could be simplified and made less repetitive:

```suggestion
/// Utility methods for creating keys ([generateRandomKey]) and configuration
```

lib/src/realtime/src/realtime_channel_interface.dart
```diff
@@ -92,6 +94,10 @@ abstract class RealtimeChannelInterface
   /// Warning: the name/ names are not channel names, but message names.
   /// See [Message.dart] for more information.
   ///
+  /// Calling subscribe the first time on a channel will automatically attach
+  /// channel. If a channel is detached, subscribing again will not reattach
```
QuintinWillison, line 98:
```suggestion
  /// that channel. If a channel is detached, subscribing again will not reattach
```

(though I'm surprised this commentary addition falls within the scope of work being done for this pull request)

QuintinWillison (changes requested):
I've done a first pass. Mostly relating to content of commentary.

android/src/main/java/io/ably/flutter/plugin/util/CipherParamsStorage.java
ben-xD: I've now opted to use `Dart side` everywhere because that is much more commonly used already in the files, so this line shouldn't need to change, but the others (with Dart-side) will. 3c13d0fc789cec3c2e1738672dc01284a5f7a2d1

android/src/main/java/io/ably/flutter/plugin/util/CipherParamsStorage.java
ben-xD: Fixed in 3c13d0fc789cec3c2e1738672dc01284a5f7a2d1

android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
ben-xD: 9b42541c1d9a9aae5d83894d5e7d19a458361fdd

README.md
```diff
@@ -518,6 +517,33 @@ channel
...
+    - Distribute a password and nonce from your server and use a key derivation function (KDF) on clients, or
+       - An example of a KDF is [HKDF](https://pub.dev/documentation/cryptography/latest/cryptography/Hkdf-class.html) and [PBKDF2](https://pub.dev/documentation/cryptography/latest/cryptography/Pbkdf2-class.html), available in [`package:cryptography`](https://pub.dev/packages/cryptography)
+    - creating a random one using: `ably.Crypto.generateRandomKey();`. You must to distribute this specific key to other clients, since calling `generateRandomKey` on another device will generate a different key.
+- Create a `CipherParams` instance by passing a key to `final cipherParams = await ably.Crypto.getDefaultParams(key: key);`. The key can be a Base64-encoded `String`, or a `Uint8List`.
+- Create a `RealtimeChannelOptions` or `RestChannelOptions` from this key: e.g. `final channelOptions = ably.RealtimeChannelOptions(cipher: cipherParams);`
+  - Alternatively, if you are only setting CipherParams on ChannelOptions, you could skip creating the `CipherParams` instance: `ably.RestChannelOptions.withCipherKey(cipherKey)` or `ably.RealtimeChannelOptions.withCipherKey(cipherKey)`.
```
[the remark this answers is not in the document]
ben-xD: Thanks, 3faa10d852f6a093e0830a60b4b9be846f0e4d81

android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java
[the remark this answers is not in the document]
ben-xD: All this code was removed in 8ecfce0cfe1eeb3c511ca8b4b6f039062064fab1

The commit message explains: 

>  Remove unused code. This code stopped being used in fc50251f67bcf6b152a75bbdf5b6be3a6f0a5cd4

ikbalkaya (approved):
Light review warning - Looks good to me

README.md
```diff
@@ -518,6 +517,30 @@ channel
...
+
+1. Get or create a key, typically in one of these ways:
+    - Generating keys on your server and distributing them (i.e. using your own APIs)
+    - Distribute a password and nonce from your server and use a key derivation function (KDF) on clients. An example of a KDF is [HKDF](https://pub.dev/documentation/cryptography/latest/cryptography/Hkdf-class.html) and [PBKDF2](https://pub.dev/documentation/cryptography/latest/cryptography/Pbkdf2-class.html), available in [`package:cryptography`](https://pub.dev/packages/cryptography)
+    - creating a random one using: `ably.Crypto.generateRandomKey();`. You must to distribute this specific key to other clients, since calling `generateRandomKey` on another device will generate a different key.
```
lmars, line 527: I don't think it's good to give different options for generating a key without explaining how the developer should choose between those options, but I also don't think it's useful to try and explain that here, it feels more like something that should be explained in depth in our main documentation and just linked to from individual SDK READMEs. I also don't think it's worth mentioning KDFs here either.

So my suggestion is to have a paragraph a bit like [`ably-ruby`](https://github.com/ably/ably-ruby#symmetric-end-to-end-encrypted-payloads-on-a-channel) and [`ably-js`](https://github.com/ably/ably-js#symmetrical-end-to-end-encrypted-payloads-on-a-channel), followed by an example of generating a key using `ably.Crypto.generateRandomKey()`:

> When a 128 bit or 256 bit key is provided to the library, the `data` attribute of all messages is encrypted and decrypted automatically using that key. The secret key is never transmitted to Ably. See https://www.ably.com/documentation/realtime/encryption.
>
> 1. Create a key by calling `ably.Crypto.generateRandomKey()` (or retrieve one from your server using your own secure API)

README.md
```diff
@@ -518,6 +517,30 @@ channel
...
+2. Create a `CipherParams` instance by passing a key to `final cipherParams = await ably.Crypto.getDefaultParams(key: key);` - the key can be a Base64-encoded `String`, or a `Uint8List`
+3. Create a `RealtimeChannelOptions` or `RestChannelOptions` from this key: e.g. `final channelOptions = ably.RealtimeChannelOptions(cipher: cipherParams);`. Alternatively, if you are only setting CipherParams on ChannelOptions, you could skip creating the `CipherParams` instance: `ably.RestChannelOptions.withCipherKey(cipherKey)` or `ably.RealtimeChannelOptions.withCipherKey(cipherKey)`.
+4. Set this options on your channel: `realtimeClient.channels.get(channelName).setOptions(channelOptions);`
```
lmars, line 530:
```suggestion
4. Set these options on your channel: `realtimeClient.channels.get(channelName).setOptions(channelOptions);`
```

example/lib/encrypted_messaging_service.dart
```diff
@@ -0,0 +1,91 @@
...
+
+  Uint8List get keyFromPassword {
```
lmars, line 24: Whilst this is only an example, it's worth stating explicitly that this is not the recommended way to derive a key from a password, production code should use a KDF or some other secure, attack-resistant mechanism.

lib/src/crypto/src/crypto.dart
```diff
@@ -0,0 +1,48 @@
...
+  static const DEFAULT_KEY_LENGTH = 256;
+  static const DEFAULT_MODE = "cbc";
+  static const KEY_TYPE_ERROR_MESSAGE =
+      'A key must either be a String or a Uint8List.';
+
+  /// Gets the CipherParams which can be used to with [RestChannelOptions] or
+  /// [RealtimeChannelOptions] to specify encryption.
+  ///
+  ///  Pass a [String] containing a base64 encoded key or a [Uint8List]
+  ///  containing raw bytes for the key. If you have a password, do not use it
```
lmars, line 22: I think it's worth stating explicitly that the key length here must match the default of 256 bits.

lib/src/crypto/src/crypto.dart
```diff
@@ -0,0 +1,48 @@
...
+  ///  directly, instead you should derive a key from this password,
+  ///  for example by using a key derivation function (KDF) such as PBKDF2.
```
lmars, line 24: Perhaps this would be more useful as an error message when a key that is not 256 bits in length is used?

lmars (commented):
My comments are mainly related to clarifying certain comments from a developer's point of view, but otherwise LGTM 👍

example/lib/encrypted_messaging_service.dart
ben-xD: Thanks. I've done it in 2b6e91b0928251dbcc8f49f916023735c3c57cbc

lib/src/crypto/src/crypto.dart
ben-xD: I actually thought the user can use any size key, and I've tested this and the app crashes if they set a key size which isn't 256 bits. **Thanks**

a6ac0c33ffd6012db31939ba482d19be9f1d02d4

lib/src/crypto/src/crypto.dart
ben-xD: Both comment and error message was added in a6ac0c33ffd6012db31939ba482d19be9f1d02d4

README.md
ben-xD: Okay that makes sense, I've done that in bba9f4c10efd41429285ca0e111a81532eebfb6c

lmars: Nice, thanks 👍

lmars (approved):

tiholic (approved):
Everything looks good to me. One minor optimization w.r.t. dart code:

lib/src/platform/src/platform.dart
```diff
@@ -51,4 +50,27 @@ class Platform {
       }
     }
   }
+
+  /**
+   * Call a platform method which always provides a result.
+   */
+  static Future<T> invokePlatformMethodNonNull<T>(String method,
```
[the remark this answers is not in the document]
ben-xD: I think refactoring it would be difficult currently. This is because `invokeRequest` wraps the data in an AblyMessage:
```
   final message = (null != argument)
        ? AblyMessage(AblyMessage(argument, handle: _handle))
        : AblyMessage(_handle);
```

I would like to remove the nested AblyMessages, and i have a draft PR for this: https://github.com/ably/ably-flutter/pull/199

Then, we can potentially merge these two method implementations.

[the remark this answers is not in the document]
tiholic: That makes sense

PR #198, opened 2021-11-03
Copy unawaited implementation into Ably Flutter to fix build failures for Flutter 2.2.3

ben-xD:
A customer has asked us to add support for Flutter 2.2.3 to avoid them upgrading to Flutter 2.5 in a [slack message](https://ably-real-time.slack.com/archives/C0178QGC093/p1635945748037800). Unfortunately, currently, we are using `unawaited` introduced in Dart 2.14, which is first available in Flutter 2.5.0. 

As a solution, we copied the 1 line of implementation into this SDK, and import this instead.

We are also planning to add tests using older Flutter versions in CI to avoid this problem in the future, but there are some challenges in doing so, which we will tackle in a separate (currently draft) PR: https://github.com/ably/ably-flutter/pull/197

### DIFF
--- a/lib/src/common/src/backwards_compatibility.dart
+++ b/lib/src/common/src/backwards_compatibility.dart
@@ -1,0 +1,14 @@
+/// This function was added to add backwards compatibility.
+///
+/// `unawaited` was previously imported in code from package:pedantic.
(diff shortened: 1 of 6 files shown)


QuintinWillison (approved):
❤️

PR #199, opened 2021-11-04
 Fix enum to be more conventional and simplify AblyMessage structure

ben-xD:
- I simplify the AblyMessage structure, so that there are fewer nested messages. To do this, I add a `handle` field to `AblyEventMessage` directly. This also allows me to stop using `ablyDo`, IMHO this extra level of indirection was confusing to me when first joining (it certainly confused Android Studio, since it caused a lot of linter errors). What makes it extra confusing was the logic that only **sometimes** created nested messages.
  - Most of the changes in this PR is probably due to this change.
- I rename the cases in the CodecType enum to be the standard naming convention. This means when it used in Swift, it will be converted nicely: e.g. `CodecTypeAblyMessage` in Objective-C will appear to be `CodecType.AblyMessage`. Currently, AblyMessageCodecType will appear as `_Value.ablyEventMessageCodecType` in Swift, because the enum is declared as `typedef NS_ENUM(UInt8, _Value)`.
- I rename classes in both Android and iOS to `AblyClientStore`. On iOS it was called AblyFlutter (confused with AblyFlutterPlugin), and in Android it was called `AblyLibrary` (vague). What does it do? It **stores ably client instances**. I've actually mostly rewritten this class in Swift, but that is for a future PR. This also allows me (in a future PR) to rename `AblyFlutterPlugin` to `AblyFlutter`: this is the main class for the plugin.
- I rename a method from `realtimeWithHandle` to `getRealtime` to be consistent with the rest equivalent: `getRest`.
- Removed a file that was duplicated from the Android library (BiConsumer) and just importing it instead

PS: Please excuse the identation / formatting. I did not want to auto format iOS and Android code since they have not been formatted for a while, and the diff would be unnecessarily large. We will tackle this in a separate PR https://github.com/ably/ably-flutter/issues/216


ben-xD:
Closing this PR because it mixes up too many things: https://github.com/ably/ably-flutter/issues/245

PR #202, opened 2021-11-04
Simplify state in example app

ben-xD:
With this change, users of the example app no longer need to create instances of Realtime and Rest. These get created automatically. This simplifies the state for the entire app, as the rest of the app can rely on Realtime and Rest instances existing. This allows me to avoid adding more complicated state management tools, like Bloc, getx, provider, etc for my upcoming symmetric encryption PR.

The UI changes from (before, left) and after (right): 
Frame 109

### DIFF
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -32,8 +32,6 @@ class _MyAppState extends State<MyApp> {
   String _ablyVersion = 'Unknown';
 
   final String _apiKey = const String.fromEnvironment(Constants.ablyApiKey);
-  OpState _realtimeCreationState = OpState.notStarted;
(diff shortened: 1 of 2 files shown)


ikbalkaya (approved):
(Attention: light review) Looks good to me

QuintinWillison (approved):
I'm on the fence as to whether I'm truly in favour of this change, as the example app was originally designed to be very manually driven, which can be helpful when debugging as-yet-unidentified issues.

Given I know that we have reasons that we're pressured to deliver #192, and this PR is now blocking that, I guess we need to land this. But I would like to revisit this so I shall create an issue for that.

PR #203, opened 2021-11-05
Remove a duplicate class and use the other one instead

ben-xD:
This is a small fix, there was duplicate class, and I'm just deleting one. The file being referenced instead is https://github.com/ably/ably-flutter/blob/main/lib/src/platform/src/rest/publish_queue_item.dart


ikbalkaya (approved):
LGTM

QuintinWillison (approved):

PR #210, opened 2021-11-09
Update `flutter_local_notifications` to avoid bintray dependency resolution error

ben-xD:
We must update `flutter_local_notifications` to allow builds to succeed. The version we currently use uses bintray, and it was not working today as per our own experiences trying to download dependencies on @ikbalkaya's machine, and https://www.reddit.com/r/devops/comments/qq3bd9/gradle_ci_fails_because_jcenterbintray_is_down/. 

I also took this opportunity to update `device_info_plus`.


ikbalkaya (approved):
Nice, this fixed issue with dependency resolution

PR #215, opened 2021-11-10
0 Fix app crashing when app is not yet started on some devices

ben-xD:
There are actually **three** bugs:
- `activityManager.getRunningAppProcesses()` can return null when no processes are running. I was iterating over it, assuming it would be an empty list instead of null. This caused a NPE like crash. (3 line fix, excluding comments)
- When the app is in terminated state, no messages are received. Solution: launch `BackgroundIsolateAndroidPlatform`. (one line fix)
- More difficult to solve: When an app is in terminated state, and a message is received, subsequent messages are not received. Specifically, these subsequent messages do not arrive to the Dart side. This is because I was using a MethodChannel created when no activity existed. When the activity is created, a new FlutterEngine (and FlutterJNI) is created. I have to use this new environment instead of the old method channel (which used the old FlutterJNI that was detached). This is done in `PushMessagingEventHandlers.load(applicationContext, methodChannel);`
  - error: `Tried to send a platform message to Flutter, but FlutterJNI was detached from native C++.`. 

See customer report: https://ably-real-time.slack.com/archives/C012DA29VM0/p1636555532089600?thread_ts=1636544116.088600&cid=C012DA29VM0

### DIFF
--- a/android/src/main/java/io/ably/flutter/plugin/push/FirebaseMessagingReceiver.java
+++ b/android/src/main/java/io/ably/flutter/plugin/push/FirebaseMessagingReceiver.java
@@ -80,6 +80,11 @@ public class FirebaseMessagingReceiver extends BroadcastReceiver {
     // This only shows processes for the current android app.
     final List<ActivityManager.RunningAppProcessInfo> appProcesses = activityManager.getRunningAppProcesses();
 
+    if (appProcesses == null) {
+      // appProcesses is null if no processes are running.
+      return false;
+    }
+
     for (ActivityManager.RunningAppProcessInfo process : appProcesses) {
       // Importance is IMPORTANCE_SERVICE (not IMPORTANCE_FOREGROUND)
       //  - when app was terminated, or
--- a/android/src/main/java/io/ably/flutter/plugin/push/FirebaseMessagingReceiver.java
+++ b/android/src/main/java/io/ably/flutter/plugin/push/FirebaseMessagingReceiver.java
@@ -1,8 +1,6 @@
 package io.ably.flutter.plugin.push;
 
 import static io.ably.flutter.plugin.push.PushMessagingEventHandlers.PUSH_ON_BACKGROUND_MESSAGE_PROCESSING_COMPLETE;
-import static io.ably.flutter.plugin.push.PushMessagingEventHandlers.PUSH_ON_BACKGROUND_MESSAGE_RECEIVED;
-import static io.ably.flutter.plugin.push.PushMessagingEventHandlers.PUSH_ON_MESSAGE_RECEIVED;
 
 import android.app.ActivityManager;
 import android.content.BroadcastReceiver;
@@ -13,8 +11,6 @@ import android.util.Log;
 
 import androidx.annotation.Nullable;
 import androidx.localbroadcastmanager.content.LocalBroadcastManager;
-
-import com.google.firebase.messaging.RemoteMessage;
 
 import java.util.List;
 
@@ -55,24 +51,15 @@ public class FirebaseMessagingReceiver extends BroadcastReceiver {
   }
 
   private void sendMessageToFlutterApplication(Context context, Intent intent) {
-    final RemoteMessage message = new RemoteMessage(intent.getExtras());
     final Boolean isApplicationInForeground = isApplicationInForeground(context);
 
     if (isApplicationInForeground) {
-      // Send message to Dart side app already running
-      final Intent onMessageReceivedIntent = new Intent(PUSH_ON_MESSAGE_RECEIVED);
-      onMessageReceivedIntent.putExtras(intent.getExtras());
-      LocalBroadcastManager.getInstance(context).sendBroadcast(onMessageReceivedIntent);
+      PushMessagingEventHandlers.sendMessageToApp(context, intent);
     } else if (AblyFlutterPlugin.isMainActivityRunning) {
-      // Flutter is already running, just send a background message to it.
-      final Intent onMessageReceivedIntent = new Intent(PUSH_ON_BACKGROUND_MESSAGE_RECEIVED);
-      onMessageReceivedIntent.putExtras(intent.getExtras());
-      LocalBroadcastManager.getInstance(context).sendBroadcast(onMessageReceivedIntent);
+      PushMessagingEventHandlers.sendBackgroundMessage(context, intent);
     } else {
-      // No existing Flutter Activity is running, create a FlutterEngine and pass it the RemoteMessage
-      new PushTerminatedIsolateRunner(context, this, message);
+      new AppRunner(context, this, intent);
     }
-
   }
 
   private Boolean isApplicationInForeground(final Context context) {
--- a/android/src/main/java/io/ably/flutter/plugin/push/FirebaseMessagingReceiver.java
+++ b/android/src/main/java/io/ably/flutter/plugin/push/FirebaseMessagingReceiver.java
@@ -81,7 +81,8 @@ public class FirebaseMessagingReceiver extends BroadcastReceiver {
     final List<ActivityManager.RunningAppProcessInfo> appProcesses = activityManager.getRunningAppProcesses();
 
     if (appProcesses == null) {
-      // appProcesses is null if no processes are running.
+      // If no processes are running, appProcesses are null, not an empty list.
+      // There is definitely not an application in foreground if no processes are running.
       return false;
     }
 
--- a/android/src/main/java/io/ably/flutter/plugin/push/FirebaseMessagingReceiver.java
+++ b/android/src/main/java/io/ably/flutter/plugin/push/FirebaseMessagingReceiver.java
@@ -69,7 +69,7 @@ public class FirebaseMessagingReceiver extends BroadcastReceiver {
 
     if (appProcesses == null) {
       // If no processes are running, appProcesses are null, not an empty list.
-      // There is definitely not an application in foreground if no processes are running.
+      // The user's app is definitely not in the foreground if no processes are running.
       return false;
     }
 
--- a/android/src/main/java/io/ably/flutter/plugin/push/FirebaseMessagingReceiver.java
+++ b/android/src/main/java/io/ably/flutter/plugin/push/FirebaseMessagingReceiver.java
@@ -70,7 +70,7 @@ public class FirebaseMessagingReceiver extends BroadcastReceiver {
       LocalBroadcastManager.getInstance(context).sendBroadcast(onMessageReceivedIntent);
     } else {
       // No existing Flutter Activity is running, create a FlutterEngine and pass it the RemoteMessage
-      new PushBackgroundIsolateRunner(context, this, message);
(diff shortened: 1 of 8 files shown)


android/src/main/java/io/ably/flutter/plugin/push/FirebaseMessagingReceiver.java
```diff
@@ -80,6 +80,11 @@ private Boolean isApplicationInForeground(final Context context) {
     // This only shows processes for the current android app.
     final List<ActivityManager.RunningAppProcessInfo> appProcesses = activityManager.getRunningAppProcesses();
 
+    if (appProcesses == null) {
+      // appProcesses is null if no processes are running.
```
ikbalkaya, line 84: This code block is pretty much self explanatory, I think you should be ok to remove this comment 😄

ikbalkaya (changes requested):
Looks good, I just requested a comment removal

android/src/main/java/io/ably/flutter/plugin/push/FirebaseMessagingReceiver.java
ben-xD: I'd to go the other way: add more comments: 56eb5316672cd2fb5ec9d19355a32bb62a1f8299

`getRunningAppProcesses` returns null when no processes are running, not an emptyList.

ikbalkaya: As you pointed out in your last comment, the doc for the function is quite clear
> Returns a list of application processes that are running on the device.
> Note: this method is only intended for debugging or building a user-facing process management UI.
> Returns:
> Returns a list of RunningAppProcessInfo records, or null if there are no running processes (it will not return an empty list). This list ordering is not specified.

. However I understand why you don't find it intuitive. So up to you

ikbalkaya (approved):
LGTM

android/src/main/java/io/ably/flutter/plugin/push/ManualFlutterApplicationRunner.java
```diff
@@ -21,20 +20,27 @@
...
-public class PushBackgroundIsolateRunner implements MethodChannel.MethodCallHandler {
-  private static final String TAG = PushBackgroundIsolateRunner.class.getName();
-  private static final String SHARED_PREFERENCES_KEY = "io.ably.flutter.plugin.push.PushBackgroundIsolate.SHARED_PREFERENCES_KEY";
-  private static final String BACKGROUND_MESSAGE_HANDLE_KEY = "BACKGROUND_MESSAGE_HANDLE_KEY";
+/**
+ * This class is used when the application was terminated when the push notification is received.
+ * See [PushMessagingEventHandlers.java] to see where push notifications being handled whilst the
+ * app is in the background or the foreground.
+ *
+ * This class can generified to launch the app manually for any purpose, but currently it
```
ikbalkaya, line 28: "This class can be generalised" ....

android/src/main/java/io/ably/flutter/plugin/push/ManualFlutterApplicationRunner.java
```diff
@@ -21,20 +20,27 @@
...
+ * launches the users app and sends it a RemoteMessage.
```
ikbalkaya, line 29: "launches the app in use"

android/src/main/java/io/ably/flutter/plugin/push/ManualFlutterApplicationRunner.java
```diff
@@ -21,20 +20,27 @@
...
+  private static final String TAG = ManualFlutterApplicationRunner.class.getName();
   private final FirebaseMessagingReceiver broadcastReceiver;
   private final RemoteMessage remoteMessage;
   private final MethodChannel backgroundMethodChannel;
 
   @NonNull
   private final FlutterEngine flutterEngine;
-
-  public PushBackgroundIsolateRunner(Context context, FirebaseMessagingReceiver receiver, RemoteMessage message) {
+  // Launch the users Flutter app and pass it the RemoteMessage.
```
ikbalkaya, line 39: "Launch the flutter app in use" ... or "User's flutter app"

android/src/main/java/io/ably/flutter/plugin/push/ManualFlutterApplicationRunner.java
```diff
@@ -21,20 +20,27 @@
...
+  // Use this when noo existing Flutter Activity is running.
```
ikbalkaya, line 40: "no"

android/src/main/java/io/ably/flutter/plugin/push/ManualFlutterApplicationRunner.java
```diff
@@ -21,20 +20,27 @@
...
+  public ManualFlutterApplicationRunner(Context context, FirebaseMessagingReceiver receiver, Intent intent) {
     this.broadcastReceiver = receiver;
-    this.remoteMessage = message;
+    this.remoteMessage = new RemoteMessage(intent.getExtras());
```
ikbalkaya, line 43: What if intent or intent.getExtras is null? You might want to annotate non-null fields with @NonNull

android/src/main/java/io/ably/flutter/plugin/push/PushMessagingEventHandlers.java
```diff
@@ -16,21 +15,52 @@
 import io.ably.flutter.plugin.generated.PlatformConstants;
 import io.flutter.plugin.common.MethodChannel;
 
+/**
+ * This class is used when the application is in the **foreground** or **background**, but not when
+ * **terminated**. See [PushTerminatedIsolateRunner.java] to see where push notifications being
+ * handled whilst the app is terminated.
+ */
```
ikbalkaya, line 22: 👍 Helpful comment

android/src/main/java/io/ably/flutter/plugin/push/PushMessagingEventHandlers.java
```diff
@@ -16,21 +15,52 @@
...
+  }
+
+  private final BroadcastReceiver broadcastReceiver;
   private PushMessagingEventHandlers(Context context, MethodChannel methodChannel) {
-    new BroadcastReceiver(context, methodChannel);
+    this.broadcastReceiver = new BroadcastReceiver(context, methodChannel);
+  }
+
+  // Send message to Dart side app already running
+  public static void sendMessageToFlutterApp(Context context, Intent intent) {
```
ikbalkaya, line 53: Can you add @NonNull annotation to the parameters please? 🙇‍♂️

android/src/main/java/io/ably/flutter/plugin/push/PushMessagingEventHandlers.java
```diff
@@ -16,21 +15,52 @@
...
+    final Intent onMessageReceivedIntent = new Intent(PUSH_ON_MESSAGE_RECEIVED);
+    onMessageReceivedIntent.putExtras(intent.getExtras());
+    LocalBroadcastManager.getInstance(context).sendBroadcast(onMessageReceivedIntent);
+  }
+
+  // Flutter is already running, just send a background message to it.
+  public static void sendBackgroundMessageToFlutterApp(Context context, Intent intent) {
```
ikbalkaya, line 60: NonNull annotation...

ikbalkaya (changes requested):
Looks good overall. I made some requests about some minor comment changes. There are also some places where nullable variables are accessed without any check or annotation.

android/src/main/java/io/ably/flutter/plugin/push/ManualFlutterApplicationRunner.java
ben-xD: c2384f5cef7ff1b04033d599516e0b293a32151b

android/src/main/java/io/ably/flutter/plugin/push/ManualFlutterApplicationRunner.java
ben-xD: c4e8f56470732dffa0cf23c0830c79f9a8b52199

android/src/main/java/io/ably/flutter/plugin/push/ManualFlutterApplicationRunner.java
ben-xD: 25716c31d55e8d765840c339cf5b3b4eea191020

android/src/main/java/io/ably/flutter/plugin/push/ManualFlutterApplicationRunner.java
ben-xD: 25716c31d55e8d765840c339cf5b3b4eea191020

android/src/main/java/io/ably/flutter/plugin/push/PushMessagingEventHandlers.java
ben-xD: I would say please assume parameters are non null unless they are declared to be nullable. However, I have also added them here: 3021a28d0802b54ca5bbf7469122acbf99b6874f

android/src/main/java/io/ably/flutter/plugin/push/PushMessagingEventHandlers.java
ben-xD: I would say please assume parameters are non null unless they are declared to be nullable. However, I have also added them here: 3021a28d0802b54ca5bbf7469122acbf99b6874f

android/src/main/java/io/ably/flutter/plugin/push/ManualFlutterApplicationRunner.java
ben-xD: If the extras is null or is wrongly structured, the remote message **still** gets created, but calling `getNotification` or `getData` leads to a NPE.

This won't happen because the FirebaseMessagingReceiver only receives intents with a specific action. There is no easy way to confirm the intent is structured correctly to add additional safety. But IMHO this case will never happen.

android/src/main/java/io/ably/flutter/plugin/push/ManualFlutterApplicationRunner.java
ikbalkaya: Sorry to be a pain, but can you make this passive. "This class can BE generalised ...

ben-xD: c06b6a3293b7690d371fb7369b06322de50809fe

ikbalkaya (approved):
Looks good :shipit:

tiholic (approved):

android/src/main/java/io/ably/flutter/plugin/push/PushMessagingEventHandlers.java
ben-xD: Resolving this because you have resolved the other conversation which had the same issue which has fixed in the same commit :).

PR #219, opened 2021-11-11
Stop saving a reference to a Future as a way to check for the method being called

ben-xD:
We are currently using the `Future` to keep track if we actually called the function yet or not. In this PR, I make it explicit by using a `_shouldResetAblyClients` boolean instead.

Whilst I am here, I decided to understand what "registerAbly" does. It doesn't **register** anything, it is called when the Plugin registers, but it specifically resets all the client instances (realtime and rest) stored inside `AblyLibrary`. I've called it `resetAblyClients`.


android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java
```diff
@@ -43,29 +43,28 @@
 public class AblyMethodCallHandler implements MethodChannel.MethodCallHandler {
   private Context applicationContext;
 
-  public interface HotRestartCallback {
-    // Called on every fresh start and on every hot restart
+  public interface OnClearPlatformInstances {
     void on();
```
tiholic, line 47: I think this method name can be renamed to suit better.

For example: `HotRestartListener.onHotRestart` would have made more sense with older class name

lib/src/platform/src/platform.dart
```diff
@@ -21,27 +21,29 @@ class Platform {
...
-  static Future? _initializer;
+  /// This field will reset to true when the state is reset. This allows us to
+  /// if an app has been restarted, or a hot restart (not hot reload) has
+  /// happened.
+  static bool _isStateReset = true;
 
-  static Future _initialize() async {
-    if (_initializer == null) {
+  /// Clears instances on the Platform side
+  static Future<void> _handleNewState() async {
```
tiholic, line 30: The name `_handleNewState` doesn't fit well in here. However, not very important as it is a private method.

tiholic (approved):

android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java
ben-xD: If we observe the behaviour of `registerAbly`, this dart side makes the method call to clear instances of clients in the platform side. This method calls the `HotRestartCallback`. I think it's not immediately obvious that when we register the plugin, we run the hot restart callback, and why.

It actually doesn't happen only onHotRestart, it happens whenever the platform method is called. Since the platform method clears instances of clients, I renamed `registerAbly` to `clearPlatformInstances`. This callback interface does exactly that, it clears platform instances.

lib/src/platform/src/platform.dart
ben-xD: I tried to pick a name which means both the app is just launching, but also the hot restart is called. Perhaps I could call it `isStateReset`.

ben-xD: In 4dce9d58a0d5fffd8f936e38efb0e758337642cc, I have renamed `_isStateReset` to `_shouldClearPlatformInstances`. And `_handleNewState` to `ensureFreshState`. What do you think?

android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java
ben-xD: I have renamed it to `ResetAblyClientsCallback`. I think this makes sense. What do you think?

lib/src/platform/src/platform.dart
```diff
@@ -18,27 +18,29 @@ class Platform {
   static final StreamsChannel streamsChannel =
       StreamsChannel('io.ably.flutter.stream', codec);
 
-  /// Initializing ably on platform side by invoking `register` platform method.
-  /// Register will clear any stale instances on platform.
-  static Future? _initializer;
+  /// This field will reset to true when the state is reset. This allows us to
+  /// if an app has been restarted, or a hot restart (not hot reload) has
+  /// happened.
+  static bool _shouldResetAblyClients = true;
```
ikbalkaya, line 24: Could you use a non-static field here 🤔

lib/src/platform/src/platform.dart
```diff
@@ -18,27 +18,29 @@ class Platform {
...
 
-  static Future _initialize() async {
-    if (_initializer == null) {
+  /// Clears instances on the Platform side
+  static Future<void> _resetOldAblyClients() async {
```
ikbalkaya, line 27: Sorry to be picky on static methods but are we not performing this methods on Platform object here? Using statics will defeat the purpose of encapsulation and will cause the design to consume other static states which makes application unmaintainable. I think those static methods, fields should be used a bit more cautiously

lib/src/platform/src/platform.dart
```diff
@@ -18,27 +18,29 @@ class Platform {
...
+    if (_shouldResetAblyClients) {
```
ikbalkaya, line 28: I think this check should be done where method call is done

lib/src/platform/src/platform.dart
```diff
@@ -18,27 +18,29 @@ class Platform {
...
       AblyMethodCallHandler(methodChannel);
-      _initializer = methodChannel.invokeMethod(PlatformMethod.registerAbly);
       BackgroundIsolateAndroidPlatform();
+      await methodChannel.invokeMethod(PlatformMethod.resetAblyClients);
+      _shouldResetAblyClients = false;
```
ikbalkaya, line 32: IMO it's always safer to use a callback or a state object that informs the callers than maintaining the state yourself.  this might be out of scope of this PR though

ikbalkaya (dismissed):
Looks good 😎 apart from some static method, field usages.

lib/src/platform/src/platform.dart
ben-xD: Not if everything else is static in this file 😂 . This conversation can be continued in the other: https://github.com/ably/ably-flutter/pull/219#discussion_r752149420

QuintinWillison (commented):
Ignoring the small comment I've made on one of your comments, I've spent perhaps 15 minutes trying to understand the intent of this pull request and I am really struggling. I don't understand what problem you're solving, nor what you mean when you say in your opening comment:

> We are currently using the Future to keep track if we actually called the function yet or not. In this PR, I make it explicit by using a _shouldResetAblyClients boolean instead.

I can't see what is more 'explicit'. Can you explain what you mean by that, as the changes you've made here simply open more questions for me.

Fundamentally I don't understand what I'm reviewing here, why it improves the codebase or what drove this change. The original work you're unwinding here, done under #18, seemed pretty well considered at the time. Are you fixing a bug? Or is this just down to 'taste'?

I also see no evidence of tests that prove this change fixes something specific in the behaviour. I remember reload and restart being pretty tricky to debug and test when we were in the early days of developing this plugin, so I would be very cautious of making major structural changes in this area without a concrete vision as to what's being solved by doing them.

lib/src/platform/src/platform.dart
ben-xD: Don't be sorry, its definitely useful to think about this.

Everything in `Platform` is static. There a few reasons why I can come up with:
- Avoids dependency injection: If platform was an instance, everything would need to be injected with it **OR** platform could just be an instance stored inside a singleton class **OR** just be singleton itself. 
- Platform is a special class where there is always only one that should exist in the application. It could be more suited for Singleton pattern.
- Some static methods in Ably Flutter (e.g. Crypto.getDefaultParams) still need to use the method channel (state) from Platform. These methods are defined to be static in the feature spec. 

https://github.com/ably/ably-flutter/blob/78144602ab1b63568e3eaea4f79b7035e541231a/lib/src/crypto/src/crypto.dart#L18-L40

Your comment prompted me to create this PR to attempt to solve this problem by using the Singleton pattern and making the state of Platform private, however the last final issue: I can't make some fields that should be internal to the class private because they are **being used in tests.**

What do you think of https://github.com/ably/ably-flutter/pull/235 ?

lib/src/platform/src/platform.dart
ben-xD: `_shouldResetAblyClients` is actually the **state** field. When the app restarts or hot-restarts, this state field is set to `true` automatically because the class is re-instantiated:
```dart
  static bool _shouldResetAblyClients = true;
```

I think the mechanics here is simpler than you think, but I'm not sure

lib/src/platform/src/platform.dart
ben-xD: This has been moved in a separate PR https://github.com/ably/ably-flutter/pull/235/files which does more refactoring 2fd960927973d2197aac957f8e3e172c20743657

lib/src/platform/src/platform.dart
```diff
@@ -18,27 +18,29 @@ class Platform {
   static final StreamsChannel streamsChannel =
       StreamsChannel('io.ably.flutter.stream', codec);
 
-  /// Initializing ably on platform side by invoking `register` platform method.
-  /// Register will clear any stale instances on platform.
-  static Future? _initializer;
+  /// This field will reset to true when the state is reset. This allows us to
+  /// if an app has been restarted, or a hot restart (not hot reload) has
```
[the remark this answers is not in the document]
ben-xD: That's been done in 8eac44ebc0230356f5f89785e7cb2b03575cda5f

ben-xD:
It's good that you are unearthing the previous work that was done. However, this PR does not change any logic, it only renames and refactors, to make it more understandable. I also add more comments to explain what is going on. As I improve the readability, I can work with the SDK better, helping me find bugs, which will be fixed (and have been fixed) in other PRs. I'm aiming to keep the PRs as understandable as possible.

`registerAbly` is pretty vague (and below I explain that it is **misleading**). The platform method `resetAblyClients` is exactly what it does, it clears all the Ably clients in the platform side. The logic is all the same.

Let me elaborate on:
> We are currently using the Future to keep track if we actually called the function yet or not. 

By function, I mean platform method called `registerAbly`. The Platform class has a field `static Future? _initializer;` which is used to check if the app has been restarted or hot reloaded. The `initializer` field is set to a Future returned by the platform method call to registerAbly. Correct use of a Future would be to (optionally) wait for when the platform method completes, or ignore it. Instead, this method is just saving the future to know if the platform method was called. It is uses a `Future` which holds a value (possibly nil in this case) to store what a boolean should (the fact that platform method was called), we don't need to store this Future for the rest of the app.

**PS:** There is actually another concept called registering a Flutter plugin, which is not related to this at all. This is when the plugins `onAttachedToEngine` method is called (which is called when `GeneratedPluginRegistrant` registers all the plugins). That is one of the motivations for not calling it `registerAbly`, plugins are registered when the platform host launches, not when the Flutter application restart. In summary, our platform method `registerAbly` does **not** register Ably. It clears Ably client instances.

However, I do have more logic changes coming, which will remove this boolean completely, as it is not necessary once we remove `static` methods from Dart. That will not change the outcome of restarting the app and hot restarting, **I will be sure to record a video showing all the functionality still working :), since hot restart is a little bit difficult to write a test for.**

lib/src/platform/src/platform.dart
ben-xD: And in actuality, if we weren't using static fields here, the constructor would just be called again. So all this logic drops out completely. This has been done in a future PR: https://github.com/ably/ably-flutter/pull/235

lukasz-szyszkowski (approved):
LGTM

android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java
QuintinWillison: @ben-xD Reading @tiholic's opening comment for this conversation, he's commenting on the naming of the _method_, not the _interface_. I agree, that `on` is a very opaque _method_ name - albeit inherited by you from the previous state of the codebase - but perhaps changing meaning now that you've renamed the owning interface. So, I think there's still work to do here to address his comment, please.

lib/src/platform/src/platform.dart
QuintinWillison: It looks like @tiholic's opening comment for this conversation has been addressed so I'm going to resolve this on his behalf.

lib/src/platform/src/platform.dart
QuintinWillison: I'm resolving this conversation on @ikbalkaya's behalf.

lib/src/platform/src/platform.dart
QuintinWillison: I'm resolving this conversation on @ikbalkaya's behalf, given this is being addressed (as @ben-xD states) under #235.

lib/src/platform/src/platform.dart
QuintinWillison: Perhaps the addition of some log output when you test `_shouldResetAblyClients` would help ascertain whether / how often these code paths get executed? I don't see a sign of any tests that verify this functionality, so I'm assuming you have to observe this with manual testing. In which case, perhaps you could document your test procedure for that? This is all very opaque to me from just looking at this code - so very hard to review.

android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java
ben-xD: I've renamed `on` to `run`, so it reads like this: `resetAblyClientsCallback.run();`. I'd like to improve it but that's the best I have so far. bf01732c9b90ecdc5168b78eceea20aee807cb33

lib/src/platform/src/platform.dart
ben-xD: I actually used the debugger to test this, so I didn't use any log lines (hot restart/reload are only relevant for debug mode). However, I've attached a video to confirm that it works. This logic (and field) is being replaced with just the constructor in https://github.com/ably/ably-flutter/pull/235.

https://user-images.githubusercontent.com/24711048/142890615-6413d0ff-c651-4fd0-b5d4-f261a4e605c4.mp4

I can't think of a test procedure I could write. Just restart the app or trigger hot restart, and this will be called as expected. I **could** explain the difference between starting the app, hot restart and reload like I did in the video though, so I've done that in 72472a8392d6d2427cc9a3071648a9635686cdb9. However, going into how to do it will depend what tool you use. If you use the terminal (flutter run), android studio or vscode. 

In the terminal (flutter run), you'll get:
Screenshot 2021-11-22 at 15 55 58

In IDE, saving triggers hot restart.
Pressing something triggers hot reload. (Android studio, its the bug/debug icon)
Pressing something else triggers hot restart. (its noticeably slow) its the "run again" button (see screenshot of button)
CleanShot 2021-11-22 at 15 54 56

Documenting this stuff will get outdated easily, and it's also relatively available information.

android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java
QuintinWillison: That is clearer, thanks.

android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java
```diff
@@ -40,31 +40,27 @@
 import io.flutter.plugin.common.MethodChannel;
 
 public class AblyMethodCallHandler implements MethodChannel.MethodCallHandler {
-  private Context applicationContext;
-
-  public interface HotRestartCallback {
-    // Called on every fresh start and on every hot restart
-    void on();
+  public interface ResetAblyClientsCallback {
```
QuintinWillison, line 43: I guess we should just ignore how much this looks like [a `Runnable`](https://developer.android.com/reference/java/lang/Runnable), and move on... 😆

QuintinWillison (approved):

PR #222, opened 2021-11-12
Fix `setOptions` to complete the Future (previously did not call `result.success`) and remove side effects in example app

ben-xD:
A few small things:

- I fix a bug in channelOptions: I now finish the platform method execution (complete the Future returned by `setOptions`) 
- I remove side effect of publishing messages when you launch the app. There are buttons for this functionality, and it confusing to see these messages get automatically sent when the app is launched.


ikbalkaya (approved):
Looks good to me 👍

QuintinWillison (approved):

PR #224, opened 2021-11-15
iOS: add `notification` in RemoteMessage

ben-xD:
On iOS, the notification is kept inside the `userInfo` dictionary. To be consistent with Android, I extract the notification from this dictionary and pass it to the dart side. 

This allows the RemoteMessage to contain a `notification` if one is sent when the app is the foreground, background or terminated. If receiving a push notification in the example app, it will now change from left screenshot to right screenshot. Notice, the original `aps.alert` object in **Notification data** is still there. I will not modify the `data` object or remove the any title/body fields, to avoid backwards compatibility issues with Ably users or confusion. 
Frame 110


ios/Classes/PushNotificationEventHandlers.swift
```diff
@@ -59,9 +59,33 @@ public class PushNotificationEventHandlers: NSObject, UNUserNotificationCenterDe
...
+            }
+        }
+        
+        if (notification.title != nil || notification.body == nil) {
+            // Return an instance of notification only if it has some data.
+            return notification;
+        }
+        
+        return nil;
+    }
```
lukasz-szyszkowski, line 90: how about this?
```swift
class Notification: NSObject {
  @objc
  var title: String?
  
  @objc
  var body: String?
  
  init?(title: String?, body: String?) {
    guard title != nil || body != nil else {
      return nil
    }
    
    self.title = title
    self.body = body
  }
}

// And the method

private func createNotification(from userInfo: [AnyHashable : Any]) -> Notification? {
  guard  let alert = userInfo["aps"] as? [AnyHashable: Any] else {
    return nil
  }
  
  return Notification(
    title: alert["title"] as? String, 
    body: alert["body"] as? String
  )
}

```

ben-xD: I wanted to only create a notification if either title and body is set. If title and body is both null, your code would still create an empty `notification`. TBH, this is a case which might never happen though, `aps` might always be set. What do you think?

lukasz-szyszkowski: In my example if title is nil and body is nil the Notification object won’t be created, that’s what guard returns.

ios/Classes/PushNotificationEventHandlers.swift
```diff
@@ -59,9 +59,33 @@ public class PushNotificationEventHandlers: NSObject, UNUserNotificationCenterDe
...
+                if let title = alert["title"] as? String {
+                    notification.title = title
+                }
+                if let body = alert["body"] as? String {
+                    notification.body = body
+                }
+            }
+        }
+        
+        if (notification.title != nil || notification.body == nil) {
```
lukasz-szyszkowski: You are returning an object if both values are nil, this is what you want?

ios/Classes/PushNotificationEventHandlers.swift
ben-xD: I've tried to implement this in a slightly different way, and there I found an issue with using convenience initializers with **failable (returns nil)** initializer. The compiler wants me to call `self.init`:

```
class Notification: NSObject {
    @objc
    var title: String?;
    @objc
    var body: String?;
    
    init?(title: String?, body: String?) {
      guard title != nil || body != nil else {
        return nil
      }
      
      self.title = title
      self.body = body
    }
    
    convenience init?(from userInfo: [AnyHashable : Any]) {
        if let aps = userInfo["aps"] as? NSDictionary {
            if let alert = aps["alert"] as? NSDictionary {
                let title = alert["title"] as? String
                let body = alert["body"] as? String
                self.init(title: title, body: body)
            } else {
                self.init(title: nil, body: nil)
                return nil
            }
        } else {
            self.init(title: nil, body: nil)
            return nil
        }
    }
}
```

- Notice the ugly usage of `self.init(title: nil, body: nil)` and `return nil`. Without `return nil`, we would return the instance.
- I cannot return `self` to finish this initializer early, which would have made it cleaner. 
CleanShot 2021-11-18 at 08 46 30

lukasz-szyszkowski: Well, you can't use return `self` in swift initialisers, if it's optional what you can return is just `nil`. So remove `return self` and change `self.init(title: nil, body: nil)` to `return nil` and it should be OK

lukasz-szyszkowski: and please change casting from `NSDictionary` (it's obj-c) to `[AnyHashable: Any]`

ben-xD: I double checked if `title` is always available in userInfo.aps.alert, and it is not: See example 4 in https://developer.apple.com/library/archive/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/CreatingtheNotificationPayload.html#//apple_ref/doc/uid/TP40008194-CH10-SW1:
```
{
    "aps" : {
        "alert" : {
            "loc-key" : "GAME_PLAY_REQUEST_FORMAT",
            "loc-args" : [ "Jenna", "Frank"]
        },
        "sound" : "chime.aiff"
    },
    "acme" : "foo"
}
```

This happens when the user wants to use localization using iOS files (Localizable.strings file). This is not supported by Ably Flutter. See [Alert Keys](https://developer.apple.com/library/archive/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/PayloadKeyReference.html#//apple_ref/doc/uid/TP40008194-CH17-SW1) for more information. 

So, in summary, it is safe to say if `alert` is set, the user should also set at least a `title`. 61bbbe22601998faf08b070a88ec92ba78376d41

ios/Classes/Notification.swift
```diff
@@ -0,0 +1,29 @@
...
+    let body: String?;
+    
+    init(title: String?, body: String?) {
+      self.title = title
+      self.body = body
+    }
+    
+    convenience init?(from userInfo: [AnyHashable : Any]) {
+        if let aps = userInfo["aps"] as? NSDictionary {
+            if let alert = aps["alert"] as? NSDictionary {
```
lukasz-szyszkowski, line 16:
```suggestion
            if let alert = aps["alert"] as? [AnyHashable: Any] {
```

ios/Classes/Notification.swift
```diff
@@ -0,0 +1,29 @@
...
+    @objc
+    let body: String?;
+    
+    init(title: String?, body: String?) {
+      self.title = title
+      self.body = body
+    }
+    
+    convenience init?(from userInfo: [AnyHashable : Any]) {
+        if let aps = userInfo["aps"] as? NSDictionary {
```
lukasz-szyszkowski, line 15:
```suggestion
        if let aps = userInfo["aps"] as? [AnyHashable: Any] {
```

ios/Classes/PushNotificationEventHandlers.swift
ben-xD: This logic has actually been removed since I am now I am creating a notification regardless of the content. However I think you misunderstand the code there. I am returning the notification if the title OR body is not nil (there is _some_ data).

lukasz-szyszkowski: for me: 
if title is nil OR body is nil (==) -> you return notification object
if title is NOT nil -> you return the notification object

if you want to have "some data" the second condition should look like: `notification.body != nil`, it's equal to `title OR body` is not nil 

This is how I understand this

ben-xD: Ohhhh. I did not notice the `==`, which should actually be `!=`. Thankfully, this code is removed 😂

ios/Classes/Notification.swift
```diff
@@ -0,0 +1,29 @@
...
+                self.init(title: title, body: body)
+            } else if let title = aps["alert"] as? String {
+                self.init(title: title, body: nil)
+            } else {
+                return nil;
+            }
+        } else {
+            return nil;
+        }
+    }
```
lukasz-szyszkowski, line 28: it's just a cosmetic change, IMO it looks better than another if. You can use it or discard it if you want - it's not a blocker 
```suggestion
    convenience init?(from userInfo: [AnyHashable : Any]) {
        guard let aps = userInfo["aps"] as? [AnyHashable: Any] else {
            return nil
        }
        
        if let alert = aps["alert"] as? [AnyHashable: Any] {
            let title = alert["title"] as? String
            let body = alert["body"] as? String
            self.init(title: title, body: body)
        } else if let title = aps["alert"] as? String {
            self.init(title: title, body: nil)
        } else {
            return nil
        }
    }
```

lukasz-szyszkowski (approved):

ikbalkaya (approved):

QuintinWillison (approved):

ben-xD:
Oops, I pressed merge and assumed Github would protect me from merging when the tests are still running.

Issue #31, opened 2020-08-05
ably-java authCallback flow expects synchronous response


tiholic:
According to [ably-java source](https://github.com/ably/ably-java/blob/main/lib/src/main/java/io/ably/lib/rest/Auth.java#L581), authCallback expects a response as a `String` or `TokenDetails` or `TokenRequest`.

Java code can call MethodChannel api to flutter side, but only asynchronously using a `MethodChannel.Response` callback. See: `io.flutter.plugin.common.MethodChannel` in https://api.flutter.dev/javadoc/

tiholic:
@paddybyers 's [comments](https://ably-real-time.slack.com/archives/C0178QGC093/p1596607401047000) on this:

Suppose a Flutter Rest client is initialised with an `authCallback`:

A **REST** operation (or any synchronous operation) should be attempted, and if it fails with a token error (ie an error with a `code` in the range `40140..40159`) then:
- call the `authCallback` for a new token;
- call a new API we're going to have to expose on each library to set the new token;
- retry the request according to the retry policy.

Later we can try to optimise this by having the library check the validity of the token before attempting an operation, but we will need to have retry flow in any case, so let's implement that.


For **Realtime** operations, then a connection can enter the disconnected state and the Reason will indicate whether or not it is a token error. Again, if it is a token error, the client needs to call the callback to get a new token, set the token, and reconnect.

The only use-case that isn't handled by those cases above ^ is in-bound reauthentication: this is where a message is received from the server saying that the token is about to be invalidated, and the library will attempt to renew it before the connection is dropped. We will need a new event to come back from the plugin to the flutter side to indicate that a new token is needed. Again, the callback is called to get a new token, and then that token is set on the plugin.

tiholic:
@paddybyers How about we stop executing further requests once the error with `code` in range `40140..40159` is thrown, queue them and resolve soon after authentication is restored? This could help us avoid multiple auth failures in the time span user code completes authentication.

tiholic:
@paddybyers 

I was able to successfully have a work around for authCallback from Java side. Tried only on Rest so far:

**ably-java**: triggers authCallback
**flutter-java**: initiates a call (**AC1**) to dart side and returns null
**flutter-dart**: issues a new token data in the form of `TokenRequest` or `TokenDetails` or `token` String
**flutter-java**: stores that token data after **AC1** is successful
**ably-java**: triggers authCallback on next message
[...]

Issue #230, opened 2021-11-17
Replace optional fields with default values


ben-xD:
In [ClientOptions](https://github.com/ably/ably-flutter/blob/main/lib/src/authentication/src/client_options.dart), we have some optional fields which should actually have a default value.

For example:
```dart
  /// When true will automatically connect to Ably when library is instanced.
  ///
  /// This is true by default. If false, will wait for an explicit
  /// [ConnectionInterface]#connect to be called before connecting
  /// https://docs.ably.com/client-lib-development-guide/features/#RTC1b
  bool? autoConnect;
```

If `autoConnect` has a default value ("This is true by default"), lets give it one. This means users won't need to handle the case where `autoConnect` is null`, since **it has a default value**. Values should be non-null by default (NNBD). 

It should be:
```
  bool autoConnect = true;
```

This can be repeated for many other fields in `ClientOptions`.

As I do this, I wanted to rename the fields to be clearer: `disconnectedRetryTimeout` changes to `disconnectedRetryTimeoutInMilliseconds`. The comments can then be reduced. For example:
```
  /// When the connection enters the [ConnectionState.suspended] state,
  /// after this delay in milliseconds, if the state is still
  /// [ConnectionState.suspended], the client library will
  /// attempt to reconnect automatically
  ///
  /// default 30,000 (30 seconds)
  /// https://docs.ably.com/client-lib-development-guide/features/#TO3l2
  int? suspendedRetryTimeout;
```
Can be reduced to:
```
  /// When the connection enters the [ConnectionState.suspended] state,
  /// after this delay in milliseconds, if the state is still
  /// [ConnectionState.suspended], the client library will
  /// attempt to reconnect automatically
  ///
  /// https://docs.ably.com/client-lib-development-guide/features/#TO3l2
  int suspendedRetryTimeoutInMilliseconds = 30000;
```

Unfortunately, these fields are in the feature spec, and I don't want to deviate from the feature spec too much.

PR #228, opened 2021-11-16
Improve android logging: Remove unnecessary print statements, and use Log.w where necessary

ben-xD:

PR #231, opened 2021-11-17
Simplify authCallback code

ben-xD:
There was a **lot of complexity** related to supporting `authCallback`, which took me a while to wrap my head around. Once I understood what was going on, I realised it was unnecessary. Here, I remove it all. 😅. It's also really nice to remove the uses of `unawaited`, `fakeAsync`, and our own copy of `unawaited` called `unawaitedWorkaroundForDartPre214` introduced in https://github.com/ably/ably-flutter/pull/198.

There was a lot of discussion over slack and github in 2020 regarding this, and I believe it was a misunderstanding of what the problem was. https://github.com/ably/ably-flutter/issues/31 If you want to understand what the old code does, I would recommend starting at:
```
    final callbackResponse = await realtime.options.authCallback!(tokenParams);
    Future.delayed(Duration.zero, realtime.authUpdateComplete);
```
It **schedules** some work to be done immediately instead of calling the method `authUpdateComplete` directly, inside the implementation of `onRealtimeAuthCallback` method. This eventually calls `connect` on the Platform side, which results in `authUpdateComplete` being called again on the Dart side. **Recursion across platforms**. By this time, **the realtime client handle is updated in the dart side** (see the real problem to know what this means).

## The real problem:

When a realtime client is made, `autoConnect` is set to true by default. This means before the Dart side can get the integer representation of the realtime client (`'handle'`), the `authCallback` is called on the Java side, which calls the dart side. Unfortunately, the dart side doesn't have the client handle yet. This means it can't get the realtime instance to get the authCallback set by the user. 

## Solution

I now disable `autoConnect` in Ably Java, and _emulate_ / _fake_ the autoConnect by manually connecting immediately once the realtime client is returned to the dart side. This difference is not visible to the user, but allows us to reduce a lot of code.

## Other fixes
These are all relevant to the problem above, `authCallback` :)

I also fix a crash which happens when `authCallback` throws an exception in the dart side:
```java
                @Override
                public void error(String errorCode, @Nullable String errorMessage, @Nullable Object errorDetails) {
                  System.out.println(errorDetails);
                  if (errorMessage != null) {
                    result.error("40000", String.format("Error from authCallback: %s", errorMessage), errorDetails);
                  }
```
This crashes because `result` shouldn't be used here, its a result for a different platform method call, so it crashes with an error about "reply already sent". So I just log the error instead. The user already has a chance to handle in the Dart side, where it matters.

Another bug is in:
```
Map<int?, Rest> get restInstances =>
    _restInstancesUnmodifiableView ??= UnmodifiableMapView(_restInstances);
```
When `restInstances` is first called, this will get the correct unmodifiable representation of `_restInstances`. But then if restInstances is changed, `_restInstancesUnmodifiableView` doesn't change with it. I was just simplifying the code when I realized the code was actually buggy! Reducing lines of code helps cut bugs :). TLDR: Unnecessary and wrong caching.

## Some random information that doesn't matter:

I had previously fixed an exception which was happening **every time** authCallback was called in https://github.com/ably/ably-flutter/pull/164. That work using `latch` has allowed me to do this work.

This took me literally the whole day to investigate. I am slightly frustrated at the complexity that was thrown to fight this small misunderstanding. But I am delighted to have finally solved it 😃. There are unfortunately many instances of this pattern of complexity in Ably Flutter.


lib/src/platform/src/realtime/realtime.dart
```diff
@@ -37,6 +36,16 @@ class Realtime extends PlatformObject {
       AblyMessage(options),
     );
     _realtimeInstances[handle] = this;
+
+    if (io.Platform.isAndroid && options.autoConnect != false) {
```
ben-xD, line 40: I don't like `options.autoConnect != false)`, but didn't want to add more complexity to this PR. This is done because `options.autoConnect` is nullable, it shouldn't be. Then it will be non null, we can just do `if (io.Platform.isAndroid && options.autoConnect) {`. See https://github.com/ably/ably-flutter/issues/230

Gonna leave this conversation here indefinitely without resolving it so reviewers understand it is temporary.

ben-xD:
Putting this back in draft because:
- I found more code that can be deleted
- Tests are failing because they are testing that things time out: e.g. "timingOutOnceThenSucceedsAuthCallback". Ideally, `authCallback` doesn't time out as a standard feature of Ably Flutter...

lukasz-szyszkowski (approved):

ikbalkaya (approved):
Light review warning - Looks good

README.md
```diff
@@ -133,21 +133,46 @@ Authenticating using [token authentication](https://ably.com/documentation/core-
...
+      print("Something went wrong in the authCallback: ${e}");
+      rethrow;
     }
   };
 this._ablyClient = new ably.Realtime(options: clientOptions);
 await this._ablyClient.connect();
 ```
 
+An example of `createTokenRequest`'s implementation making a network request to your server could be:
+```dart
```
QuintinWillison, line 153: Before _and after_ code fenced segments it is desirable to include a blank line (vertical white space) to enhance readability when viewing the markdown raw or editing it. That also conforms with other instances in this file and elsewhere across our repositories.

android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java
```diff
@@ -147,6 +151,7 @@ private void handleAblyException(@NonNull MethodChannel.Result result, @NonNull
   @Override
   public void onMethodCall(@NonNull MethodCall call, @NonNull MethodChannel.Result rawResult) {
     final MethodChannel.Result result = new MethodResultWrapper(rawResult);
+    System.out.println("Ably Flutter platform method called: " + call.method);
```
QuintinWillison, line 154: Should this not go via `Log`?

QuintinWillison (dismissed):
I've only done a very light review, however a couple of small things shone out at me.

android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java
ben-xD: Oh yes, thanks, done in b0c3b535687023cde55f5de889956b3975b7e97a
This should now look like:
```
I/System.out(13771): (VERBOSE): io.ably.flutter.plugin.push.PushMessagingEventHandlers: onMethodCall: Ably Flutter platform method "setRestChannelOptions" invoked.
```

I had removed this line before https://github.com/ably/ably-flutter/pull/228 and re-added it after.

README.md
ben-xD: Thanks, I've actually updated this section in a different PR (https://github.com/ably/ably-flutter/pull/250). To avoid merge conflicts, I added the whitespace in 4b1a922e9bc97af9e1dc81ea5b7f1d1da1c4c651

PR #233, opened 2021-11-18
Add default values to ClientOptions

ben-xD:
It is nice to avoid optionals, and to be NNBD: non null by default. In this PR, I set the default values and make fields non-optional. 

Closes https://github.com/ably/ably-flutter/issues/230

### DIFF
--- a/lib/src/platform/src/realtime/realtime.dart
+++ b/lib/src/platform/src/realtime/realtime.dart
@@ -37,7 +37,7 @@ class Realtime extends PlatformObject {
     );
     _realtimeInstances[handle] = this;
 
-    if (io.Platform.isAndroid && !options.autoConnect) {
+    if (io.Platform.isAndroid && options.autoConnect) {
       // On Android, clientOptions.autoConnect is set to `false` to prevent
       // the authCallback being called before we get the realtime handle.
       // If this happens, we won't be able to identify which realtime client
--- a/lib/src/platform/src/realtime/realtime.dart
+++ b/lib/src/platform/src/realtime/realtime.dart
@@ -37,7 +37,7 @@ class Realtime extends PlatformObject {
     );
     _realtimeInstances[handle] = this;
 
-    if (io.Platform.isAndroid && options.autoConnect != false) {
+    if (io.Platform.isAndroid && !options.autoConnect) {
       // On Android, clientOptions.autoConnect is set to `false` to prevent
       // the authCallback being called before we get the realtime handle.
       // If this happens, we won't be able to identify which realtime client
--- a/lib/src/authentication/src/client_options.dart
+++ b/lib/src/authentication/src/client_options.dart
@@ -4,7 +4,7 @@ import 'package:ably_flutter/ably_flutter.dart';
 ///
 /// https://docs.ably.com/client-lib-development-guide/features/#TO1
 class ClientOptions extends AuthOptions {
-  /// initializes [ClientOptions] with log level set to info
+  /// Set fields on [ClientOptions] to configure it.
   ClientOptions();
 
   /// initializes [ClientOptions] with a key and log level set to info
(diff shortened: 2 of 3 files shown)


lib/src/platform/src/realtime/realtime.dart
```diff
@@ -37,7 +37,7 @@ class Realtime extends PlatformObject {
     );
     _realtimeInstances[handle] = this;
 
-    if (io.Platform.isAndroid && options.autoConnect != false) {
+    if (io.Platform.isAndroid && !options.autoConnect) {
```
lukasz-szyszkowski, line 40: it was `!= false` aka `true` now it's negation to `bool` which (in most cases) is `false` - is this intended?

ben-xD: Thanks, fixed in 54fcf8e89d5e1504c9d9996bf4879f600b2b08a8

ben-xD:
Moved back to draft because unit tests are failing. I suspect some of these options are being ignored by the platform side in Ably Flutter 😢.

lukasz-szyszkowski:
> Moved back to draft because unit tests are failing. I suspect some of these options are being ignored by the platform side in Ably Flutter 😢.

```
type 'Null' is not a subtype of type 'int' in type cast
```
I'm guessing you're expecting `Null` somewhere but this value is not optional anymore.

ben-xD:
> > Moved back to draft because unit tests are failing. I suspect some of these options are being ignored by the platform side in Ably Flutter 😢.
> 
> ```
> type 'Null' is not a subtype of type 'int' in type cast
> ```
> 
> I'm guessing you're expecting `Null` somewhere but this value is not optional anymore.

It actually happened during deserialization, (I was expecting non null, but it was null). If it is null, I just don't set it. 30f9275

ikbalkaya (approved):

lukasz-szyszkowski (approved):

QuintinWillison (commented):
It's been quite some time since these interfaces were incepted, probably by @tiholic, but I seem to recall we discussed at that time that we should not be explicitly hard coding these values into the Flutter plugin because that means that they are effectively _always_ overriding the implicit defaults in the `ably-java` and `ably-cocoa` implementations... i.e. if the user's application Dart code doesn't explicitly override client options then the defaults for the underlying 'native' SDK should be used.

If we're deviating from that then I would expect this to need to be documented clearly. Also, from a reviewers perspective, this PR is quite difficult to cross-check as you've introduced a load of new values (typically integers) that are opaque / magic without a load of research to double check.

ben-xD:
> we should not be explicitly hard coding these values into the Flutter plugin because that means that they are effectively always overriding the implicit defaults
I think

> If we're deviating from that then I would expect this to need to be documented clearly. Also, from a reviewers perspective, this PR is quite difficult to cross-check as you've introduced a load of new values (typically integers) that are opaque / magic without a load of research to double check.

They're not magic per se, they are actually in the feature spec, but more importantly, they're in the documentation comments above each method.

ben-xD:
> we should not be explicitly hard coding these values into the Flutter plugin because that means that they are effectively always overriding the implicit defaults

The benefit of overriding it in the Dart side:
- is that the options are guaranteed (to an extent) to be consistent between iOS and Android.
- Users can see the default value by checking `clientOptions`. 
- We get to remove optional types, when they should not be optional, they have **default values**. The fact we are using dependencies which set those defaults should be an implementation detail. Logic is simplified.

The disadvantage of overriding it in the Dart side: 
- The dependencies won't know if the user had explicitly set the dependency or not. IMHO, this is not a disadvantage. Ably Flutter is the user of Ably Cocoa and Ably Java, not the end customer. 

I can't see any true disadvantages here.

QuintinWillison:
@ben-xD Thanks for making such [a clear and succinct argument in response](https://github.com/ably/ably-flutter/pull/233#issuecomment-977905622). I'm convinced. 😄

QuintinWillison (approved):

PR #235, opened 2021-11-18
Platform encapsulation

ben-xD:
@ikbalkaya started a conversation about the generous use of statics in `platform.dart`. https://github.com/ably/ably-flutter/pull/219#discussion_r752149420
> Sorry to be picky on static methods but are we not performing this methods on Platform object here? Using statics will defeat the purpose of encapsulation and will cause the design to consume other static states which makes application unmaintainable. I think those static methods, fields should be used a bit more cautiously

This PR makes `Platform` a singleton instead of having every method and field a static one, and removes direct usages of `Platform.methodChannel` and `Platform.streamsChannel` outside of Platform class (encapsulation).

It also removes any complexity which was added to handle hot restart issues. This was entirely due to using static methods. Now that we have a constructor in `Platform.dart`, hot restarts works for free: the constructor is called at launch and whenever hot restart occurs. I owe this discovery to @ikbalkaya, thanks 👍 . Keep asking good (and difficult) questions 😃

Also includes small log line improvements which I noticed could be improved

## Note about testing

I added optional arguments to the. Platform singleton for tests to override the method channel. This allows them to configure method channel (call `setMockMethodCallHandler`) without manually accessing the internal state of Platform after it is instantiated.

Hot reload and restart are not tricky in themselves. The complexity in the code arose from a misunderstanding of Flutter, and a misuse of static.
- When hot reload happens, no code is executed, so we don't have to worry. We don't need to do anything.
- When hot restart happens, we want to reset the client instances the Dart side loses references (handles) to the ably clients on the platform side. There was a lot of logic to check if a hot restart happened, and was only necessary when using static methods. If we have a singleton + constructor, this is called everytime the app hot restarts or restarts. We can just call the platform method to reset the Ably instances. I called this `resetAblyClients`.


lib/src/platform/src/platform.dart
```diff
@@ -34,8 +34,18 @@ class Platform {
...
+    return await _methodChannel!.invokeMethod<T>(method, arguments);
+    } on PlatformException catch (platformException) {
+    // Convert some PlatformExceptions into AblyException
+      if (platformException.details is ErrorInfo) {
+        throw AblyException.fromPlatformException(platformException);
+      } else {
+        rethrow;
+      }
+    }
+  }
```
lukasz-szyszkowski, line 47: please fix the indentation

ben-xD: It's autoformatted, if i change it I will fail the format check CI 🙈.

lukasz-szyszkowski (approved):

ikbalkaya (approved):
Looks good to me 😎  Thanks for addressing this

PR #237, opened 2021-11-18
Hide API key secret inside API key UI

ben-xD:
In this UI enhancement, I hide the API secret key from the API key string. This allows me to show the example app to more people without worrying that I am leaking API keys. This should allow other users to share screenshots more safely too:
CleanShot 2021-11-18 at 23 31 25@2x

I also replaced the Text widget with I previously created called `TextRow`, so it now the labels are in blue:
CleanShot 2021-11-18 at 23 41 27@2x

This is the same widget used when receiving push messages or encrypted messages:
CleanShot 2021-11-18 at 23 43 49@2x

### DIFF
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -674,7 +674,7 @@ class _MyAppState extends State<MyApp> {
                           ]),
                     )
                   else
-                    Text('API key: ${redactApiKey(_apiKey)}'),
+                    Text('API key: ${hideApiKeySecret(_apiKey)}'),
                   const Divider(),
                   const Text(
                     'Realtime',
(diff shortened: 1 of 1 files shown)


lukasz-szyszkowski:
If you plan to share this apk/ipa with others, please consider the code obfuscation or preventing code from an easy read, e.g. build type `release` instead of `debug` etc

lukasz-szyszkowski (approved):

ben-xD:
> If you plan to share this apk/ipa with others, please consider the code obfuscation or preventing code from an easy read, e.g. build type `release` instead of `debug` etc

👍 I don't plan to share this APK/IPA. It's just for screenshots. If sharing with others, we should use token authentication instead. You can see how to implement this in Ably Flutter in https://github.com/ably/ably-flutter/#configure-a-client-options-object.

QuintinWillison (approved):
Nice.

PR #240, opened 2021-11-19
Example app: Move out Realtime and Rest UI to separate widget and service

ben-xD:
I move out code related to Realtime and rest clients in the example app into separate widgets and services for a better separation of concerns.

Mainly, it takes a big step in readability of the example app: the `main.dart` file goes from 600+ lines to 55. This is really important for users. If they are interested in a specific functionality, they can jump to that widget immediately, because the app entry point now looks like this:

https://github.com/ably/ably-flutter/blob/1d056aa198001bdbf89c7c11f3b8627a219d6573/example/lib/main.dart#L24-L49


lukasz-szyszkowski (approved):

QuintinWillison (approved):
Very difficult to review with complete certainty due to the amount of code that's moved, but looks sensible to me (albeit a very light review).

PR #242, opened 2021-11-19
Example app: Paginated results viewer widget

ben-xD:
This PR requires you to know what a paginated result is. Take a look at the [docs](https://ably.com/documentation/realtime/types#paginated-result) for more information. In summary: because some results can be quite long, we break them up into pages (pagination), which need to be requested sequentially.

The example app does not explain `PaginatedResult` very well, and is confusing to use (and contains bugs). See 2 minute recording at bottom of page.

In this improvement, the `PaginatedResultViewer` is a widget that includes a refresh button (to fetch the latest data), clear the page (delete button) the page number, and 2 navigation buttons (next page and go to first page). I wish there was a way to get the **previous page**. I now allow users to browse through the pages of a paginated result.

Please take a look at the GIF below to see a quick demo:

CleanShot 2021-11-19 at 18 53 21

With this change, the sliver (scroll view contents) gets longer and longer. I plan to separate each sliver (realtime, rest, push notifications) into a separate screen that the user can navigate to. Will be done in a separate PR.

## Other small fixes
I also replace the deprecated usage of FlatButton with TextButton. I also remove any manual customization of button colors. This is difficult to maintain and untidy. Theming of the example app can be done in a future PR if desired. 

If you'd like to know the difference between the original example app and mine, please watch this 2 minute video **with audio explanation**. Left is the previous example app, and right is the output of this PR.

https://user-images.githubusercontent.com/24711048/142721008-7209ad5e-44c4-464b-a444-55c881198e84.mp4

I did reveal the API key in the video (because it uses the old UI), but this UI has been improved in a separate PR, and the API key is deleted.


lukasz-szyszkowski (approved):

lukasz-szyszkowski (approved):

example/lib/main.dart
```diff
@@ -189,83 +186,49 @@ class _MyAppState extends State<MyApp> {
     _subscriptionsToDispose.add(_channelStateChangeSubscription);
   }
 
-  // https://github.com/dart-lang/sdk/issues/37498
-  // ignore: missing_return
-  static String opStateDescription(
```
QuintinWillison, line 194: This method and the one below presumably stopped being used in a previous bit of work but you're cleaning up now? Just to be clear I understand why these are disappearing - and, presumably, that ideally could have been done under a discrete PR?

QuintinWillison (approved):
Lots of change in this PR, in particular in `example/lib/main.dart`, so it's a bit difficult to review without spending a lot of time to go in depth (which I've not got right now). So, at surface level, looks sensible but my review was very light.

example/lib/main.dart
ben-xD: Yes, exactly correct.

It was removed in https://github.com/ably/ably-flutter/pull/202 and was some code that I didn't remove.

PR #243, opened 2021-11-20
Replace deprecated method calls in tests

ben-xD:
Following https://docs.flutter.dev/release/breaking-changes/mock-platform-channels to remove deprecated method calls.

Specifically: 
> Code that does not have access to a WidgetTester can refer to TestDefaultBinaryMessengerBinding.instance!.defaultBinaryMessenger instead of tester.binding.defaultBinaryMessenger.


lukasz-szyszkowski (approved):

ikbalkaya (approved):

QuintinWillison:
I know we've had the Android 29 test run fail, but I think it's clear that failure doesn't relate to the contents of the PR so I'm still going to land it.

PR #244, opened 2021-11-20
Serialize LogLevel using constants

ben-xD:
There was a TODO comment left in the code: 

```
/// TODO(tiholic) convert [LogLevel] to enum and update encoder to pass
///  right numeric values to platform methods
class LogLevel {
```

This PR completes that TODO: Instead of using serializing logLevels to numbers, we use Strings, just like other enums in the codebase.


lukasz-szyszkowski (approved):

ios/Classes/codec/AblyFlutterReader.m
```diff
@@ -5,15 +5,13 @@
...
-static ARTLogLevel _logLevel(NSNumber *const number) {
-    switch (number.unsignedIntegerValue) {
-        case 99: return ARTLogLevelNone;
-        case 2: return ARTLogLevelVerbose;
-        case 3: return ARTLogLevelDebug;
-        case 4: return ARTLogLevelInfo;
-        case 5: return ARTLogLevelWarn;
-        case 6: return ARTLogLevelError;
-    }
+static ARTLogLevel _logLevel(NSString *const logLevelString) {
```
ikbalkaya, line 8: Just a comment on this that doesn't need a code change and is just a recommendation for the future reference. it's common to use a dictionary in cases where keys / values are known at compile time. In such methods you would usually query this dictionary for the existence of a value and dealt with the case when the value does not exist.
Something like this:
```
if(logLevel[logLevelString]){
      return logLevel[logLevelString];
}
return ARTLogLevelWarn;
```

lib/src/generated/platform_constants.dart
```diff
@@ -256,6 +257,15 @@ class TxFormFactorEnum {
   static const String other = 'other';
 }
 
+class TxLogLevelEnum {
```
ikbalkaya, line 260: The name ends with 'Enum' but it's a class 🤔 Is there a reason why you can't use enums in this case? If so I would recommend you change the name of this class.

ikbalkaya (changes requested):

lib/src/generated/platform_constants.dart
ben-xD: This is not an enum. Anything in `platform_constants.dart` are constants used for serialization between Dart and Platform side. This is the platform constants representing the LogLevel enum cases in JSON.

It might be clear enough to just call it `TxLogLevel` though. What do you think?

ios/Classes/codec/AblyFlutterReader.m
ben-xD: You mentioned this previously and I have considered it.

- Using a map might be fewer characters, but it breaks up the code. Your code snippet has to check for existence in the map (developers have to check the map to double check when coding, and this is mental overhead). This an additional layer of indirection. What the code really is doing is a more implicit/hidden if-else anyway, it's clearer to see it happen IMHO. Also,
 - I could put in the method, at which point we are recreating the map on every method invocation. This might be bad (maybe not).
 - If I put the map outside the method, it is more mental overhead for a developer to jump to that map.
- It doesn't provide any additional checks or protection. The compiler warns us about the return types in both cases.
- It is not flexible. What if one day we want to deserialise a nested value. Using a map actually makes the code more rigid, and we are unable to call methods. In the enum case, this doesn't matter, but then we have 2 patterns in the codebase, as opposed to the shorter syntax I am using.

In the near future, I would like to rewrite this file in Swift, and perhaps split it up into separate files. At that time, I will implement both ways to have a better appreciation.

lib/src/generated/platform_constants.dart
ben-xD: I will write some instructions about how serialization works in Ably Flutter

QuintinWillison: I understand from @ben-xD that he will write those instructions in a subsequent PR (see https://github.com/ably/ably-flutter/issues/263), therefore I am resolving this conversation on @ikbalkaya's behalf.

Also, given this comment was on a `generated/` file, I think this only strengthen's my resolve to encourage us to move to a situation where this repository _just_ contains source files. Generated files should be present ephemerally, only at build time. I've created https://github.com/ably/ably-flutter/issues/264 to track this.

ios/Classes/codec/AblyFlutterReader.m
QuintinWillison: I am resolving this conversation on @ikbalkaya's behalf.

QuintinWillison (approved):

PR #249, opened 2021-11-22
Run dispose implementation (clearing rest/realtime client instances) immediately, not with `dispatch_async`

ben-xD:
I'm moving this code from running in the background, to instead run synchronously. 

This fixes a race condition bug where dispose is first called from the Dart side to clear all the instances of realtime and rest. Unfortunately the actual implementation of `dispose` might only happen afterwards, because of `dispatch_async`. By that time, a user might have created a new realtime instance. Then, this dispose method finally gets called, and their client is deleted, almost immediately after they create it.

This was actually discovered because I **removed** the button `create realtime`, and created the realtime client on launch instead. User applications would likely create a realtime instance either on application launch or user log in too.

### DIFF
--- a/ios/Classes/AblyFlutter.m
+++ b/ios/Classes/AblyFlutter.m
@@ -131,17 +131,6 @@
         [NSException raise:NSInvalidArgumentException format:@"completionHandler cannot be nil."];
     }
     
-    dispatch_assert_queue(dispatch_get_main_queue());
-
-    // This is contrived for now but the point is that we can introduce a clean,
-    // asynchronous close via a background queue here if required.
-    dispatch_async(dispatch_get_main_queue(), ^{
-        [self dispose];
-        completionHandler();
-    });
-}
-
--(void)dispose {
     for (ARTRealtime *const r in _realtimeInstances.allValues) {
         [r close];
     }


lukasz-szyszkowski (approved):

ben-xD:
Here is an explanation of the code:

When a users application (or example app) launches for the first time, or a hot restart happens (Flutter app restarts without restarting iOS host application), the Dart side calls a method: `registerWithCompletionHandler`. This calls `AblyFlutter disposeWithCompletionHandler`. 

**Bug:** The `registerWithCompletionHandler` method is first called when the app launches, but the clearing does not happen immediately. Then, the user might programmatically create a realtime at app launch too. This saves a realtime client in `AblyFlutter.m`'s `_realtimeInstances`. However, finally, the following code is executed:
```objective-C
    for (ARTRealtime *const r in _realtimeInstances.allValues) {
        [r close];
    }
    [_realtimeInstances removeAllObjects];
    [_restInstances removeAllObjects];
```
deleting the users realtime client.

ios/Classes/AblyFlutter.h
```diff
@@ -25,14 +25,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 -(ARTPaginatedResult *) getPaginatedResult:(NSNumber *const) handle;
 
-/**
- This method must be called from the main dispatch queue. It may only be called
- once.
- 
- @param completionHandler Will be called on the main dispatch queue when all
- platform objects have been closed down cleanly.
- */
```
ben-xD, line 34: I removed this comment because it is wrong. `disposeWithCompletionHandler` could (and **was**) being called more than once. Whenever the app hot restarts (Flutter app restarts but iOS app doesn't), this method was being called.

Then I remove the outdated @param, since i removed it, because it is no longer necessary.

lukasz-szyszkowski (approved):

ios/Classes/AblyFlutterPlugin.m
```diff
@@ -36,7 +36,7 @@ -(void)registerWithCompletionHandler:(FlutterResult)completionHandler;
 };
 
 static const FlutterHandler _resetAblyClients = ^void(AblyFlutterPlugin *const plugin, FlutterMethodCall *const call, const FlutterResult result) {
-    [plugin registerWithCompletionHandler:result];
+    [plugin reset:result];
```
ben-xD, line 39: I renamed this `reset` because it is being called to reset clients.

QuintinWillison (approved):

PR #252, opened 2021-11-23
Remove ON_VALUE usage because `hasAuthCallback` is always set

ben-xD:
Xcode was warning about an unused returned value when we used:
```
[co initWithClientOptions: o hasAuthCallback: value];
```

 I just call the initializer directly and return the result.

Also, there was no need for the use of `ON_VALUE` macro. `dictionary[TxClientOptions_hasAuthCallback]` is always `false` or `true`, never `nil` or `NSNull`.

When you see
```
  _writeToJson(
        jsonMap, TxClientOptions.hasAuthCallback, v.authCallback != null);
```
This is saying, we only set the key/value pair to `jsonMap` if `v.authCallback != null` is non null. Well, `!=` never returns null.


ios/Classes/codec/AblyFlutterReader.m
```diff
@@ -146,12 +146,10 @@ use an explicitly received null from Dart (manifesting this side as a nil id)
...
 
-    AblyFlutterClientOptions *const co = [AblyFlutterClientOptions new];
-    ON_VALUE(^(const id value) {
-        [co initWithClientOptions: o hasAuthCallback: value];
-    }, dictionary, TxClientOptions_hasAuthCallback);
-
-    return co;
+    
+    return  [[AblyFlutterClientOptions alloc]
+             initWithClientOptions:o
```
lukasz-szyszkowski, line 151: can you change `o` for something more readable?

ben-xD: Sure, that has been done in ed3851f23844f049f9285f462eb0096a7a5b845e. 😇

lukasz-szyszkowski (approved):

ben-xD:
@lukasz-szyszkowski, I reverted the rename of variables (e.g. `o` to `clientOptions`) and have put this in a separate PR https://github.com/ably/ably-flutter/pull/262.

PR #259, opened 2021-11-25
SDK Dart platform code improvements

ben-xD:
I do some refactoring here: 

- Remove AblyBase, since we could not 'extend' from it anyway, and was just implementing the interface, which is implicit in Dart. This means the constructor code in AblyBase was ignored, and any future implementations in the abstract class, `AblyBase` would be ignored. It is even confusing because the comment said "base class", but classes simply "implemented" the interface, not "extended" it's implementation. This was a non-idiomatic use of abstract class in Dart.
- Add migration guide, for users to use to update across breaking changes.
- Rename `pushEvents` to `activationEvents` to be clearer
- Rename XNative classes to XInternal, to signify they exist because they are internal to SDK, and not to be externalized.
- Remove unnecessary `PushNative` and `PushChannelNative` classes, we can expose them directly and call them `Push` and `PushChannel` directly.
- Replace complex (and buggy) Rest and Realtime constructor with 2 constructors. Previously, your API key was totally ignored when you called `ably.Rest(key: yourKey, options: yourOptions)`. Even tests were using it like this.


lukasz-szyszkowski (approved):

UPDATING.md
```diff
@@ -0,0 +1,8 @@
+# Updating / Migration Guide
+
+This guide lists the changes needed to upgrade from one version of Ably to a newer one when there are breaking changes.
+
+## [Upgrading from v1.2.6]
+
+- `ably.Push.pushEvents` is renamed to `ably.Push.activationEvents`, to be more meaningful. It provides access to events related to setting up push notification, such as activation, deactivation and notification permission. This was done to help future users clearly distinguish between `activationEvents` and `notificationEvents`.
+- When instantiating `Rest` or `Realtime` with an API key, replace `ably.Rest(key: yourApiKey)` with `ably.Rest.fromKey(yourApiKey)`. This was done because using `ably.Rest(key: yourApiKey, options: clientOptions)` was misleading (`yourApiKey` was ignored).
```
ikbalkaya, line 8: I think you could soften the tone a little here after 'because' 
Even if it was misleading, would clients be able to pass their key using some other medium(eg. ClientOptions)? If so you could say that "it wasn't neccessary because key was already provided through X. We simplified the API to enable you to use the Rest only using key..."

lib/src/platform/src/realtime/realtime.dart
```diff
@@ -54,58 +56,74 @@ class Realtime extends PlatformObject {
...
-  @override
   Connection get connection => _connection;
 
-  @override
-  Auth? auth;
+  /// a custom auth object to perform authentication related operations
+  /// based on the [options]
+  ///
+  /// https://docs.ably.com/client-lib-development-guide/features/#RSC5
+  // Auth? auth;
```
ikbalkaya, line 65: I think this comment is no longer needed

lib/src/platform/src/realtime/realtime.dart
```diff
@@ -54,58 +56,74 @@ class Realtime extends PlatformObject {
...
 
-  @override
-  ClientOptions options;
+  /// [ClientOptions] indicating authentication and other settings for this
+  /// instance to interact with ably service
+  late ClientOptions options;
 
-  @override
+  /// a push object interacting with Push API
+  /// viz., subscribing for push notifications, etc
```
ikbalkaya, line 72: I think those comments belong to the class implementation itself and not needed here. (It's probably there already) Some for other dependencies in this class

lib/src/platform/src/realtime/realtime.dart
```diff
@@ -54,58 +56,74 @@ class Realtime extends PlatformObject {
...
+  // }) {
+  //   throw UnimplementedError();
+  // }
+
+  /// gets stats based on params as a [PaginatedResult]
+  ///
+  /// https://docs.ably.com/client-lib-development-guide/features/#RSC6
+  // Future<PaginatedResult<Stats>> stats([Map<String, dynamic>? params]) {
+  //   throw UnimplementedError();
+  // }
```
ikbalkaya, line 108: Commented out code 🤔

lib/src/platform/src/realtime/realtime.dart
```diff
@@ -54,58 +56,74 @@ class Realtime extends PlatformObject {
...
-  @override
-  Future<DateTime> time() {
-    throw UnimplementedError();
-  }
+  /// returns server time
+  ///
+  /// https://docs.ably.com/client-lib-development-guide/features/#RSC16
+  // Future<DateTime> time() {
+  //   throw UnimplementedError();
+  // }
```
ikbalkaya, line 115: Commented out code

lib/src/platform/src/rest/rest.dart
```diff
@@ -36,42 +32,56 @@ class Rest extends PlatformObject implements AblyBase {
...
+  /// https://docs.ably.com/client-lib-development-guide/features/#RSC19
+  // Future<HttpPaginatedResponse> request({
+  //   String? method,
+  //   String? path,
+  //   Map<String, dynamic>? params,
+  //   Object? body,
+  //   Map<String, String>? headers,
+  // }) {
+  //   throw UnimplementedError();
+  // }
```
ikbalkaya, line 56: Commented out code

lib/src/platform/src/rest/rest.dart
```diff
@@ -36,42 +32,56 @@ class Rest extends PlatformObject implements AblyBase {
...
-  @override
-  Future<PaginatedResult<Stats>> stats([Map<String, dynamic>? params]) {
-    throw UnimplementedError();
-  }
+  /// gets stats based on params as a [PaginatedResult]
+  ///
+  /// https://docs.ably.com/client-lib-development-guide/features/#RSC6
+  // Future<PaginatedResult<Stats>> stats([Map<String, dynamic>? params]) {
+  //   throw UnimplementedError();
+  // }
```
ikbalkaya, line 63: ++

lib/src/platform/src/rest/rest.dart
```diff
@@ -36,42 +32,56 @@ class Rest extends PlatformObject implements AblyBase {
...
+  /// returns server time
+  ///
+  /// https://docs.ably.com/client-lib-development-guide/features/#RSC16
+  // Future<DateTime> time() {
+  //   throw UnimplementedError();
+  // }
 
-  @override
+  /// a push object interacting with Push API, such as
+  /// subscribing for push notifications by clientId.
```
ikbalkaya, line 73: Similar situation here. I'd leave the documentation , comment to the class implementation

lib/src/push_notifications/src/push_channel.dart
```diff
@@ -1,28 +1,58 @@
...
+class PushChannel extends PlatformObject {
+  final String _name;
+
+  /// A rest client used platform side to invoke push notification methods
+  final Rest? rest;
+
+  /// A realtime client used platform side to invoke push notification methods
+  final Realtime? realtime;
+
+  /// Pass the channel name and an Ably realtime or rest client.
```
ikbalkaya, line 16: I do not think this comment is needed

ikbalkaya (changes requested):
Code looks good 😎
 My requests are mainly about comments and documentation 
* Lots of commented out code. 
* There is a pattern where the objects are documented on their declaration site. I think documentation should be and be part of their implementaion. An exception would be when you have to explain 'why' you added that object as the dependency.

lib/src/platform/src/rest/rest.dart
ben-xD: This documentation is for the SDK user, and this type (Rest) are being exposed to the user. The implementation is not exposed to the user, so they won't see the documentation.

UPDATING.md
ben-xD: I guess in that line, I am explaining the breaking change and why I **had** to make the change. A breaking change should not just be done to improve it slightly (simplifying the API), but for a very good reason: it was misleading users (if they passed an API key in some cases, it is ignored). So IMHO in this case the tone is important, to inform users that we **had** to make the change, not just for a small improvement.

ben-xD: Alternatively, I could say "to fix a bug in API key being ignored". What do you think?

ikbalkaya (approved):

PR #262, opened 2021-11-26
Rename variables to be more meaningful

ben-xD:
I had previously had these renames in a different PR (https://github.com/ably/ably-flutter/pull/252), but reverted the commit which renamed this variables because it was taking too much attention. 

This PR only renames some variables in objective-C, and replaces dictionary access to the square bracket syntax.


lukasz-szyszkowski (approved):

ikbalkaya (approved):
LGTM

PR #265, opened 2021-11-29
Example app: "release" channel button

ben-xD:
This PR fixes the "Release" channel button. Previously, if a user released a channel, the state of the channel would no longer be updated on the UI. Now, you can detach or release from the channel, and still attach later.


lukasz-szyszkowski (approved):

ikbalkaya (approved):

QuintinWillison:
I'm a little concerned that this PR conflates a library implementation fix (new call to super in `lib/src/platform/src/realtime/realtime_channels.dart`) with example app fix, but given @ikbalkaya and @lukasz-szyszkowski have approved it then I'll assume this has been explained, or at least understood by someone with more context than I.

Issue #216, opened 2021-11-11
Format iOS and Android code (automate it, and add checks to CI)


ben-xD:
Files are not consistently formatted. 

- We should autoformat all files, not just dart files
- We should find a way to autoformat this (without an IDE, this might be hard for Swift/objective-C)
- We should then add formatting checks to CI

PR #266, opened 2021-11-29
Format Swift and Java code

ben-xD:
I have formatted the Swift and Java code using:
- Swift: `swiftformat --swiftversion 5.4 .`
- Java: `java -jar google-java-format-1.13.0-all-deps.jar --replace **/*.java`.

There will be a separate PR to add either checks or code to automatically format and commit the code in CI (after discussion), and documentation on how to format code.

Using an autoformatter alongside feature changes in PRs would cause the file diff to change so much, that it would be confusing to reviewers. Therefore, I had to previously avoid using autoformatting in Java.

PR #267, opened 2021-11-29
Formatting script, CI checks and documentation improvement

ben-xD:
We previously had files of the same programming language styled differently, and this was fixed in https://github.com/ably/ably-flutter/pull/266. 

This PR details how to format the files, and adds a script to help autoformat the code.

Example output:
```
❓ Checking file formatting dependencies are installed...
✅ File formatting dependencies exist.

❓ Formatting Dart files...
Formatted 185 files (0 changed) in 1.22 seconds.
✅ Formatted Dart files.

❓ Formatting Java files...
✅ Formatted Java files.

❓ Formatting Swift files...
Running SwiftFormat...
SwiftFormat completed in 0.24s.
0/12 files formatted.
✅ Formatted Swift files.
```

It also has a pre-commit to check if files are formatted. It doesn't format the files.

The following is how it looks in SourceTree:

Screenshot 2021-12-04 at 15 47 39

Closes https://github.com/ably/ably-flutter/issues/216
Closes https://github.com/ably/ably-flutter/issues/263
Closes https://github.com/ably/ably-flutter/issues/271


lukasz-szyszkowski (commented):
I wonder if it's possible to use the same shell script `format-files.sh` in `.githooks/pre-commit` ?

ben-xD:
> I wonder if it's possible to use the same shell script `format-files.sh` in `.githooks/pre-commit` ?

Thanks for your comment.

I've moved out the initial checks for dependencies into a separate file and call that instead. See 0689c364a9abdfde0517c27666c1c0b269a3d82f.

I can't find a way to reduce some apparent duplication in the actual formatting tools without adding complexity. The format-files.sh **formats** files, the pre-commit only **checks** the files. The reason why it looks like its duplication is because in both of them, I am using the same tools. However, for each tool, I use different options. (e.g. `--set-exit-if-changed`).

lukasz-szyszkowski (approved):

ikbalkaya (approved):

ben-xD:
@QuintinWillison had comments on Slack regarding this PR:

> So, the essence was:
> - I feel what you've done is too convoluted - it's 'surprising' to me and doesn't feel idiomatic for Java or Flutter, plus is not consistent with what we do anywhere else in SDKs
> - I'm not comfortable with the Git hook folder - I've not researched deeply but the 'dot' folder feels like something that could automatically 'kick in' in a developer's environment without them realising it and actually cause surprise / pain
> - In terms of checkstyle - as I said, we already have a standard elsewhere (albeit underutilised, but it's on my TODO list to improve that) - and, anyway, I would expect lints / checks / analysis to be invoked at the Gradle level (I believe we do have that in the Flutter stack)

My response was:

> It's solving a problem that I am facing: files are not formatted and I can't use auto-formatting in the IDE because it changes the entire file (as it is not formatted at all). At the moment, no one else faces this issue so I can see that context is difficult to acquire.
> 
> To address your three points:
> I don't see any issues I can solve from what you've mentioned, unless I move to something you have declared to be the standard. Google-java-format is a tool which exists and is popular. Apart from me stopping using it, I'm not sure why its unidiomatic. If it's too complicated Im happy to see suggestions or more specific complaints in that PR. 
> The .githooks is not opt-out, its opt-in, as mentioned previously. This is not the .git/hooks folder which automatically kicks in. This is explained in the contribution guide change I made: 
> - To automatically check formatting before committing, you can use the git hook shared in this repo. 
>   - Run git config core.hooksPath .githooks 
> I think I can elaborate further so people understand what this is doing. (e.g. .githooks is just a folder where we can share githooks, and they are optional and opt-in)
> I think there has to be significant reasoning to switch to checkstyle or to claim there is a "standard" (checkstyle) we follow. If you want to spend more time or effort choosing between the 2 so we can choose the better one for this case, sure, but we don't have a standard: we have one (or some) repos which use checkstyle (and not very well at the moment, as I discovered myself previously in https://ably-real-time.slack.com/archives/C019LDWM9UH/p1632408898015500).
> 
> My point in using google-java-format and not a linter is that developers never have to worry about formatting, or configuring the formatter, or discussing formatting. Formatting should be boring, resulting in much lower maintenance in the long run. Your mention of "very few rules yet enabled" when using checkstyle (in message) and previously in conversation that you wanted to gradually add rules over time, and that single rules causes a lot of lines to change was very worrying to me if I had to do the same to Ably Flutter (I want to focus on higher priority work). This would be a bad path to follow on. We just want code that is consistent, not to style the code or configure it in a special way. Or manually edit files to meet the linter. Google-java-format is solving this problem very well.
> 
> As an aside: IMHO this type of discussion that should be in the PR, so that we keep a record of our decisions. #open-for-all. I'm not sure why we bring more difficult conversations and where these decisions get made in Slack.

I have since renamed `.githooks` to `hooks`, so it is not a **hidden** directory, as requested by @QuintinWillison in a video meeting we had.

Overall, using [google-java-format](https://github.com/google/google-java-format) is not unidiomatic to Java (it's got Java in the name) and was specifically designed for Java. I'm also avoiding more work down the line unlike Quintin ("it's on my TODO list to improve that"). There is no work stemming off from this PR, just **problems solved**.

This PR has been waiting 15 days.

lukasz-szyszkowski (approved):

QuintinWillison:
The way this work had been structured and approached differs too greatly with the norms we're establishing elsewhere across the SDK Team at Ably. We will take another run at checking code formatting in CI in the future, but right now there are more important things to be focussing on (including a steady stream of bugs). This PR adds noise as well as too much risk.

PR #269, opened 2021-11-30
Add integration test for API level 31

ben-xD:
Newer versions of Android are available, and we should test the SDK on them. Here, I run integration tests on Android API level 31 (latest) and 24 (anecdotally, this is where some changes were made on Android where I have seen things to suddenly stop working)).

I needed to specify architecture and target to allow Android API level 30 and 31 emulators to work. I also needed to update the command usde (avdmanager) because `/Users/runner/Library/Android/sdk/tools/bin/avdmanager` is no longer available for newer versions of Android SDK.

### DIFF
--- a/.github/workflows/flutter_integration.yaml
+++ b/.github/workflows/flutter_integration.yaml
@@ -36,7 +36,7 @@ jobs:
   android:
     strategy:
       matrix:
-        api-level: [21, 24, 30]
+        api-level: [21, 24, 31]
       fail-fast: false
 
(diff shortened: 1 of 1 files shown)


lukasz-szyszkowski (approved):

ikbalkaya (approved):

QuintinWillison:
This PR changes too much at one time so I am not satisfied that it takes a cautious enough approach. For that reason I am closing it and we'll run at this issue another time.

PR #273, opened 2021-12-01
iOS simulator enhancement for Push activation/deactivation

ben-xD:
Currently, when calling `realtime.push.activate/deactivate` on iOS simulators, the method never returns because Ably Cocoa never completes its callbacks.

This means users had to code around the problem, either by adding a timeout if it were a simulator, or avoiding calling the method at all if it were the simulator. This is annoying and leads to more complex for users.

To fix this, we check for the iOS simulator on the iOS side, and throw and error immediately instead of never completing the future.

# Alternative
We could create a random APNs token if running on iOS simulator, so that subscriptions can still be created. However, this might pollute the data, since these device registrations would be for simulators, and completely useless.

### DIFF
--- a/ios/Classes/handlers/PushHandlers.swift
+++ b/ios/Classes/handlers/PushHandlers.swift
@@ -9,18 +9,28 @@ public class PushHandlers: NSObject {
         if PushActivationEventHandlers.getInstance(methodChannel: ably.channel).flutterResultForActivate != nil {
             returnMethodAlreadyRunningError(result: result, methodName: "activate")
         } else if let push = getPush(instanceStore: ably.instanceStore, call: call, result: result) {
-            PushActivationEventHandlers.getInstance(methodChannel: ably.channel).flutterResultForActivate = result
-            push.activate()
+            #if targetEnvironment(simulator)
+            print("Device activated for push immediately because you application is running in simulator. APNs does not work on a simulator.")
+                result(nil)
+            #else
+                PushActivationEventHandlers.getInstance(methodChannel: ably.channel).flutterResultForActivate = result
+                push.activate()
+            #endif
         }
     }
 
     @objc
     public static let deactivate: FlutterHandler = { ably, call, result in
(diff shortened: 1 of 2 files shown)


ios/Classes/handlers/PushHandlers.swift
```diff
@@ -9,18 +9,29 @@ public class PushHandlers: NSObject {
         if PushActivationEventHandlers.getInstance(methodChannel: ably.channel).flutterResultForActivate != nil {
             returnMethodAlreadyRunningError(result: result, methodName: "activate")
         } else if let push = getPush(instanceStore: ably.instanceStore, call: call, result: result) {
-            PushActivationEventHandlers.getInstance(methodChannel: ably.channel).flutterResultForActivate = result
-            push.activate()
+            #if targetEnvironment(simulator)
+                result(FlutterError(code: "PushHandlers_activate", message: "Device activation for push notifications does not work on iOS simulator because APNs is not supported.", details: nil))
+                return
+            #else
```
lukasz-szyszkowski, line 15: te solution is OK, but  `return` above is redundant

ben-xD: I know it's redundant _now_, but maybe I'm afraid that code gets added below the else that this code will be missed, and then this code . Premature optimization, so i fixed in: c87076b332c81ac369f62bbf750c2ab326f5eb6e

lukasz-szyszkowski: I understand, so you can leave `return` but you should remove `#else` then :D

ben-xD: I've removed all the returns in a new commit `9b89a44eb8d9d04ad80c72f2f41280301be4effd`. After thinking about it, I prefer this way since if/else is clearly on the same indentation level.

lukasz-szyszkowski (approved):

lukasz-szyszkowski (approved):

ikbalkaya (approved):
Looks good 👍

QuintinWillison:
Poorly documented. I can't see what value this adds. Actually, I think it makes debugging harder.

Issue #276, opened 2021-12-06
Investigate push activation not completing on iOS when running in release configuration


ben-xD:
When running the application in release mode, the device fails to complete activation when it first launches. The last event in the state machine is a transition from `NotActivated` to `WaitingForPushDeviceDetails`. A workaround would be to close and relaunch the app. Once relaunched, activation/deactivation works.

```
2021-12-06 15:00:34.685283+0700 Runner[21776:7802732] DEBUG: ARTPushActivationStateMachine: transition: ARTPushActivationStateNotActivated -> ARTPushActivationStateWaitingForPushDeviceDetails
```

To reproduce it:
- Run the example app using either 
  - `flutter run --dart-define=ABLY_API_KEY=$ABLY_API_KEY --release`.
  - Xcode "**release**" configuration **and** debug executable **unchecked** (see screenshot)
![CleanShot 2021-12-06 at 09 58 04](https://user-images.githubusercontent.com/24711048/144826023-3348d069-3482-4c32-9511-98a32d3a17d8.png)
- Attempt to activate the device by scrolling down to the **"Push Notifications"** sliver of the scroll view, and press "Activate"

Other notes:
- This only happens on iOS
- Other platform methods are working (e.g. connect realtime). 

# Problem?
This requires further investigation.

This might it is an Ably Cocoa, specifically an Activation State Machine issue, potentially a race condition which occurs when the app runs faster (e.g. in release mode).

ben-xD:
After further investigation, it looks like it is because the APNs token is not provided to the app early enough in some cases, and the realtime and rest clients are instantiated without it. 

To test this out, use:
```
  ably_flutter:
    git:
      url: git://github.com/ably/ably-flutter.git
      ref: 4685ad628509ca4261c1a283fdc2c90a8fc2cdd0 
```

PR #277, opened 2021-12-06
Send deviceToken to Ably clients which already have been instantiated

ben-xD:
This PR does 2 things:
- It fixes https://github.com/ably/ably-flutter/issues/276
- It allows users to give Ably flutter the device token if they're having issues activating push with Ably. This happens because in some cases (e.g. a customer), `didRegisterForRemoteNotifications` does not get called in Ably Flutter when running on iOS. This might happen due to method swizzling by other dependencies. This is a **workaround** for users who cannot fix the fact `didRegisterForRemoteNotifications` is not being called in Ably Flutter.

### DIFF
--- a/PushNotifications.md
+++ b/PushNotifications.md
@@ -142,7 +142,6 @@ import ably_flutter
     override func application(_ application: UIApplication, didFailToRegisterForRemoteNotificationsWithError error: Error) {
         NSLog("application:didFailToRegisterForRemoteNotificationsWithError was called with error: %@", error.localizedDescription)
         AblyInstanceStore.sharedInstance().didFailToRegisterForRemoteNotificationsWithError_error = error;
-
     }
 }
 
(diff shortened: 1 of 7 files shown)


ben-xD:
@lukasz-szyszkowski , I would like to change the pattern of calling a method in the singleton instance of `AblyInstanceStore`, seen in `AblyInstanceStore.sharedInstance().didRegisterForRemoteNotifications(withDeviceToken: deviceToken)`

lukasz-szyszkowski (approved):

ikbalkaya (approved):
LGTM

QuintinWillison:
We're pressured for time to get a release out. So, I'm going to land this work despite having not observed checks pass yet, due to the fact that [the most recent commit](https://github.com/ably/ably-flutter/pull/277/commits/eb23ffe554fa92e456d3e8592a717cb25a6d1377) should not fail CI and [the preceding commit](https://github.com/ably/ably-flutter/pull/277/commits/60505047a0ae8dd94f124219d93effe4f0765534) passed CI.

Issue #260, opened 2021-11-25
Implement stats Rest and Realtime clients


ben-xD:
Implement following methods in Rest and Realtime
https://github.com/ably/ably-flutter/blob/b3cb8bf32efdcca3a314a612bf41950777950846/lib/src/platform/src/realtime/realtime.dart#L104-L105
and

https://github.com/ably/ably-flutter/blob/b3cb8bf32efdcca3a314a612bf41950777950846/lib/src/platform/src/rest/rest.dart#L61

These methods `throw UnimplementedError();`.

QuintinWillison:
To explain why this issue [was closed](https://github.com/ably/ably-flutter/issues/260#event-6139204388), it was a duplicate of https://github.com/ably/ably-flutter/issues/106.

PR #282, opened 2021-12-14
Time function implementation

ikbalkaya:
This PR contains` time()` function implementation that is raised in https://github.com/ably/ably-flutter/issues/260


ben-xD:
**Observation:** Rest also has an equivalent method to get the time, and we may have opportunity to implement that in this PR:

**Possible solution:** What we could do is have only 1 Platform method, `time`, which will use the handle provided to it to get either realtime or rest, and then call the time method on it. This is because the actual implementation for rest and realtime time() invocations in Ably Flutter would look identical. In the time method, we could use the handle to first try to get realtime client. If that is null, we could then try get the rest client.

**Example:** This is similar to what I did in `pushActivate. You can call rest.activate or realtime.activate, but it doesn't really matter for us, so I do:
https://github.com/ably/ably-flutter/blob/07aedc6f8292a979bbee43680098c77f5baf7118/android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java#L627-L636

android/src/main/java/io/ably/flutter/plugin/generated/PlatformConstants.java
```diff
@@ -5,445 +5,444 @@
 
 package io.ably.flutter.plugin.generated;
```
ben-xD, line 6: There's quite a big diff here, so I have formatted it using the java tool that this file was formatted with previously. This reduced the diff: 97a6dde2486b0b7ff1d467d57c3891c126b079fe

ikurek (approved):
LGTM 🙂

ikurek:
I'm closing this, in favor of #310 which is a rework of this PR on top of `main` instead of `enhancement/formatting`

PR #283, opened 2021-12-21
Stats implementation

ikbalkaya:
This PR includes full implementation of Rest.stats function for Flutter plugin. As stats is a slightly complicated and big object, this pull request might seem a bit huge but the scope is still just a map of a function.
Implemenation contains

* New type definitions on Flutter side for code generation
* New encoders for Java
* New encoders for Swift/Objective-C
* Function mappings
* A paginated result viewer addition to example app.

There were some differences between native implementations
* Structures were formed slightly differently. I created an issue in ably-java for this https://github.com/ably/ably-java/issues/733
* The way stats params were received was a bit more structured in Objective C than Java. In Objective C there is a dedicated object for this, where it's a Param[] in Java. I have created a separate category in Objective C to build that object from the map sent from Dart side.


bin/README.md
```diff
@@ -1,47 +0,0 @@
-## Code generation to keep platform constants in sync
```
ikurek, line 1: Is this removed purposely? I think it may be usefull 🤔

ikurek (approved):

PR #284, opened 2021-12-29
Fix / Debug CI

QuintinWillison:
I'm not convinced there's anything in the contents of this PR that _definitely_ fixed the tests... we were having instability in the sandbox environment prior to Christmas, so that settling down may have solved things.

The only thing that may have made a difference is https://github.com/ably/ably-flutter/pull/284/commits/fdcb82d3969c19dac322337865c171056c8729b5. 🤔


QuintinWillison:
I merged the contents of the branch underlying this PR into #255 and that has now also passed as green in CI. This is a good sign.

PR #289, opened 2022-01-07
Add workaround for firebase messaging

ikbalkaya:
This PR adds a warning for users who use Firebase messaging library along with Ably library

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,8 +96,6 @@ If you are using [official Firebase messaging library package](https://pub.dev/p
         tools:node="remove">
 </receiver>
 
-https://pub.dev/packages/firebase_messaging
-
 ### Troubleshooting
 
 - Running on simulator on M1 macs:
--- a/README.md
+++ b/README.md
@@ -92,9 +92,11 @@ Warning
 
 If you are using [official Firebase messaging library package](https://pub.dev/packages/firebase_messaging) along with this library, you must add following block to your Android application's manifest file under application tag. This is a workaround that prevents a conflict risen from two libraries installed together. 
 
+```xml
 <receiver android:name="io.ably.flutter.plugin.push.FirebaseMessagingReceiver"
         tools:node="remove">
 </receiver>
+```
 
 ### Troubleshooting
 
--- a/README.md
+++ b/README.md
@@ -88,16 +88,6 @@ Under the run/ debug configuration drop down menu, click `Edit Configurations...
 
 See [PushNotifications.md](PushNotifications.md) for detailed information on getting PN working with the example app.
 
-#### [Important Note regarding Firebase and Push Notifications](https://github.com/ably/ably-flutter/issues/226)
-
-If you are using the [official Firebase messaging library package](https://pub.dev/packages/firebase_messaging) along with this library, then you must add the following block to your Android application's manifest file within the `application` element. This is a workaround that prevents a conflict rising from the two libraries installed together. 
-
-```xml
-<receiver android:name="io.ably.flutter.plugin.push.FirebaseMessagingReceiver"
-        tools:node="remove">
-</receiver>
-```
-
 ### Troubleshooting
 
 - Running on simulator on M1 macs:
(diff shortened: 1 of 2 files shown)


AndyNicks (approved):

README.md
```diff
@@ -88,6 +88,16 @@ Under the run/ debug configuration drop down menu, click `Edit Configurations...
 
 See [PushNotifications.md](PushNotifications.md) for detailed information on getting PN working with the example app.
 
+###### **Please note:**
```
QuintinWillison, line 91: I think this needs to be H4, given previous heading is H3... or, in other words, this feels like misuse of a heading so I would perhaps also change the heading text (also makes this deep linkable):

```suggestion
#### Important Note regarding Firebase and Push Notifications
```

README.md
```diff
@@ -88,6 +88,16 @@ Under the run/ debug configuration drop down menu, click `Edit Configurations...
...
+
+If you are using [official Firebase messaging library package](https://pub.dev/packages/firebase_messaging) along with this library, you must add following block to your Android application's manifest file under application tag. This is a workaround that prevents a conflict risen from two libraries installed together.
```
QuintinWillison, line 93:
```suggestion
If you are using the [official Firebase messaging library package](https://pub.dev/packages/firebase_messaging) along with this library, then you must add the following block to your Android application's manifest file within the `application` element. This is a workaround that prevents a conflict rising from the two libraries installed together. 
```

QuintinWillison (changes requested):

README.md
```diff
@@ -88,6 +88,16 @@ Under the run/ debug configuration drop down menu, click `Edit Configurations...
 
 See [PushNotifications.md](PushNotifications.md) for detailed information on getting PN working with the example app.
 
+#### [Important Note regarding Firebase and Push Notifications](https://github.com/ably/ably-flutter/issues/226)
```
QuintinWillison, line 91: This readme actually has two sections with the same "Push Notifications" heading. 🙄 
[This one](https://github.com/ably/ably-flutter/tree/add-workaround_for_firebase_messaging#push-notifications), and [another one later on](https://github.com/ably/ably-flutter/tree/add-workaround_for_firebase_messaging#push-notifications-1).

Perhaps that was the initial impetus for the creation of the [dedicated, discrete document `PushNotifications.md`](https://github.com/ably/ably-flutter/blob/add-workaround_for_firebase_messaging/PushNotifications.md).

Regardless of reasons, the reality is that there are two places that could do with this information being visible... therefore my suggestion would be to move this new content to the dedicated document.

README.md
```diff
@@ -88,6 +88,16 @@ Under the run/ debug configuration drop down menu, click `Edit Configurations...
...
+
+If you are using the [official Firebase messaging library package](https://pub.dev/packages/firebase_messaging) along with this library, then you must add the following block to your Android application's manifest file within the `application` element. This is a workaround that prevents a conflict rising from the two libraries installed together.
```
QuintinWillison, line 93:
```suggestion
If you are using the [official Firebase messaging library package](https://pub.dev/packages/firebase_messaging) along with this library, then you must add the following block to your Android application's manifest file within the `application` element. This is a workaround that prevents a conflict that happens when the two libraries are installed together.
```

QuintinWillison (changes requested):

QuintinWillison:
Sorry, @ikbalkaya, I've spotted things on my second pass of this PR that I should have spotted the first time around... including something I introduced in one of my suggestions. My bad! 🤦

README.md
ikbalkaya: I was aware of that but I thought this would be a bit more prominent if we have this on the original readme. And I also think that the readme here is in needs of shortening, Perhaps we can then move push notification section here. But I will move this to the separate file for now

ikbalkaya: https://github.com/ably/ably-flutter/pull/289/commits/b72803873bccfe4099a92e0dd9fb999388f288ae

README.md
ikbalkaya: https://github.com/ably/ably-flutter/pull/289/commits/2e6c8ff3cf785104366d69ffccbd5d7e96c432f5

Issue #290, opened 2022-01-07
Investigate and possibly fix  broadcast receiver issue that causes ANR


ikbalkaya:
Issue spotted at https://github.com/ably/ably-flutter/issues/226#issuecomment-1006637559

There is an ANR problem that is raised from broadcast receiver implemented to interact push notifications in order to communicate to Dart side. This issue is happening when our library is used with Firebase messaging library. However the issue might be there and Firebase messaging library might be aggravating it. 

Investigate following
* Understand the impact of using multiple broadcast receivers that are accepting that has same intent filters. Will they both receive the intent? Will there be a priority? Is there a scheduling, priority etc that is designed for this case? Official documentation may not be enough to understand these, so might be a good idea to delve into Android source code.
* Compare implementation of broadcast receiver with the official one. 
* Try to identify execution block that causes ANR. This issue might be hard to replicate but hopefully adding the firebase library will make it easier. Then try to move this block to a background thread with minimal impact (without losing or changing any functionality) and see if that could quickly solve the issue (As recommended by @QuintinWillison thanks!)
* If above solutions do not work, try to adopt approach from [Firebase's official package ](https://github.com/FirebaseExtended/flutterfire/tree/master/packages/firebase_messaging/firebase_messaging)

ikbalkaya:
Some initial thoughts after comparing Firebase's official plugin with our Plugin. 
* Our plugin tries to do something quite similar to what Firebase's official plugin does. Also architecturally   looks like Ben B. tried to create something similar to Firebase team did.
* Firebase's implementation has a store to store / read /write messages, a job service for proper enqueue and dequeue and a background executor. We try to immediately execute incoming messages by trying to manually starting the app. I think we should follow the Firebase plugin implementation path. 
* We may not be able to directly use Firebase's Flutter library for now however I think our eventual aim to be using their library directly and customising the behaviour in Flutter side. That's a more pragmatic approach as we do not want to reinvent the wheel. 
* We can temporarily use Firebase library's Android plugin code, customising some behaviour on our side and making sure the dart-native channels are set up properly so that we don't have to change code in the dart side. I will also create the same package as what official plugin does in order to avoiding any clash for users who additionaly installs Firebase's official library.

ikbalkaya:
I tried to use components from Firebase messaging library and it ended up being very convoluted. It wasn't easy to migrate background channels used in the library to ours so I gave up.

[...]

PR #291, opened 2022-01-11
Fix anr issue

ikbalkaya:
This issue is a workaround that fixes ANR by moving blocking operations to a background threads.

* Moves `sendMessageToFlutterApplication` to a background thread

### DIFF
--- a/android/src/main/java/io/ably/flutter/plugin/push/FirebaseMessagingReceiver.java
+++ b/android/src/main/java/io/ably/flutter/plugin/push/FirebaseMessagingReceiver.java
@@ -29,6 +29,16 @@ public class FirebaseMessagingReceiver extends BroadcastReceiver {
 
   @Override
   public void onReceive(Context context, Intent intent) {
+    if (intent.getExtras() == null){
+      return;
+    }
+
+    listenForFlutterApplicationToFinishProcessingMessage(context);
+
+    Executors.newSingleThreadExecutor().execute(() -> {
+      sendMessageToFlutterApplication(context, intent);
+    });
+
     listenForFlutterApplicationToFinishProcessingMessage(context);
 
     Executors.newSingleThreadExecutor().execute(() -> {
--- a/android/src/main/java/io/ably/flutter/plugin/push/FirebaseMessagingReceiver.java
+++ b/android/src/main/java/io/ably/flutter/plugin/push/FirebaseMessagingReceiver.java
@@ -29,10 +29,6 @@ public class FirebaseMessagingReceiver extends BroadcastReceiver {
 
   @Override
   public void onReceive(Context context, Intent intent) {
-    if (intent.getExtras() == null){
-      return;
-    }
-
     listenForFlutterApplicationToFinishProcessingMessage(context);
 
     Executors.newSingleThreadExecutor().execute(() -> {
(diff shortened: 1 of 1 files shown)


QuintinWillison (changes requested):
In respect of @ikbalkaya [opening comment](https://github.com/ably/ably-flutter/pull/291#issue-1099036716):

> Checks for intent.getExtras() not being null. I know this is not directly related to this task but I felt that this is an important thing to quickly fix as this is essentially a very small PR

I'm concerned about the nature of this change, or more specifically what conclusion or assumption led to it being added to this PR.

First, given that `intent` is what's passed to `sendMessageToFlutterApplication`, then I would have expected it would be the responsibility of that method's implementation to check `intent.getExtras()` isn't `null`. Checking it in this method feels inappropriate, in that it assumes the implementation of the method being called needs the extras.

But, secondly, this additional check doesn't appear to solve anything described in the opening issue. I think this pull request should have a narrow scope, focussed on fixing only the problem described. I've not seen anything (e.g. in #290) that describes `null` being encountered here being a cause of issues. 🤔

android/src/main/java/io/ably/flutter/plugin/push/FirebaseMessagingReceiver.java
```diff
@@ -28,8 +29,15 @@
...
   @Override
   public void onReceive(Context context, Intent intent) {
+    if (intent.getExtras() == null){
+      return;
+    }
+
     listenForFlutterApplicationToFinishProcessingMessage(context);
-    sendMessageToFlutterApplication(context, intent);
+
+    Executors.newSingleThreadExecutor().execute(() -> {
```
KacperKluka, line 38: Isn't it too much memory consuming to create a new thread executor for each push notification we receive? :thinking:

KacperKluka (commented):
I've looked at the code and I think moving the work to a background thread shouldn't cause any leaks or troubles :thinking: My only concern is about creating new thread executors for each message :wink:

ikbalkaya:
> In respect of @ikbalkaya [opening comment](https://github.com/ably/ably-flutter/pull/291#issue-1099036716):
> 
> > Checks for intent.getExtras() not being null. I know this is not directly related to this task but I felt that this is an important thing to quickly fix as this is essentially a very small PR
> 
> I'm concerned about the nature of this change, or more specifically what conclusion or assumption led to it being added to this PR.
> 
> First, given that `intent` is what's passed to `sendMessageToFlutterApplication`, then I would have expected it would be the responsibility of that method's implementation to check `intent.getExtras()` isn't `null`. Checking it in this method feels inappropriate, in that it assumes the implementation of the method being called needs the extras.
> 
> But, secondly, this additional check doesn't appear to solve anything described in the opening issue. I think this pull request should have a narrow scope, focussed on fixing only the problem described. I've not seen anything (e.g. in #290) that describes `null` being encountered here being a cause of issues. 🤔

When comparing this with the official Firebase plugin, I realised they did that check before processing message. I tried sending an intent without any extras and that caused a crash. I just then struggled to prevent my urge to fix it 😄 but I agree, it's out of scope of this PR. So I removed it here https://github.com/ably/ably-flutter/pull/291/commits/6087a907e7ce84fb0254b4ba6a8f3956827ef0ce. I will open another issue to fix that

android/src/main/java/io/ably/flutter/plugin/push/FirebaseMessagingReceiver.java
ikbalkaya: You are right. I provided a single static executor service to be reused here https://github.com/ably/ably-flutter/pull/291/commits/07a865cdbcbf229e89cc283b7ca0a095b85e0480.

QuintinWillison (approved):

Issue #304, opened 2022-01-24
resultForDeactivate and resultForActivate are set to null in some situations in Android plugin


ikbalkaya:
**Problem**
`resultForDeactivate` and `resultForActivate` are set to null (sometimes, in some devices) that resulted customer reported crashes happening. See
https://github.com/ably/ably-flutter/issues/298

**Suspected cause**
`PushActivationEventHandlers` class has a declared inner broadcast receivers that access to `resultForActivate` and `resultForDeactivate` on `onRecieve` method. However it's possible (?) that PushActivationEventHandlers is disposed while broadcast receiver still receives messages (most probably on background).

ikbalkaya:
I did some tests to confirm whether broadcast receiver can exist in absense of `PushActivationEventHandlers`. Looks like it's the case. Setting up the `PushActivationEventHandlers` instance or any of its instance variables to null won't affect the ability for broadcast receiver to receive messages.

Issue #292, opened 2022-01-11
Consider making push an optional plugin


ikbalkaya:
### Problem
We are facing with some functional issues which are related to our push notification implementation. We are also facing compatibility issues when another library is used alongside with ours. See https://github.com/ably/ably-flutter/issues/226

We have two main issues
1. We install our custom implementations to apps which uses our library. For example we declare services, broadcast receivers to Android apps and users  unless they opt out in their manifest files. This situation creates a conflict described https://github.com/ably/ably-flutter/issues/226 and presents some inflexibility to users' side.
2. We are not using (or unable to use) [popular library](https://github.com/FirebaseExtended/flutterfire/tree/master/packages/firebase_messaging/firebase_messaging) to lessen our burden dealing with internals of push notifications. This path was followed as this library would also expose iOS plugin too, but we do not want to support Firebase's iOS plugin. 

Some thoughts on how we can address these problems
* We can create a separate plugin (such as ably-push) that will expose a common interface for users to interact with. Then we can provide our default implementations.  [Firebase messaging seems to have adopted a similar pattern](https://github.com/FirebaseExtended/flutterfire/tree/master/packages/firebase_messaging/firebase_messaging_platform_interface) . This looks like a federated Plugin, but they seem to have implemented only their web plugin based on this interface. iOS and Android plugins are still in the same package. 

* I was trying to find a way to use official Firebase library and how Firebase library is designed make things a bit harder. "Create your own implementation of Firebase" seems to be their motto, which means that we conform their interface and provide our own implementation. iOS and Android implementation is part of the app facing library and it's difficult to design something based on only a single platform. "Create an implementation of ably push (eg. Firebase" was my thinking and we can let user opt in for our Firebase implementation if they want to use Ably capabilities in push code

To be discussed...

ikbalkaya:
I have a theoretical idea on how we can solve this issue.

[...]

PR #307, opened 2022-01-24
Separate activation broadcast receiver from channel handler class

ikbalkaya:
This PR fixes **activation handling** code by 
* Fixes some issues with `PushActivationHandlers` singleton, by providing an instance method and converting some static methods into instance methods
* Moves activation broadcast receiver to a separate file, creating methods for registering with callback and unregistering.
* Respects the FlutterEngine lifecycle by registering the receiver when it's attached to engine and deregisters it when it's detached. 

Closes https://github.com/ably/ably-flutter/issues/304

Please note that changes here won't cause a manual application launch, meaning that any message from broadcast receiver will be received only when the Flutter engine is attached. I did this to prevent further bugs raising from inconsistencies related to Flutter engine lifecycle. 
However this is still a temporary solution. A full solution would be to separate third party push plugin so that we can reuse code directly from Firebase library (https://github.com/ably/ably-flutter/issues/292) or completely refactor push notification code on our plugin

Note: Java library will send activation results using broadcasts. That method might have been chosen for native Android apps, but it could be nice if we could have a async alternative that would give results through callback. If so we wouldn't need to deal with broadcast receivers on our plugin.

### DIFF
--- a/android/src/main/java/io/ably/flutter/plugin/push/PushActivationEventHandlers.java
+++ b/android/src/main/java/io/ably/flutter/plugin/push/PushActivationEventHandlers.java
@@ -31,7 +31,7 @@ public class PushActivationEventHandlers {
         pushActivationReceiver = new PushActivationReceiver();
     }
 
-    public static void instantiate(Context context, MethodChannel methodChannel) {
+    public static synchronized void instantiate(Context context, MethodChannel methodChannel) {
         if (instance == null) {
             instance = new PushActivationEventHandlers(context, methodChannel);
         }
--- a/android/src/main/java/io/ably/flutter/plugin/push/PushActivationEventHandlers.java
+++ b/android/src/main/java/io/ably/flutter/plugin/push/PushActivationEventHandlers.java
@@ -31,12 +31,16 @@ public class PushActivationEventHandlers {
     }
   }
 
-  public static void setResultForActivate(MethodChannel.Result result) {
-    instance.resultForActivate = result;
+  public static PushActivationEventHandlers getInstance() {
+    return instance;
   }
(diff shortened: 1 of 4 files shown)


android/src/main/java/io/ably/flutter/plugin/push/PushActivationEventHandlers.java
```diff
@@ -1,106 +1,97 @@
...
+            instance = new PushActivationEventHandlers(context, methodChannel);
+        }
+    }
 
-  public static void setResultForDeactivate(MethodChannel.Result result) {
-    instance.resultForDeactivate = result;
-  }
+    public static PushActivationEventHandlers getInstance() {
+        return instance;
+    }
```
KacperKluka, line 42: This can lead to NPEs since you need to first call `instantiate()` before calling this method to assign some value to the `instance` field. However, I see that this is how it has worked before so I wouldn't want to block the PR on this thing :wink:

KacperKluka (approved):
Overall looks good :+1: One small problem but it was present before so I'll let you decide what to do with it :wink:

android/src/main/java/io/ably/flutter/plugin/push/PushActivationEventHandlers.java
ikbalkaya: You are right and you are right 😄 This should also be protected by the its use case. Meaning only when flutter engine is attached (meaning methods will be able to communicate over a channel) this will be called. But this will most probably be history for next couple of changes for push notifications

ikurek (approved):
LGTM 🙂

PR #309, opened 2022-02-01
Readme enchancements for example app

ikurek:
I've noticed that example app requirements aren't complete, so I did a few changes to make running ample app easier:
* Added a missing `launch.json` file. It was mentioned in the docs, but it wasn't commited, most probably because of gitignore covering the whole `vscode` folder
* Added a short note about Firebase instance configuration being required for Android builds. If there's no configuration file in the reposiotry, building the sample app fails, so I think it's worth to have that information in readme file

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ Under the run/ debug configuration drop down menu, click `Edit Configurations...
 #### Visual Studio Code
 
 - Under `Run and Debug`,
-  - Select the gear icon to view `launch.json`
+  - Select the gear icon to view [launch.json](.vscode/launch.json)
   - Add your Ably API key to the `configurations.args`, i.e. replace `replace_with_your_api_key` with your own Ably API key.
   - To choose a specific device when more than one are connected: to launch on a specific device, make sure it is the only device plugged in. To run on a specific device when you have multiple plugged in, add another element to the `configuration.args` value, with `--device-id=replace_with_device_id`
     - Make sure to replace `replace_with_your_device` with your device ID from `flutter devices`
--- a/README.md
+++ b/README.md
@@ -54,6 +54,10 @@ Features that we do not currently support, but we do plan to add in the future:
 
 - To run the example app, you need an Ably API key. Create a free account on [ably.com](https://ably.com/) and then use your API key from there in the example app.
 - Clone the project
+- For Android, a Firebase instance configuration is required to build the app. In order run the application on Android device:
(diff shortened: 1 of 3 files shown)


QuintinWillison (approved):
Looks sensible, thanks.

Please make sure to check, if you've not already done so, whether any of this conflicts with what's in the [contributing guide](https://github.com/ably/ably-flutter/blob/main/CONTRIBUTING.md).

README.md
```diff
@@ -54,6 +54,10 @@ Features that we do not currently support, but we do plan to add in the future:
 
 - To run the example app, you need an Ably API key. Create a free account on [ably.com](https://ably.com/) and then use your API key from there in the example app.
 - Clone the project
+- For Android, a Firebase instance configuration is required to build the app. In order run the application on Android device:
+  - Create a Firebase project following [official Firebase guide](https://firebase.google.com/docs/android/setup#create-firebase-project) (you can also use an existing one)
+  - Register the sample app with your Firebase project using `io.ably.flutter.example` as package name
+  - Download `google_services.json` file from Firebase and put it into [example/android/app](example/android/app) directory
```
ikbalkaya, line 60: There is a whole other Readme for push notifications here https://github.com/ably/ably-flutter/blob/main/PushNotifications.md
I think PN related documentation is all covered there. I would argue that we should aim for a much shorter Readme for this repo and move PN related documentation to here, but probably something for the future

ikbalkaya (changes requested):
Good improvements , thank you. I  think we could also provide a similar build configuration for intelliJ users (in a different PR). However push notification documentation is currently completely separate from this main Readme. Could you please check if there is any missing steps you can see on that documentation?

ikurek:
> Good improvements , thank you. I think we could also provide a similar build configuration for intelliJ users (in a different PR). However push notification documentation is currently completely separate from this main Readme. Could you please check if there is any missing steps you can see on that documentation?

I don't think there's anything missing in the push notification docs, for me the sample started working just fine after I did what was described there

README.md
ikurek: The Firebase setup is described more in-depth in the `PushNotifications.md` file, but I thought it would be nice to describe the minimum required to run the app in the base documentation as well, because when I tried to run the sample, even without push notifications, the build failed because of missing Firebase configuration.

ikbalkaya: Ok, I do agree with that (minimal information to run the app on main readme file). There is some work we will later do on push notifications :)

ikbalkaya (approved):
LGTM, thanks

PR #321, opened 2022-02-07
`AblyInstanceStore` concurrency

### LINKED ISSUE #308: `java.lang.ArrayIndexOutOfBoundsException` thrown by `AblyInstanceStore`'s `setPaginatedResult` method
We've had a customer report that a small subset of their app users have experienced this crash.

Information provided via Firebase Crashlytics...

```
Fatal Exception: java.lang.ArrayIndexOutOfBoundsException
src.length=11 srcPos=0 dst.length=25 dstPos=0 length=12
java.lang.System.arraycopy (System.java)
com.android.internal.util.GrowingArrayUtils.insert (GrowingArrayUtils.java:141)
android.util.LongSparseArray.put (LongSparseArray.java:218)
io.ably.flutter.plugin.AblyInstanceStore.setPaginatedResult (AblyInstanceStore.java:105)
io.ably.flutter.plugin.AblyMethodCallHandler$3.onSuccess (AblyMethodCallHandler.java:280)
io.ably.flutter.plugin.AblyMethodCallHandler$3.onSuccess (AblyMethodCallHandler.java:277)
io.ably.lib.http.BasePaginatedQuery$ResultRequest$1.onSuccess (BasePaginatedQuery.java:215)
io.ably.lib.http.BasePaginatedQuery$ResultRequest$1.onSuccess (BasePaginatedQuery.java:212)
io.ably.lib.http.HttpScheduler$AsyncRequest.setResult (HttpScheduler.java:321)
io.ably.lib.http.HttpScheduler$AblyRequestWithFallback.run (HttpScheduler.java:206)
java.util.concurrent.ThreadPoolExecutor.runWorker (ThreadPoolExecutor.java:1167)
java.util.concurrent.ThreadPoolExecutor$Worker.run (ThreadPoolExecutor.java:641)
java.lang.Thread.run (Thread.java:919)
```

From the crash log for a Nokia 5.4 device running Android 10.

Other potentially salient details:

- 6 instances of this crash seen across 14,600 installations of the app
- **OS**: 50% running Android 9
- **Devices**: 33% Samsung, 33% Xiaomi, 17% Vivo, 17% HMD Global
- **Device states**: 50% background

Also see [this internal Ably Support Team Slack thread](https://ably-real-time.slack.com/archives/C8SPU4589/p1643257377074100).

QuintinWillison:
Fixes #308.

### DIFF
--- a/android/src/main/java/io/ably/flutter/plugin/AblyInstanceStore.java
+++ b/android/src/main/java/io/ably/flutter/plugin/AblyInstanceStore.java
@@ -124,15 +124,15 @@ class AblyInstanceStore {
         }
     }
 
-    ClientHandle reserveClientHandle() {
+    synchronized ClientHandle reserveClientHandle() {
         return new ReservedClientHandle(handleSequence.next());
     }
 
-    AblyRest getRest(final long handle) {
+    synchronized AblyRest getRest(final long handle) {
         return restInstances.get(handle);
     }
 
-    AblyRealtime getRealtime(final long handle) {
+    synchronized AblyRealtime getRealtime(final long handle) {
         return realtimeInstances.get(handle);
     }
 
@@ -146,23 +146,23 @@ class AblyInstanceStore {
      * @param handle integer handle to either AblyRealtime or AblyRest
      * @return AblyBase
      */
-    AblyBase getAblyClient(final long handle) {
+    synchronized AblyBase getAblyClient(final long handle) {
         AblyRealtime realtime = getRealtime(handle);
         return (realtime != null) ? realtime : getRest(handle);
     }
     
-    Push getPush(final long handle) {
+    synchronized Push getPush(final long handle) {
         AblyRealtime realtime = getRealtime(handle);
         return (realtime != null) ? realtime.push : getRest(handle).push;
     }
     
-    PushChannel getPushChannel(final long handle, final String channelName) {
+    synchronized PushChannel getPushChannel(final long handle, final String channelName) {
         return getAblyClient(handle)
                 .channels
                 .get(channelName).push;
     }
 
-    long setPaginatedResult(AsyncPaginatedResult result, Integer handle) {
+    synchronized long setPaginatedResult(AsyncPaginatedResult result, Integer handle) {
         long longHandle;
         if (handle == null) {
             longHandle = handleSequence.next();
@@ -173,11 +173,11 @@ class AblyInstanceStore {
         return longHandle;
     }
 
-    AsyncPaginatedResult<Object> getPaginatedResult(long handle) {
+    synchronized AsyncPaginatedResult<Object> getPaginatedResult(long handle) {
         return paginatedResults.get(handle);
     }
 
-    void reset() {
+    synchronized void reset() {
         for (int i = 0; i < realtimeInstances.size(); i++) {
             long key = realtimeInstances.keyAt(i);
             AblyRealtime r = realtimeInstances.get(key);
--- a/android/src/main/java/io/ably/flutter/plugin/AblyInstanceStore.java
+++ b/android/src/main/java/io/ably/flutter/plugin/AblyInstanceStore.java
@@ -23,9 +23,7 @@ import io.ably.lib.types.ClientOptions;
  * client by calling @{AblyInstanceStore#getRealtime(final long handle)}.
  */
 class AblyInstanceStore {
-
     private static final AblyInstanceStore instance = new AblyInstanceStore();
-    private long nextHandle = 1;
 
     // Android Studio warns against using HashMap with integer keys, and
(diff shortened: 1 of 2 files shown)


android/src/main/java/io/ably/flutter/plugin/AblyInstanceStore.java
```diff
@@ -34,38 +32,107 @@
...
     }
 
-    AblyRest getRest(final long handle) {
-        return restInstances.get(handle);
+    private static class HandleSequence {
+        private volatile long nextHandle = 1;
+
+        synchronized long next() {
+            return nextHandle++;
+        }
```
KacperKluka, line 124: This whole class could be replaced by the standard `AtomicInteger()` class and its `getAndIncrement()` method :wink: However, the current code also works so I'll let you decide whether to change it or not

android/src/main/java/io/ably/flutter/plugin/AblyInstanceStore.java
```diff
@@ -34,38 +32,107 @@
...
+            final AblyRealtime realtime = new AblyRealtime(clientOptions);
+            realtime.setAndroidContext(applicationContext);
+            realtimeInstances.put(handle, realtime);
+            return handle;
+        }
+
+        synchronized long use() {
+            if (used) {
+                throw new IllegalStateException("Reserved handle has already been used to create a client instance (handle=" + handle + ").");
+            }
```
KacperKluka, line 114: You declare the private field `used` with the initial value `false` but then you never change it to `true`, however, here you want to throw an exception if it's `true` :thinking: When should the `used` field change to `true`? :thinking:

KacperKluka (commented):
Looks good, I've only spotted two things :wink:

android/src/main/java/io/ably/flutter/plugin/AblyInstanceStore.java
QuintinWillison: Well spotted, thanks! Will fix that.

android/src/main/java/io/ably/flutter/plugin/AblyInstanceStore.java
QuintinWillison: Good suggestion, thanks. My JDK knowledge isn't as deep as I would like it to be. 😁 I'll take a look at swapping that in.

QuintinWillison:
Converting this PR back to Draft state while I work on @KacperKluka's review comments.

android/src/main/java/io/ably/flutter/plugin/AblyInstanceStore.java
QuintinWillison: https://github.com/ably/ably-flutter/pull/321/commits/bf67c686c6453d1a5dbd23f55170ed8330b12ca3

android/src/main/java/io/ably/flutter/plugin/AblyInstanceStore.java
QuintinWillison: https://github.com/ably/ably-flutter/pull/321/commits/003ccbf921496d95d3ad5f81ef090da8db32b9eb

android/src/main/java/io/ably/flutter/plugin/AblyInstanceStore.java
```diff
@@ -34,38 +34,100 @@
...
-     * Returns a handle representing the next client that will be created. This handle can be used
-     * to get the client **after** it is instantiated using [createRest] or [createRealtime].
+     * A reserved client handle. Safe to be used from any thread.
+     *
+     * Instances support the creation of a single Rest or Realtime instance, where only one of the
+     * create methods may be called and it may only be called once.
      */
-    long getHandleForNextClient() {
-        return nextHandle;
+    interface ClientHandle {
```
ikurek, line 49: A little style suggestion - maybe the `ClientHandle` and/or it's implementation can be moved to a separate file? I'm not a fan of too long files, and from logical point of view, I'm not sure if it makes sense for this interface to be located inside `AblyInstanceStore`, since it's also used in other classes like `AblyMehtodCallHandler`. That's nothing blocking, since it's just related to the code style 🙂

ikurek (approved):
The solution looks fine to me 🙂

android/src/main/java/io/ably/flutter/plugin/AblyInstanceStore.java
QuintinWillison: I am broadly supportive of that viewpoint, however there's always a balance in my view for such 'micro' interfaces where they only relate to an API offered by the primordial class in this file. The only reason `AblyMethodCallHandler` uses this interface is because it _also_ uses other code from this file for the same purpose.

So, in this case, I'm going to leave this as is. In future I would be open to it being refactored, if the context changes.

KacperKluka (approved):
LGTM :+1:

PR #326, opened 2022-02-09
Simplified GitHub actions flow

### LINKED ISSUE #317: No need for GitHub workflows to explicitly run `flutter pub get` in the `example` folder
See this [internal Slack thread](https://ably-real-time.slack.com/archives/C0178QGC093/p1644235380888039) where @ikurek and myself discuss [this line](https://github.com/ably/ably-flutter/blob/94e3cb001bfd8b6f4c5421c1f47e48f72c49eb88/.github/workflows/check.yaml#L17) in the `check` workflow. Igor explained it to me:

> The example folder is part of the Ably plugin structure, and test-integration is not, so when you read pubspec file from the main direcotry, it only fetches the dependencies from plugin components, so root (ably-flutter) and example directories 

Because I was observing that `test_integration` wasn't in the mix here:

```
ably-flutter % flutter pub get
Running "flutter pub get" in ably-flutter...                     2,274ms
Running "flutter pub get" in example...                          2,077ms
```

### LINKED ISSUE #325: Enable Flutter cache on Github Actions
Each Github Actions step redownloads the entire Flutter engine with each CI run. This step usually runs for around a minute on each action (we have 5 of them attached to each PR right now). This can be simplified, reducing the CI runtime with a solution provided by [flutter-action](https://github.com/subosito/flutter-action) step  (which we already use) out of the box:

```yaml
  with:
    channel: 'stable'
    cache: true
```

The caching should reduce the need to redownload Flutter with each CI action run, reducing the CI runtime by around 4-5 minutes on each pull request.

ikurek:
Closes #325 and closes #317 , and additionally updates `flutter-action` step version to `2`. This should greatly reduce the CI runtime by reducing the number of components downloaded with each run

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -12,7 +12,9 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
-      - uses: subosito/flutter-action@v1
+      - uses: subosito/flutter-action@v2
+        with:
+          channel: 'stable'
 
       - name: 'Install dependencies'
         run: |
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -21,9 +21,7 @@ jobs:
       - name: 'Install dependencies'
         run: |
           flutter pub get
-          cd test_integration
-          flutter pub get
-          cd ..
+          cd test_integration && flutter pub get
 
       - name: 'Check formatting issues in the code'
         run: flutter format --set-exit-if-changed .
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -14,7 +14,6 @@ jobs:
 
       - uses: subosito/flutter-action@v2
         with:
-          channel: 'stable'
           cache: true
 
       # This step requires fetch of test_integration packages because flutter format and 
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -17,20 +17,22 @@ jobs:
           channel: 'stable'
           cache: true
 
+      # This step requires fetch of test_integration packages because flutter format and 
+      # flutter analyze check test_integration package too
       - name: 'Install dependencies'
         run: |
           flutter pub get
-          cd test_integration && flutter pub get && cd ..
-          cd example && flutter pub get && cd ..
+          cd test_integration
+          flutter pub get
+          cd ..
 
       - name: 'Check formatting issues in the code'
         run: flutter format --set-exit-if-changed .
-
-      - name: 'Flutter analyze help'
-        run: 'flutter analyze --help'
         
+      # Suppressing info - deprecated entries fall under info, and CI shouldn't fail if there are any Deprecated APIs
+      # FIXME(ikurek): We should probably just configure linter to ignore deprecation, instead of disabling errors here
       - name: 'Statically analyze the Dart code for any errors'
-        run: 'flutter analyze --no-pub --no-fatal-infos .'  # suppressing info - deprecated entries fall under info, and CI shouldn't fail if there are any Deprecated APIs
+        run: 'flutter analyze --no-pub --no-fatal-infos .'
 
       - name: 'Run unit tests'
         run: flutter test --no-pub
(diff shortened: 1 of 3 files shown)


.github/workflows/check.yaml
```diff
@@ -12,13 +12,13 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
-      - uses: subosito/flutter-action@v1
+      - uses: subosito/flutter-action@v2
+        with:
+          channel: 'stable'
```
QuintinWillison, line 17: Please prefer _convention over configuration_. This is [the default](https://github.com/subosito/flutter-action/blob/5e1529bc12adae5b26f30f0a4046ac743d6f0c5d/action.yml#L14) value for `channel` so does not need to be specified. Same goes for other instances in this PR.

KacperKluka (approved):
LGTM :+1:

.github/workflows/check.yaml
```diff
@@ -12,13 +12,13 @@ jobs:
...
+          cache: true
```
QuintinWillison, line 18: Why is this here?

1. The [documentation for this action](https://github.com/subosito/flutter-action#usage) states that this interacts with [GitHub's `actions/cache`](https://github.com/actions/cache), which you're not using in this workflow, so it's doing nothing
2. GitHub runners are ephemeral so this would only work for this single workflow runtime instance anyway, so I fail to see what you're optimising for

QuintinWillison (changes requested):

.github/workflows/check.yaml
ikurek: Personally, I prefer to specify values like that to make it clear that the step should always run with latest stable version, but if the convention is different, that's fine 🙂 Removed those definitions in https://github.com/ably/ably-flutter/pull/326/commits/34fbc661f27552ba14413eebae362848f110bc7c

.github/workflows/check.yaml
ikurek: 1. Ive checked the [example usage of this flow in the official repository](https://github.com/subosito/flutter-action/blob/main/.github/workflows/workflow.yml) and the `test_cache` job doesn't define `actions/cache` anywhere, so I thought that this is configured by the step itself, it also looks like that from the yaml perspective, since the code for runner looks like that:

```yaml
runs:
  using: 'composite'
  steps:
    - if: ${{ inputs.cache == 'true' }}
      uses: actions/cache@v2
      with:
        path: ${{ inputs.cache-path }}
        key: ${{ inputs.cache-key }}-${{ inputs.channel }}-${{ inputs.flutter-version }}
```

My thinking is, this should setup caching by itself, but it's possible my thinking isn't correct

ikurek: 2. Currently, on each pull request we open 5 checks, and each of them redownloads the Flutter engine. My idea was to reduce the amount of downloads, by simply enabling the cache for flutter step. I'm not too familiar with GitHub actions, but as far as I understand, according to [official docs](https://docs.github.com/en/actions/advanced-guides/caching-dependencies-to-speed-up-workflows#about-caching-workflow-dependencies) these dependencies can be cached between build for 7 days. If that's not the case, then your point is valid and enabling the cache doesn't make sense

QuintinWillison: 1. Hmm. Curious. If it's 'composite', then why do we need to feed it in as an input? That doesn't make logical sense. 😕 
2. Oh, ok, it sounds like my understanding was incorrect regarding the ephemeral nature of the cache. It _does_ outlive that single workflow run. There's something to learn every day!

ikurek:
Added a small commit to fix invalid paths for `test_integration` directory: https://github.com/ably/ably-flutter/pull/326/commits/4eae6f8e2b09adf37931ebe43da6dc26a674d18b

.github/workflows/check.yaml
ikurek: I think the `composite` is just here so the action can use multiple steps - one for cache setup and one to run the script.

QuintinWillison (approved):

PR #327, opened 2022-02-09
Separated rest and realtime ChannelOptions and HistoryParams models

### LINKED ISSUE #241: Realtime types (e.g. `RealtimeHistoryParams`) extending Rest ones (e.g. `RestHistoryParams`) can cause problems (crashes)
Realtime types (e.g. `RealtimeHistoryParams`) extending Rest ones (e.g. `RestHistoryParams`) can cause problems (crashes)

Using 
```
                          _rest!.channels.get(defaultChannel).history(
                                ably.RealtimeHistoryParams(
                                  limit: 10,
                                  untilAttach: true,
                                ),
                              ),
```

Will crash the app with 
> *** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'ARTRealtimeHistoryQuery used from outside ARTRealtimeChannel.history'

Notice that `RealtimeHistoryParams` is passed as an argument to a `RestChannel.history`. Because in Ably Flutter, `RealtimeHistoryParams` is a  `RestHistoryParams`, the user is allowed to do this. We should stop having `RealtimeHistoryParams` being a `RestHistoryParams`.

There are other instances of this behaviour, where a realtime class extends a rest class because it shares some code. We should consider cases like this and object oriented programming, and find a solution to it.

ikurek:
Closes #241, the `rest` and `realtime` models of `ChannelOptions` and `HistoryParams` are now a separate beings, I've also updated the documentation with correct ID's and reference URL's for each class component, and included additional information regarding default values of optional parameters


KacperKluka (approved):
LGTM :+1:

QuintinWillison (approved):

PR #312, opened 2022-02-04
Android embedding v2

ikurek:
This PR is a migration supposed to solve #311, by fully adapting Flutter Android Embedding V2 following the [official migration guide](https://github.com/flutter/flutter/wiki/Upgrading-pre-1.12-Android-projects). Besides the migration itself, additional compatibility changes include:
* Forced Kotlin version for example app (Flutter recommends setting 1.5.31)
* Changed Android target and compile SDK versions to latest (32)
* Updated a few Flutter dependencies for example app
* Updated test dependency version for dart library
* Set Android application as exported in Manifest file, to comply with Android S requirements

PR #328, opened 2022-02-10
Removed deprecated registration method from Android Embedding V1

### LINKED ISSUE #178: Drop support for Android v1 embedding API
The Android v2 embedding API has been released at least 2 years ago, as the original V1 embedding API was deprecated 2 years ago: https://github.com/flutter/engine/blame/233a6e01fbb0f3cb407da7bee6817ccdcb6a9822/shell/platform/android/io/flutter/app/FlutterActivity.java#L19

The following comments were added in Ably Flutter in January 2020: 
```
    // This static function is optional and equivalent to onAttachedToEngine. It supports the old
    // pre-Flutter-1.12 Android projects. You are encouraged to continue supporting
    // plugin registration via this function while apps migrate to use the new Android APIs
    // post-flutter-1.12 via https://flutter.dev/go/android-project-migration.
    //
    // It is encouraged to share logic between onAttachedToEngine and registerWith to keep
    // them functionally equivalent. Only one of onAttachedToEngine or registerWith will be called
    // depending on the user's project. onAttachedToEngine or registerWith must both be defined
    // in the same class.
```

It is unlikely any users still use this. If they do, they should upgrade to use the new embedding API, which is shown in https://github.com/flutter/flutter/wiki/Upgrading-pre-1.12-Android-projects

We should remove the unused `public static void registerWith(Registrar registrar)` method since we do not currently test it.

ikurek:
The migration to Android Embedding V2 was started with #312, because latest versions of Flutter dropped the support for V1 entirely. This PR resolves #178 by removing the last component of Android Embedding V1.

The V1 is deprecated for a long time now, and we probably don't want to allow it's usage, because it can lead to unexpected behavior in android-related components. The removed method is the last component of V1 embedding I could find, so this should finish the migration process.


android/src/main/java/io/ably/flutter/plugin/AblyFlutter.java
```diff
@@ -39,22 +39,6 @@ public void onAttachedToEngine(@NonNull FlutterPluginBinding flutterPluginBindin
...
-    // them functionally equivalent. Only one of onAttachedToEngine or registerWith will be called
-    // depending on the user's project. onAttachedToEngine or registerWith must both be defined
-    // in the same class.
-    public static void registerWith(Registrar registrar) {
-        Context applicationContext = registrar.context().getApplicationContext();
-        final AblyFlutter ably = new AblyFlutter();
-        registrar.addNewIntentListener(ably);
-        ably.setupChannels(registrar.messenger(), applicationContext);
-    }
-
```
ikbalkaya, line 57: Can you / should you inform users about this adding some content to Readme?

ikbalkaya (changes requested):
A request to inform users that we only support embedding v2 (if appropriate)

android/src/main/java/io/ably/flutter/plugin/AblyFlutter.java
ikurek: Good question, but in our case, I don't think it's necessary to disclose this anywhere.
* This was deprecated in Flutter version 1.22, as far as I can tell, our plugin targets Flutter > 2.0, so I highly doubt it's even possible to use this in current configuration.
* Wider context on this is provided in official documentation [here](https://docs.flutter.dev/release/breaking-changes/android-v1-embedding-create-deprecation#context), but in general, this is a very old solution that should have been removed quite a while ago. Following Flutter release notes: `As of Q2 2020, only 26% of applications used the v1 embeddings`, so I think we can safely assume that this shouldn't affect anyone

So I think disclosing this may be just unwanted noise in the docs

ikbalkaya (approved):

KacperKluka (commented):
Looks good, however, I don't have much flutter knowledge, therefore, I'm going to leave a comment instead of an approval :wink:

QuintinWillison (approved):

PR #331, opened 2022-02-14
VS code launch configurations

### LINKED ISSUE #330: Create launch configurations to simplify running on VS Code
For now, Android Studio users have access to complete set of configurations in `.run` directory including:

* Running the example app
* Running the integration test app
* Running the integration test driver
* Running all unit tests available in project

This configuration makes it easy to run all components in Ably library project, but there's no corresponding solution for VS Code. Because VS Code is one of two officially recommended Flutter development environments and it's popularity is pretty high, I think it would be nice to support the same run configurations as Android Studio does

ikurek:
I've included additional run configurations for VS Code, to match those available on Android Studio. This `launch.json` includes additional configuration for:

* integration tests app
* integration tests driver
* compound configuration for integration tests (runs both app and driver)
* unit tests

There are also small changes to readme files to include information about VS Code launch configurations. This should close #330

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -2,14 +2,51 @@
     "version": "0.2.0",
     "configurations": [
         {
-            "name": "example",
+            "name": "Example App",
             "cwd": "example",
             "request": "launch",
             "type": "dart",
             "flutterMode": "debug",
             "args": [
-                "--dart-define",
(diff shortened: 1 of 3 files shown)


lukasz-szyszkowski (approved):

ikbalkaya (approved):
LGTM

Issue #335, opened 2022-02-21
Missing link to `ApplicationState` in documentation


ikurek:
Dartdoc output indicates that a link to `[ApplicationState]` in `PushNotificationEvents` class doesn't lead anywhere. Most probably the `ApplicationState` class was removed and the comment link is a leftover and should be removed. Dartdoc output:
```
warning: unresolved doc reference [ApplicationState]
    from ably_flutter.PushNotificationEvents.onNotificationTap: (file:///home/runner/work/ably-flutter/ably-flutter/lib/src
```

It would be nice to remove that link, to avoid end-user confusion with this method

PR #336, opened 2022-02-21
Remove deprecated ApplicationState docs reference

ikurek:
This should close https://github.com/ably/ably-flutter/issues/335, and remove Dartdoc generation warning.

It looks like `ApplicationState` class was removed completely and this is just a leftover comment. After investigation, I decided to remove it entirely, because current implementation doesn't provide any way to differentiate if push message was received in foreground or background, so there was no replacement to include in the docs.


ikbalkaya (approved):
LGTM 👍

QuintinWillison (approved):

PR #337, opened 2022-02-22
Update sample app message encryption

### LINKED ISSUE #334: Investigate and fix encryption in sample app
When investigating #207 with @ikbalkaya we've noticed that encrypted messages don't work in the sample application. When an encrypted message is sent it appears decrypted in the sample app:

<p align="center">
    
</p>

I've followed the hint in the sample application that states:

>You can confirm messages are encrypted by visiting your application dashboard on ably.com > "Dev console" tab > "Channels", and attach to the "encrypted-test-channel" channel.

But on the web panel, the messages also don't appear encrypted:

<p align="center">
    
</p>

There's probably some sort of bug in the sample application causing messages that should be encrypted to be sent in plaintext

ikurek:
This closes #334. I've removed the `EncryptedMessagingService` and instead I've added a switch for realtime and rest channels to enable/disable encryption ad-hoc. Benefits of this solution:

* Less code in sample app
* Easier to understand for end-users how to enable and disable encryption on channel
* Encrypted data is now visible in the app itself. Previously, all messages were decrypted before display. Now if encryption is disabled, messages that were encrypted are seen as encrypted in message history)

Here's a sample of message display with encryption on and off:

<p align="center">
    
    
</p>

I think the whole sample application could use a little bit of UI/UX changes, but it's out of scope so I left the message display as-is


QuintinWillison (approved):
Thanks, looks good to me. I appreciate you 'scoping down' the work as this made it much quicker and easier for reviewers. Nice work! 😍

ikbalkaya (approved):
Looks good to me 👏

PR #338, opened 2022-02-23
Updated Firebase Cloud messaging dependency on Android to version 23.0.0

### LINKED ISSUE #320: Upgrade `firebase-messaging` to version 23.0.0
We're currently on version 22.0.0.

From Google's [release notes](https://firebase.google.com/support/release-notes/android#messaging_v23-0-0):

> Added methods for determining and controlling whether Google Play services is set as the app’s notification delegate. By default, FCM will now set Google Play services as the app’s notification delegate so that it is allowed to display notifications for the app. This could be used in the future to show an app’s notifications without needing to start the app, which may improve message reliability and timeliness.

Does that carry any implications for us?

Also, Google state that it now requires target devices to run Android 4.4 or higher, with minimum API Level 19 (KitKat).

ikurek:
Should resolve #320. I've tested it and I don't see any issues that could affect Ably SDK after the update. The minimal required Android version was already set to KitKat (API 19) so it shouldn't be an issue.


KacperKluka (approved):
LGTM :+1:

PR #340, opened 2022-02-25
LogHandler removal & logging improvements

### LINKED ISSUE #238: Fix LogHandler (use it)
LogHandler is an option on `ClientOptions`. Users can set it, but unfortunately Ably Flutter ignores it. This is because the log handler is dart code which can't be passed directly to the platform side.

We would have to store the logHandlers, and pass a handle to the Platform side instead. When Ably Cocoa and Ably Java want to use them, this would call the dart side.

ikurek:
This should resolve #238, although it does not use the mentioned `LogHandler`. Instead, the changes contain:

* Removing the `LogHandler` entirely from Dart code
* Replacing Android native logging with `Log` implementation from `ably-java`. Luckily, their implementation is the same so it was only a matter of imports
* Replacing ios `print` and `NSLog` calls with `ARTLog` implementation from `ably-cocoa`. This was not possible everywhere, because unlike `Log` from `ably-java` the `ARTLog` isn't a singleton, so I only updated a few places

With these changes, most of the logging from Android and iOS platform native code should respect logging level set in Ably `ClientOptions`.

### DIFF
--- a/lib/src/authentication/src/client_options.dart
+++ b/lib/src/authentication/src/client_options.dart
@@ -28,7 +28,7 @@ class ClientOptions extends AuthOptions {
   @Deprecated(
     'Not used, as log messages are handled by the default mechanism '
     'in the underlying SDK. This instance variable will be removed '
-    'in future.',
+    'in a future release.',
   )
   LogHandler? logHandler;
 
--- a/lib/src/authentication/src/client_options.dart
+++ b/lib/src/authentication/src/client_options.dart
@@ -19,6 +19,19 @@ class ClientOptions extends AuthOptions {
   /// https://docs.ably.com/client-lib-development-guide/features/#TO3a
   String? clientId;
 
+  /// Custom log handler
+  ///
+  /// For discussion about removing this component see
+  /// https://github.com/ably/ably-flutter/issues/238
+  ///
+  /// https://docs.ably.com/client-lib-development-guide/features/#TO3c
+  @Deprecated(
+    'Not used, as log messages are handled by the default mechanism '
+    'in the underlying SDK. This instance variable will be removed '
+    'in future.',
+  )
+  LogHandler? logHandler;
+
   /// Controls the level of verbosity of log messages from the library
   ///
   /// Use constants from [LogLevel] to pass arguments
--- a/lib/src/authentication/src/client_options.dart
+++ b/lib/src/authentication/src/client_options.dart
@@ -19,11 +19,6 @@ class ClientOptions extends AuthOptions {
   /// https://docs.ably.com/client-lib-development-guide/features/#TO3a
   String? clientId;
 
-  /// Custom log handler
-  ///
-  /// https://docs.ably.com/client-lib-development-guide/features/#TO3c
-  LogHandler? logHandler;
-
   /// Controls the level of verbosity of log messages from the library
   ///
   /// Use constants from [LogLevel] to pass arguments
--- a/lib/src/logging/src/log_handler.dart
+++ b/lib/src/logging/src/log_handler.dart
@@ -1,9 +1,0 @@
-import 'package:ably_flutter/ably_flutter.dart';
-
-/// Custom handler to handle SDK log messages
-///
-/// https://docs.ably.com/client-lib-development-guide/features/#TO3c
-typedef LogHandler = void Function({
-  String? msg,
-  AblyException? exception,
-});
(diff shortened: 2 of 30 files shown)


lib/src/authentication/src/client_options.dart
```diff
@@ -19,11 +19,6 @@ class ClientOptions extends AuthOptions {
   /// https://docs.ably.com/client-lib-development-guide/features/#TO3a
   String? clientId;
 
-  /// Custom log handler
-  ///
-  /// https://docs.ably.com/client-lib-development-guide/features/#TO3c
-  LogHandler? logHandler;
```
QuintinWillison, line 25: I'm wondering if there might be any customers who are currently using this interface, despite the fact that it wasn't actually implemented. Perhaps an alternative approach, given we're only going to be doing the next release of this plugin as a patch version bump, might be:

1. **For the compiler**: Annotate both this instance variable and the `LogHandler` class as deprecated
2. **At runtime**: Fail as fast as possible if a value is assigned to it (either on setting this member if that's possible, or when these client options are supplied to us)

i.e. here, something like:

```dart
@Deprecated('Not used, as log messages are handled by the default mechanism in the underlying SDK. This instance variable will be removed in future.') LogHandler? logHandler;
```

This might just be a gentler approach for those who get automatically upgraded to the version of the SDK that incorporates this change because it's come in via SemVer patch.

QuintinWillison (commented):
Just a thought. What do you / others think?

lib/src/authentication/src/client_options.dart
ikurek: @QuintinWillison I thought about this, and there's a reason I didn't do this - I think introducing deprecation for a feature that never actually worked may lead to more confusion than simply removing it. If you prefer the deprecation approach in cases like that, I'll restore this and add `@deprecated` warning as you requested 🙂

igorkurek: @QuintinWillison It's done in https://github.com/ably/ably-flutter/commit/19de88ca1fdb39cc236fd6e71f7e4fae157ae89c 🙂 I've also added a deprecation warning to the `LogHandler` type itself

ikurek: @QuintinWillison It's done in https://github.com/ably/ably-flutter/commit/19de88ca1fdb39cc236fd6e71f7e4fae157ae89c 🙂 I've also added a deprecation warning to the `LogHandler` type itself

QuintinWillison: Thanks. I have at least one follow-up comment but will submit that in a second review pass.

lib/src/authentication/src/client_options.dart
```diff
@@ -21,7 +21,15 @@ class ClientOptions extends AuthOptions {
...
   /// Custom log handler
   ///
+  /// For discussion about removing this component see
+  /// https://github.com/ably/ably-flutter/issues/238
+  ///
   /// https://docs.ably.com/client-lib-development-guide/features/#TO3c
+  @Deprecated(
+    'Not used, as log messages are handled by the default mechanism '
+    'in the underlying SDK. This instance variable will be removed '
+    'in future.',
```
QuintinWillison, line 31:
```suggestion
    'in a future release.',
```

lib/src/logging/src/log_handler.dart
```diff
@@ -2,7 +2,14 @@ import 'package:ably_flutter/ably_flutter.dart';
 
 /// Custom handler to handle SDK log messages
 ///
+/// For discussion about removing this component see
+/// https://github.com/ably/ably-flutter/issues/238
+///
 /// https://docs.ably.com/client-lib-development-guide/features/#TO3c
+@Deprecated(
+  'All usages of this type are marked as deprecated '
+  'so it will be removed in future releases',
```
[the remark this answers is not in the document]
ikurek: Sorry, my English isn't perfect, I appreciate you checked those messages 🙂 I commited a fix in https://github.com/ably/ably-flutter/pull/340/commits/a0d22cab50f7fa3781054cc118b78385f0fa3eea

lib/src/authentication/src/client_options.dart
ikurek: Fixed in https://github.com/ably/ably-flutter/pull/340/commits/a0d22cab50f7fa3781054cc118b78385f0fa3eea

lib/src/logging/src/log_handler.dart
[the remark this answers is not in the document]
QuintinWillison: You're English is excellent, @ikurek, you know that! 😉 ... there will always be some things that only someone with English as first language will spot. Team work. 😁

ikbalkaya (approved):
Quickly scanned through changes, all looks good to me

lukasz-szyszkowski (approved):

Issue #232, opened 2021-11-18
Fix flaky Android emulators


ben-xD:
CI jobs are randomly failing with `Original error: ext.flutter.driver: (112) Service has disappeared`. We need to investigate further. 

For an example, see https://github.com/ably/ably-flutter/runs/4241734082?check_suite_focus=true

QuintinWillison:
Also seen [here (Android 21)](https://github.com/ably/ably-flutter/runs/4807954432?check_suite_focus=true) and [here (Android 29)](https://github.com/ably/ably-flutter/runs/4807954468?check_suite_focus=true).

I notice that we have a comment in our [integration testing workflow](https://github.com/ably/ably-flutter/blob/main/.github/workflows/flutter_integration.yaml):

> ubuntu-latest cannot be used as it is only a docker container, and unfortunately running hardware acceleration in container is not allowed by android emulator.

Which immediately makes me ask the question "do we need hardware acceleration?". Perhaps we can opt out of that in order to have tests run more reliably... macOS hosts are always less available also, due to GitHub limits.

ikurek:
@QuintinWillison It's a matter of performance - without hardware acceleration the emulator is very slow, so the tests will take longer co complete. I've seen this issue raised multiple times on different Ci platforms and usually it's best to stick with MacOS, even the CI step developer has an open issue about this: https://github.com/ReactiveCircus/android-emulator-runner/issues/46

QuintinWillison:
@ikurek Sorry, I'm only just catching up with some notifications so seen [your comment](https://github.com/ably/ably-flutter/issues/232#issuecomment-1035106323). We've been successfully using `ubuntu-latest` for just over a month in [`ably-java`'s emulation workflow](https://github.com/ably/ably-java/blob/main/.github/workflows/emulate.yml), and my preference (at the current time) is for us to switch all Android virtual device emulation runs to this configuration in order to reduce contention for limited availability of macOS hosts.

To be really clear:

- **Our priority is not performance or speed.**
- **Our priority has to be test stability.**

There is a possible plan for us to do something to upgrade our capacity to support more simultaneous macOS hosts, but it's not come to anything yet. FYI, @jaley. So, in the meantime, let's switch to Ubuntu and get some more predictability for workflow runs.

PR #341, opened 2022-02-28
Switched CI integration tests to run on ubuntu instead of macos

ikurek:
This is related to #232 but I'm not sure if it will solve the problem entirely. So I'm not closing the issue for now and I'll watch the CI for a while to see if the errors still occur.

Besides switching from `macos` runner to `ubuntu` I've also added a few additional params to the virtual machine (copy/paste from [`ably/java`'s integration test config](https://github.com/ably/ably-java/blob/main/.github/workflows/emulate.yml) but it should be fine for our use case).

### DIFF
--- a/.github/workflows/flutter_integration.yaml
+++ b/.github/workflows/flutter_integration.yaml
@@ -61,6 +61,5 @@ jobs:
           api-level: ${{ matrix.api-level }}
           emulator-options: -no-snapshot-save -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim
           script: |
-            /Users/runner/Library/Android/sdk/tools/bin/avdmanager list
             flutter pub get
             cd test_integration && ./run_integration_tests.sh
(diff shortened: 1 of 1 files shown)


.github/workflows/flutter_integration.yaml
```diff
@@ -61,7 +59,7 @@ jobs:
         uses: reactivecircus/android-emulator-runner@v2
         with:
           api-level: ${{ matrix.api-level }}
+          emulator-options: -no-snapshot-save -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim
           script: |
-            /Users/runner/Library/Android/sdk/tools/bin/avdmanager list
```
ikbalkaya, line 65: Is there a reason you removed this 🤔

ikurek: Yes, two reasons to be exact:

1. Because of the switch from `macOS` to `ubuntu` this `avdmanager` path is no longer valid, to he command would fail because of missing `avdmanager`
2. The whole call to `avdmanager list` does nothing useful - it only prints the list of available Android virtual devices to the logs and I failed to find any purpose of it in our CI flow

So I think it's a good idea to simply get rid of it 🙂

QuintinWillison (approved):

ikbalkaya (approved):

Issue #318, opened 2022-02-07
Example App should be able to build and run without Firebase configured via `google-services.json`


QuintinWillison:
Friction is _way_ too high for those who wish to test the SDK without support for push notifications.

Relates to #316.

I observed on my local machine:

```
example % flutter build apk

💪 Building with sound null safety 💪


FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:processReleaseGoogleServices'.
> File google-services.json is missing. The Google Services Plugin cannot function without it.
   Searched Location:
  /Users/quintinwillison/code/ably/ably-flutter/example/android/app/src/release/google-services.json
  /Users/quintinwillison/code/ably/ably-flutter/example/android/app/src/google-services.json
  /Users/quintinwillison/code/ably/ably-flutter/example/android/app/src/Release/google-services.json
  /Users/quintinwillison/code/ably/ably-flutter/example/android/app/google-services.json


* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org/

BUILD FAILED in 56s
Running Gradle task 'assembleRelease'...                           56.8s
Gradle task assembleRelease failed with exit code 1
```

@ikurek [responded (internal Slack message) that](https://ably-real-time.slack.com/archives/C0178QGC093/p1644237008457189?thread_ts=1644236687.002689&cid=C0178QGC093):

> I agree, but sadly there’s no easy way to solve this - as long as we embed Firebase into the project, this file will be required with each start. For now the only options I can see are:
> - Include some sample google-services.json file attached to example Firebase instance (I don’t think it’s the best idea TBH)
> - Create separate example for push notifications

I wonder if there's another approach, given how flexible Gradle is, where the build can detect if `google-services.json` is present and only enable the Google Services Plugin / Firebase complexities in that case. 🤔

PR #344, opened 2022-03-04
Sample app default google-services file

ikurek:
This should resolve https://github.com/ably/ably-flutter/issues/318 and make the sample app work without any additional configuration. It;s based on `feature/example-app/provisioning` because they both alter sample application and readme files and I wanted to avoid conflicts.

The most important change is the dummy `google-services.json` file. I've decided to use this approach, since it seemed to be the most reasonable one. File structure is compliant with Firebase requirements, but it doesn't contain any data inside. This is just enough to run the app, as long as no Firebase services are needed. 

Google states that these files do not contain any sensitive information, but I'm not a fan of putting any server instance data or API keys inside the repository. In our use case, it doesn't really matter if Firebase is able to connect in the sample app or not, since it won't be able to connect to push services anyway, because Firebase instance has to be linked with Ably instance first, and that can't be done with any sample instance of Firebase that we would use.

I've also added error popups for the sample app, to indicate errors related to push notifications.

### DIFF
--- a/PushNotifications.md
+++ b/PushNotifications.md
@@ -39,7 +39,7 @@ To get push notifications setup in your own app, read [Setting up your own app](
 - Create a Firebase project, and in the Project settings, add an android app with your unique application ID. Follow the steps provided on the setup process, or the following:
     - You can leave `Debug signing certificate SHA-1` empty.
     - Download the generated `google-services.json` file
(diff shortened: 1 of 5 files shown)


KacperKluka (approved):
LGTM :+1:

Issue #61, opened 2020-11-09
Use named arguments for options classes and make all fields private


tiholic:
Python allows kwargs to pass all the possible options to create a ClientOptions object. With flutter, we need to have properties of all such objects (`ClientOptions`, `ChannelOptions`, etc) as public and SDK users need to instantiate the class and then set these properties.

So, as per the current implementation:
```dart
//usage 1
final options = ClientOptions();
options.environment = 'sandbox';
options.logLevel = LogLevel.verbose;
```
```dart
//usage 2
final options = ClientOptions()
    ..environment = 'sandbox'
    ..logLevel = LogLevel.verbose;
```

It would be more intuitive to dart developers if we can consider changing the constructors to use named arguments and make all the fields as private.

With the named arguments approach, the usage would be:
```dart
final options = ClientOptions(
  environment: 'sandbox',
  logLevel: LogLevel.verbose,
);
```

PR #348, opened 2022-03-18
Refactor of ClientOptions

ikurek:
This is a part of #61, including changes only for `ClientOptions` and `AuthOptions`, without moving the fields as private. Main changes:
* Updated constructor of `AuthOptions` and `ClientOptions` to use named parameters
* Deprecated `fromKey` methods, since the object can be now created with constructor invocation
* Refactored all places where `ClientOptions` was created with public parameter setters 

Although this PR is fairly large, most of the changes were done in integration testing, and none of the changes are breaking in any way - users should be able to migrate without any issues. Moving the variables as private is a breaking change, so I decided to implement this later, in a separate PR, to give users some more time to adopt these changes before we deprecate the public fields

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -337,7 +337,7 @@ client libraries.
 ### Supported functionality
 
 - Create a REST or Realtime instance by passing `ClientOptions`:
-    - `ClientOptions` can be created by passing an API token (`ClientOptions(key: key)`)
+    - `ClientOptions` can be created by passing an API token (`ClientOptions.fromKey`)
     - `defaultTokenParams`, `authCallback` and `logHandler` are not supported yet
 - Get a REST channel and publish messages
 - Listen for Realtime connection state changes using a stream subscription
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -337,7 +337,7 @@ client libraries.
 ### Supported functionality
 
 - Create a REST or Realtime instance by passing `ClientOptions`:
-    - `ClientOptions` can be created by passing an API token (`ClientOptions.fromKey`)
+    - `ClientOptions` can be created by passing an API token (`ClientOptions(key: key)`)
     - `defaultTokenParams`, `authCallback` and `logHandler` are not supported yet
 - Get a REST channel and publish messages
 - Listen for Realtime connection state changes using a stream subscription
--- a/lib/src/authentication/src/client_options.dart
+++ b/lib/src/authentication/src/client_options.dart
@@ -4,14 +4,6 @@ import 'package:ably_flutter/ably_flutter.dart';
 ///
 /// https://docs.ably.com/client-lib-development-guide/features/#TO1
 class ClientOptions extends AuthOptions {
-  /// Set fields on [ClientOptions] to configure it.
-  ClientOptions();
-
-  /// initializes [ClientOptions] with a key and log level set to info
-  ///
-  /// See [AuthOptions.fromKey] for more details
-  ClientOptions.fromKey(String key) : super.fromKey(key);
-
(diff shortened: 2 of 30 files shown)


KacperKluka (changes requested):
Overall everything looks good but I think that one thing is missing - update in the README and Ably docs :wink: Since we've deprecated the `fromKey` I believe we shouldn't suggest it in our usage guides.

I took a quick look and found at least 3 places (but there may be more) in which we're using the deprecated `fromKey` approach:
- project [README](https://github.com/ably/ably-flutter#configure-a-client-options-object) (let's update it as a part of this PR)
- [Ably quickstart guide](https://ably.com/documentation/quick-start-guide?lang=flutter#step-2) (probably we should inform the docs team about it, what's the best way to do this @QuintinWillison ? :thinking: )
- [plugin details on pub.dev](https://pub.dev/packages/ably_flutter) (probably this is taken from the README and will be updated automatically with the next release, right?)

KacperKluka (approved):
LGTM now, thanks :wink: The only thing left is the `fromKey()` reference in the Ably docs but this is outside of this PR scope, as the change should probably be made after this PR is released in a new `ably-flutter` version :thinking:

CHANGELOG.md
```diff
@@ -337,7 +337,7 @@ client libraries.
 ### Supported functionality
 
 - Create a REST or Realtime instance by passing `ClientOptions`:
-    - `ClientOptions` can be created by passing an API token (`ClientOptions.fromKey`)
+    - `ClientOptions` can be created by passing an API token (`ClientOptions(key: key)`)
```
QuintinWillison, line 340: We shouldn't be modifying old change log entries, unless they were wrong at the point of release they were made for (in this case, [release tag `1.0.0+dev.1`](https://github.com/ably/ably-flutter/releases/tag/1.0.0%2Bdev.1)).

lib/src/authentication/src/client_options.dart
```diff
@@ -195,6 +187,83 @@ class ClientOptions extends AuthOptions {
   /// https://docs.ably.com/client-lib-development-guide/features/#TO3l7
   int channelRetryTimeout = 15000;
 
+  /// Initializes an instance with defaults
+  ClientOptions({
```
QuintinWillison, line 191: If the order of these parameters is arbitrary (that is, given they are named, their order carries no meaning) then I wonder if we should consider sorting them alphabetically. 🤔 

On a quick search I can't see any evidence there are any lint rules that can enforce this for us, so perhaps not worth it. Just an idea.

QuintinWillison (changes requested):
Looks good to me, albeit admitting I have skim read much of the contents.

CHANGELOG.md
ikurek: Sorry, I ran search & replace across all MD files and I missed that this one was in CHANGELOG 🤦 Reverted this in https://github.com/ably/ably-flutter/pull/348/commits/155469282583605d8d9340ca22adbc133bc90b48

lib/src/authentication/src/client_options.dart
ikurek: I don't think there's a lint rule for that, as far as I know Dart doesn't define this, but I think it may be a nice idea to have them sorted in all class constructors. However, I'd rather not do it in scope of this PR 🙂

QuintinWillison: Fair enough and a good call. Perhaps create an issue for this, @ikurek, as a potential code quality improvement for the backlog?

ikurek: It's created: https://github.com/ably/ably-flutter/issues/351 🙂

Issue #349, opened 2022-03-21
Fix AblyException assigning 'Message' to 'Code' field


ikurek:
`AblyException` class has a constructor that looks like this:

```dart
AblyException([
    this.code,
    this.message,
    this.errorInfo,
]);
```

Which specifies all parameters as optional, positional parameters. This caused a bit of confusion in the codebase - it's not possible to create an instance of `AblyException` with `message` set, if no `code` is set. This led to constructor calls like this:

```dart
AblyException(
    'Platform communication error. Response cannot be null for $method',
)
```

Which returns an object like this:

```dart
code = 'Platform communication error. Response cannot be null for $method',
message = null,
errorInfo = null
```

It would be good to fix this issue and unify construction of `AblyException` objects, so users can use the 'message` or `code` fields without any additional checks, and there's no need to provide `code` if we want to provide a `message`.

PR #350, opened 2022-03-21
Updated AblyException constructor to avoid argument missmatches

ikurek:
This should close https://github.com/ably/ably-flutter/issues/349 and fix the construction of `AblyException` instances. Changes include:
* Refactoring `AblyException` constructor to use named parameters instead of optional positional parameters
* Updating invocations of `AblyException` constructor to assign message contents to `message` field
* Removal of unnecessary `late` modifier on `AblyException.errorInfo`


KacperKluka (approved):
LGTM :+1:

Issue #179, opened 2021-09-21
Avoid additional wrapper types for platform method invocations


ben-xD:
# AblyMessages are sometimes nested, and sometimes not: 
```
  final message = (null != argument)
      ? AblyMessage(AblyMessage(argument, handle: _handle))
      : AblyMessage(_handle);
```
In fact, the concept of AblyMessage is not actually necessary. I suggest we remove this completely. No other plugins have a concept of data type which wraps the JSON serialization format, they just use the JSON/ map. 

This AblyMessage behaviour is a source of confusion/ complexity when writing new platform methods.

QuintinWillison:
The opening comment for this issue refers to using the 'JSON / map'... on that particular point, I think it's worth bearing in mind the mistake (IMO) that was made in the current `ably-java` API design whereby we expose a third-party API (GSON) as part of our API. See: https://github.com/ably/ably-java/issues/573

I think we should avoid making that kind of mistake in our Flutter plugin API. Therefore, my preference would be for whatever we expose instead of `AblyMessage` (if that's the decision we make) to be something Dart-native (i.e. perhaps preferring to present raw JSON as bytes or string, or perhaps parsing it to a Map - if that even makes sense in the strongly typed world of Dart 🤔).

PR #357, opened 2022-03-31
Experimental method to reset push device on Android

ikurek:
This PR adds an experimental method `Push.reset()` that clears the state of `ActivationStateMachine` for Push Notifications. This method can be used to solve some issues with activating push devices in Ably service, but since it's an experimental solution it shouldn't be used without a good reason.

Internally, the `reset()` method call clears `SharedPreference` storage and restores `ActivationStateMachine` to its initial state, removing all local device data and making it possible to perform a clean activation of device.

### DIFF
--- a/android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java
+++ b/android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java
@@ -682,7 +682,6 @@ public class AblyMethodCallHandler implements MethodChannel.MethodCallHandler {
       Integer handle = (Integer) message.message;
       instanceStore.getPush(handle)
               .getActivationContext()
-              .getActivationStateMachine()
               .reset();
     } catch (AblyException e) {
       handleAblyException(result, e);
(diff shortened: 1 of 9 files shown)


android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java
```diff
@@ -675,6 +676,18 @@ private void pushDeactivate(@NonNull MethodCall call, @NonNull MethodChannel.Res
     }
   }
 
+  private void pushReset(@NonNull MethodCall call, @NonNull MethodChannel.Result result) {
+    final AblyFlutterMessage message = (AblyFlutterMessage) call.arguments;
+    try {
+      Integer handle = (Integer) message.message;
```
KacperKluka, line 682: The names are quite vague :sweat_smile: And the `message.message` could be improved :wink: Could you think of some better names? :thinking:

KacperKluka (approved):
Overall it looks good and I don't want to block the PR but if you'd like to I've left a suggestion :wink:

android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java
ikurek: I'm aware of that, and already working on it in scope of #179, but it's a large change, so for now I decided to stick to previous convention which missmatches `handle` and `message` values in multiple cases 😞

QuintinWillison (approved):

Issue #359, opened 2022-04-05
Use consistent type for platform method parameters


ikurek:
Currently, calls to platform methods can take any data type as an argument, for example:

`Push.deactivate()` doesn't pass any params:
```dart
Future<void> deactivate() => invoke(PlatformMethod.pushDeactivate);
```

`Crypto.generateRandomKey()` passes `int` as an argument:
```dart
static Future<Uint8List> generateRandomKey({keyLength = defaultKeyLengthInBits}) => 
    Platform().invokePlatformMethodNonNull<Uint8List>(PlatformMethod.cryptoGenerateRandomKey, keyLength);
```

`Push.subscribeDevice()` passes a `Map<String, dynamic>`, although it only has one argument inside:
```dart
Future<void> subscribeDevice() => invoke(
    PlatformMethod.pushSubscribeDevice, {TxTransportKeys.channelName: _name});
```

I think it would be best to unify those and make methods like `invoke()`, `invokeRaw()` or `invokePlatformMethodNonNull()` always use one format for arguments, f.e. `Map<String, dynamic>` since it seems to be the most straightforward solution for Dart. I think it would make handling platform methods much easier on Android and iOS side, if arguments are always passed in the same, consistent format

PR #360, opened 2022-04-05
AblyMessage fixes and refactor

### LINKED ISSUE #179: Avoid additional wrapper types for platform method invocations

ikurek:
This PR closes #179, it's a rather large change to the codebase, but it should help us identify issues easier in the future, and reduce complexity of the codebase. Changes include:

* Dart side:
    * Refactored `AblyMessage`, to avoid unnecessary nesting of objects. Previously, `AblyMessage` would often wrap another `AblyMessage` instance which required additional checks on iOS and Android side, since it was never really known if the message was wrapped or not.
    * Refactored `AblyMessage`, to always use `AblyMessage.handle` field for passing `handle` values. Previously, `handle` values were often passed as `AblyMessage.message` leaving `AblyMessage.handle` set to null, which made no sense and introduces additional checks in platform code (we had to check  in witch field the `handle` value is present)
    * Introduced constructors with named parameters to `AblyMessage` together with an `empty()` constructor to create instances with empty parameter collections
    * Changed platform method callers like `invoke()` or `invokeRaw()` to always provide `AblyMessage` instances to platform method calls. This should prevent errors with parsing and casting method calls on Android and iOS. By default, each method will now provide and empty map of parameters if they are not defined anywhere
* Android side
    * Refactored `AblyMethodCallHandler`, removing unnecessary nested message checks. I've also unified type of values for `handle`, which now always uses `long`, so checks and casts for long/int/Number types were also removed.
    * Removed `ablyDo` lambda from `AblyMethodCallHandler`. The lambda was only used to unwrap nested `AblyFlutterMessage` instances and cast message types, since I removed the wrapping, it became obsolete
* iOS side
    * Refactored `AblyFlutter`, removing unnecessary nested message checks
    * Refactored `PushHandlers`, removing unnecessary nested message checks
    
This is a first step in unifying the communication between Flutter and platform SDK's. I think the next change in that matter would be https://github.com/ably/ably-flutter/issues/359


KacperKluka (approved):
I've skimmed that huge java class and omitted iOS files but overall it looks good and I think it changes the codebase for the better :+1:

lukasz-szyszkowski (approved):

QuintinWillison (approved):

PR #362, opened 2022-04-05
Simpler hashCode implementation for Message-related classes

### LINKED ISSUE #132: Conform hashCode implementations
Avoid unnecessary steps like concatenating component hash codes as a string to then convert back to hash code again... this should all be basic maths.

Triggered by [this comment](https://github.com/ably/ably-flutter/pull/123#discussion_r650793251).

ikurek:
This should close #132. I've removed the original implementations of `hashCode` and replaced it with a single implementation that calculates hash from values of provided objects.


KacperKluka (approved):
Looks good :+1: I'm assuming that the hash code was not used in any kind of persistence as the new version of hash code calculation will produce different results than the old one :wink:

Issue #120, opened 2021-05-17
Update errorReason on Realtime.connection


tiholic:
`ErrorReason` is not being updated currently along with `ConnectionState`

PR #365, opened 2022-04-08
Updated realtime connection to update errorReson

ikurek:
Should close https://github.com/ably/ably-flutter/issues/120. The task isn't too descriptive, but I have figured that it's related to `Connection` class not updating the `errorReason` field, when it's state is changed. This small change should update the error every time the state of connection is updated


KacperKluka (approved):
LGTM :+1:

QuintinWillison (approved):

Issue #83, opened 2021-02-15
Enum for direction


tiholic:
`direction` `string` used in `RestHistoryParams` and similar query filter param classes can be an enum which can be encoded to a string while passing on to platform side.

PR #366, opened 2022-04-08
Enum for Direction parameter in History

ikurek:
This PR should resolve https://github.com/ably/ably-flutter/issues/83. I've replaced previously used `String` value with an enum shared between `RealtimeHistoryParams` and `RestHistoryParams`. Sadly, this is a breaking change, because it's not possible to deprecate unnamed constructors in Dart, but since it's a rather small difference, I don't think it should cause too many issues. I've left a note about this in UPDATING file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -237,7 +237,7 @@ void getHistory([ably.RestHistoryParams params]) async {
 getHistory();
 
 // sorted and filtered history
-getHistory(ably.RestHistoryParams(direction: 'forwards', limit: 10));
+getHistory(ably.RestHistoryParams(direction: HistoryDirection.forwards, limit: 10));
 ```
 
 Get REST Channel Presence:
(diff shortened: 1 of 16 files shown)


KacperKluka (approved):
LGTM

QuintinWillison (approved):
Recognising that this is a breaking change to users, however we're not going to be properly conforming to semantic versioning until next month (see https://github.com/ably/engineering/pull/27), therefore we accept we may need to handle support queries/complaints relating to this in the meantime and once this has been released in the next `patch` release.

FYI, @stmoreau @AndyNicks

stmoreau:
As these are breaking changes, is there a critical need to merge these changes until we're able to do a `major` release?

ikurek:
> As these are breaking changes, is there a critical need to merge these changes until we're able to do a `major` release?

It's nothing really urgent, so I guess we can wait. On the other hand, it's a small change and it should be relatively easy for users to adapt to it.

ikurek:
@stmoreau @QuintinWillison I've changed the base of this PR to `integration/version-2`, so we can merge breaking changes in the next major version

Issue #84, opened 2021-02-15
Lint rule to avoid for implicit dynamic variables


tiholic:
See TODO: https://github.com/ably/ably-flutter/blob/fd15cba20d3ec4f9768df14b10b1ad88e4a4c174/analysis_options.yaml#L11

Ref: https://github.com/dart-lang/sdk/issues/33749

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-633) by [Unito](https://www.unito.io)

PR #375, opened 2022-04-21
Lint rule to disable implicit dynamic values

ikurek:
This should solve https://github.com/ably/ably-flutter/issues/84. The issue was raised a while ago, and the `implicit dynamic` option was already deprecated, but it was replaced with `strict type checks`. The changes here are mostly cosmetic, but overall they remove ~200 lint warnings in the codebase and remove implicit dynamic typing, which is discouraged in Dart for a while. The changes include:

* Enabled `strict-casts` and `strict-inference` lint rules, as a replacement for `implicit-dynamic`. Also, removed `avoid_annotating_with_dynamic` and `avoid_types_on_closure_parameters` since they conflict with `strict-inference`
* Added declared types to all methods that used implicit dynamic values
* Widened the scope of `dynamic` variables where it was possible. There were some cases where the variable/method did not need `dynamic` at all, as it's type was declared during asignement
* Removed unnecessary cast methods like `transformState` or `transformStateChange`. With declared variable types, there's no need to use methods that cast variables to their type in cases like this, a simple `as` cast is enough, and reduces a lot of boilerplate code
* Added `void` return types for `Future` methods that do not return any value. Previously these methods returned `dynamic` which was always a `void` type

### DIFF
--- a/README.md
+++ b/README.md
@@ -620,7 +620,7 @@ However, using a neat little workaround will fix this...
 Instead of `await subscriptionToBeCancelled.cancel();`, use
 
 ```dart
-Future.delayed(Duration.zero, () {
+Future<void>.delayed(Duration.zero, () {
     subscriptionToBeCancelled.cancel();
 });
 ```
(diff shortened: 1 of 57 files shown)


lukasz-szyszkowski (approved):

KacperKluka (approved):
Looks good (after a skim-read) :+1:

Issue #87, opened 2021-02-15
Add tests for history API with authcallback


tiholic:
`rest` and `realtime` `history` are missing integration tests with `authCallback`

See:

https://github.com/ably/ably-flutter/blob/5134b09e90e917f7abd234ee55e02e9ad01500a1/test_integration/lib/test/test_names.dart#L18

https://github.com/ably/ably-flutter/blob/5134b09e90e917f7abd234ee55e02e9ad01500a1/test_integration/lib/test/test_names.dart#L33

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-635) by [Unito](https://www.unito.io)

PR #388, opened 2022-05-10
Integration tests for History with AuthCallback

ikurek:
This should close https://github.com/ably/ably-flutter/issues/87. I've added new tests to cover the cases of fetching history when AuthCallback is used in `ClientOptions`

### DIFF
--- a/test_integration/lib/config/test_names.dart
+++ b/test_integration/lib/config/test_names.dart
@@ -16,11 +16,12 @@ class TestName {
   static const String restEncryptedPublishSpec = 'restEncryptedPublishSpec';
   static const String restCapabilities = 'restCapabilities';
   static const String restHistory = 'restHistory';
+  static const String restHistoryWithAuthCallback =
+      'restHistoryWithAuthCallback';
   static const String restTime = 'restTime';
   static const String restPublishWithAuthCallback =
       'restPublishWithAuthCallback';
 
-  // TODO(tiholic) handle restHistoryWithAuthCallback
   static const String restPresenceGet = 'restPresenceGet';
   static const String restPresenceHistory = 'restPresenceHistory';
 
--- a/test_integration/lib/config/test_names.dart
+++ b/test_integration/lib/config/test_names.dart
@@ -3,53 +3,57 @@
 /// dependencies and can be imported in driver test files.
 
 class TestName {
-  static const String platformAndAblyVersion = 'platformAndAblyVersion';
+  // platform and key
   static const String appKeyProvisioning = 'appKeyProvisioning';
+  static const String platformAndAblyVersion = 'platformAndAblyVersion';
 
+  // crypto
+  static const String cryptoEnsureSupportedKeyLength =
+      'cryptoEnsureSupportedKeyLength';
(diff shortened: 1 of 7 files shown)


test_integration/lib/config/test_names.dart
```diff
@@ -3,51 +3,57 @@
...
+  static const String restTime = 'restTime';
+
+  // helpers
+  static const String testHelperFlutterErrorTest = 'testHelperFlutterErrorTest';
+  static const String testHelperUnhandledExceptionTest =
+      'testHelperUnhandledExceptionTest';
 
-  // This is not a test, but a way to retrieve
   // more information of failures from any of the tests cases
+  // This is not a test, but a way to retrieve
```
KacperKluka, line 57: Now the comment lines are in the inverted order :sweat_smile: Please revert it to how it was before :wink:

test_integration/lib/test/realtime/realtime_history_with_auth_callback_test.dart
```diff
@@ -0,0 +1,36 @@
...
+
+  final channel = realtime.channels.get('test');
+  await publishMessages(channel);
+
+  await channel.history();
+  await Future<void>.delayed(TestConstants.publishToHistoryDelay);
+
+  return {
+    'handle': await realtime.handle,
+    'authCallbackInvoked': authCallbackInvoked,
```
KacperKluka, line 34: Instead of checking if the auth callback was invoked maybe we should check if the `channel.history()` has returned the correct result? Or maybe both this and the auth callback? :thinking: I'm wondering what exactly do we want to test here :wink:

test_integration/lib/test/rest/rest_history_with_auth_callback_test.dart
```diff
@@ -0,0 +1,36 @@
...
+
+  final channel = rest.channels.get('test');
+  await publishMessages(channel);
+
+  await channel.history();
+  await Future<void>.delayed(TestConstants.publishToHistoryDelay);
+
+  return {
+    'handle': await rest.handle,
+    'authCallbackInvoked': authCallbackInvoked
```
KacperKluka, line 34: Like above :wink:

KacperKluka (changes requested):

test_integration/lib/test/realtime/realtime_history_with_auth_callback_test.dart
ikurek: There's an entirely separate suite of tests checking correctness of history in [rest_history_test.dart](https://github.com/ably/ably-flutter/blob/main/test_integration/lib/test/rest/rest_history_test.dart). These tests are only supposed to check if `authCallback` is executed when specific rest/realtime methods are called. The structure of tests like this already exists in a project, for example in [rest_publish_with_auth_callback.test](https://github.com/ably/ably-flutter/blob/main/test_integration/lib/test/rest/rest_publish_with_auth_callback_test.dart), I only added a missing one for history.

KacperKluka (approved):

QuintinWillison (approved):
I've merely skim read the contents of this PR for lack of time, but sounds sensible and more tests are _always_ a good thing. Thanks!

PR #392, opened 2022-05-11
Use GitHub OIDC for AWS access to builds S3 bucket

lmars:
Inline with the change made to `ably/sdk-upload-action` in https://github.com/ably/sdk-upload-action/pull/37.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -8,6 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       deployments: write
+      id-token: write
     steps:
       - uses: actions/checkout@v2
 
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yml
```diff
@@ -26,10 +27,16 @@ jobs:
...
-      - uses: ably/sdk-upload-action@v1
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-region: eu-west-2
+          role-to-assume: arn:aws:iam::884096606738:role/ably-sdk-builds-ably-flutter
+          role-session-name: "${{ github.run_id }}-${{ github.run_number }}"
+
+      - name: Upload Documentation
+        uses: ably/sdk-upload-action@v1.2.2
```
owenpearson, line 38: Any reason to lock in to a patch version here? Surely just leaving it at v1 will give the same behaviour?

owenpearson (changes requested):

owenpearson (approved):
LGTM 👍

PR #393, opened 2022-05-12
Release/1.2.14

ikurek:


ikurek:
I've switched to draft with this release, due to Flutter 3 being released just now. I'd like to check compatibility with it before we release a new version of flutter

ikurek:
Blocked until https://github.com/ably/ably-flutter/issues/394 is resolved

QuintinWillison (approved):
As a minor comment, though we don't need to block the release on it, we're trying to thin down change log entries so that they don't include items which have no direct impact on customers. i.e. we only want to include things that developers might need to be aware of.

So things like test changes and internal refactors do not need to be included.

ikurek:
> Tthings like test changes and internal refactors do not need to be included.

I've removed some entries in https://github.com/ably/ably-flutter/pull/393/commits/6c801db8aa1f683319361f35b14c8c9a5fc40644

Issue #394, opened 2022-05-12
Check requirements and update for Flutter 3.0.0


ikurek:
Flutter 3 was just released, and `ably-flutter` is not compatible with it. The SDK constraint in doesn't allow versions above 2:

https://github.com/ably/ably-flutter/blob/45f6fa4dda5e2471a434cbeddb0f3d6036058da1/pubspec.yaml#L7

And even after it's changed to support version 3, the `pub get` command fails with this output:

```
[ably-flutter] flutter pub get
Running "flutter pub get" in ably-flutter...                    
Because test 1.19.5 depends on test_api 0.4.8 and every version of flutter_test from sdk depends on test_api 0.4.9, test 1.19.5 is incompatible with flutter_test from sdk.
So, because ably_flutter depends on both flutter_test from sdk and test 1.19.5, version solving failed.
pub get failed (1; So, because ably_flutter depends on both flutter_test from sdk and test 1.19.5, version solving failed.)
```

We should update required dependencies and release a fix, to allow builds on latest Flutter versions

PR #398, opened 2022-05-13
Fix for dependency conflicts between Flutter versions

ikurek:
This should close https://github.com/ably/ably-flutter/issues/394. Flutter 3.0.0 was released yesterday, and Ably SDK wasn't compatible with it. I've investigated the issues and here are the changes:

* Updated the minimum required version of Flutter for `ably-flutter` package from `2.2.3` to `2.5.0`. This would usually be a breaking change, but for `2.2.3` the library wouldn't work anyway, because `2.2.3` includes Dart `2.13.4` and for package dependencies Dart `2.14.0` is required. Trying to build the package with `2.2.3` will fail due to dependency constraints.
* Set a minumum required Flutter version for the example app to `2.8.0`. Previously it wasn't set at all, and  due to its construction, example app won't work on any Flutter version below `2.8.0`. Below this, it will fail on Android devices due to missing values in Android Manifest
* Added an override for test dependencies - it was required, because latest version of Flutter conflicts with `test` package. I've created an issue to investigate if it's possible to simplify test dependencies: https://github.com/ably/ably-flutter/issues/397

This should make the library compatible with all versions of Flutter from `2.5.0` to `3.0.0`.


KacperKluka (approved):

pubspec.yaml
```diff
@@ -18,11 +18,20 @@ dev_dependencies:
   flutter_lints: ^1.0.4
   flutter_test:
     sdk: flutter
+  test: ^1.21.1
 
-  # Specifying a slightly older version minimum for the 'pure' Dart test package
-  # because the flutter_test package needs an older version of test_api.
-  test: 1.19.5
-
+dependency_overrides:
```
QuintinWillison, line 23: I notice you are pinning exact versions here, rather than using `^` or `>=` syntax to be less specific and support automatic upgrades.

I'm finding it difficult, on a quick Google, to find any good docs on whether that is even possible - but I did find [this video](https://www.youtube.com/watch?v=Nn6kle_DThM) which seems to suggest you can, if that would work for what you're trying to achieve. 🤔

QuintinWillison (approved):
Just one minor comment, which could be fixed up later in another PR if you prefer.

PR #401, opened 2022-05-17
Fix for test dependencies

ikurek:
This should close https://github.com/ably/ably-flutter/issues/396. Previously `ably-flutter` and `test_integration` were dependent on both`flutter_test` and `test`, with misaligned versions. Because `flutter_test` is built on top of `test` using both packages can cause issues. I've removed the `test` dependency entirely and refactored all components that used it. This allows us to remove `dependency_overrides` entirelly and always run tests against current installed of Flutter SDK

PR #403, opened 2022-05-17
Set Flutter CI version to ^3.0.0

ikurek:
Since `ably-flutter` is ready to be used with Flutter 3, I've updated the CI workflows to use stable Flutter versions 3.0.x. This is also a blocker for #401 because latest versions of `test` dependency requires Flutter 3 to run. I've also updated the formatting of YAML files to increase consistency

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -13,21 +13,21 @@ jobs:
 
       - uses: subosito/flutter-action@v2
         with:
-          flutter-version: "3.0.x"
+          flutter-version: '3.0.x'
           cache: true
 
       # This step requires fetch of test_integration packages because flutter format and
       # flutter analyze check test_integration package too
-      - name: "Install dependencies"
+      - name: 'Install dependencies'
         run: |
           flutter pub get
           cd test_integration && flutter pub get
 
-      - name: "Check formatting issues in the code"
+      - name: 'Check formatting issues in the code'
         run: flutter format --set-exit-if-changed .
 
-      - name: "Statically analyze the Dart code for any errors"
-        run: "flutter analyze --no-pub ."
+      - name: 'Statically analyze the Dart code for any errors'
+        run: 'flutter analyze --no-pub .'
 
-      - name: "Run unit tests"
+      - name: 'Run unit tests'
         run: flutter test --no-pub
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -8,26 +8,26 @@ jobs:
   check:
     runs-on: ubuntu-latest
     steps:
-
       - name: Checkout
         uses: actions/checkout@v2
 
       - uses: subosito/flutter-action@v2
         with:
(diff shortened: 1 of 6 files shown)


.github/workflows/check.yaml
```diff
@@ -8,26 +8,27 @@ jobs:
...
     runs-on: ubuntu-latest
     steps:
-
       - name: Checkout
         uses: actions/checkout@v2
 
       - uses: subosito/flutter-action@v2
         with:
+          flutter-version: "3.0.x"
+          channel: "stable"
```
QuintinWillison, line 17: This is the default for this action - see [here](https://github.com/subosito/flutter-action/blob/0c3f14223a08fa950c8a4c00bcfb834e65744135/action.yml#L15).

.github/workflows/check.yaml
```diff
@@ -8,26 +8,27 @@ jobs:
   check:
     runs-on: ubuntu-latest
     steps:
-
       - name: Checkout
         uses: actions/checkout@v2
 
       - uses: subosito/flutter-action@v2
         with:
+          flutter-version: "3.0.x"
```
QuintinWillison, line 16: Is there a reason not to leave this input unspecified, which according to [the docs for this action](https://github.com/subosito/flutter-action/blob/main/README.md#usage), should be the latest release for this channel?

QuintinWillison (changes requested):
There are inputs being configured with default values - please prefer convention over configuration.

Also, what's with the unrelated changes of single quotes to double quotes?

.github/workflows/check.yaml
ikurek: I think the step doesn't check major version changes - our runners did not get an update to Flutter 3, although it's released to stable channel for a while now, and the checks are still run with 2.10.5.

Besides, I think in that case, we may want to pin a major/minor versions on CI - if the runner would update to the latest stable version (3.0.0), all of our CI checks would start failing, so to avoid this, pinning the version seems to be a good idea

.github/workflows/check.yaml
ikurek: I've added those in a separate commit, because I thought it would be better to have both version and channel defined. I removed the commit, so now we only specify version

ikurek:
> Also, what's with the unrelated changes of single quotes to double quotes?

I wanted to align them - my formatters in VS Code and Android Studio always complain that these YAML files aren't properly formatted, so I thought I may make the quotes consistent if I'm changing them anyway. I've removed the changes to quotes, but decided to leave some minor formatting changes (like removing unnecessary whitespace)

ikurek:
I had to do a minor fix for Flutter lint here: https://github.com/ably/ably-flutter/pull/403/commits/ff15f01455a3e7febffd2947e3bd1e3e98444792, so the CI check doesn't fail

QuintinWillison (approved):
Nice! 😁

KacperKluka (approved):
LGTM :+1:

PR #404, opened 2022-05-17
Use a secret to reference the AWS SDK account ID

lmars:
It's not secret information, but makes it easier to migrate resources to a different AWS account in the future (something we'll likely be doing soon).


owenpearson (approved):
Sounds like a good idea 🙂

Issue #149, opened 2021-08-10
Throw errors instead of returning null when serialization/ deserialization fails


ben-xD:
In Java, if serialization/ deserialization fails, instead of propagating the error, it is logged and ignored, and `null` is returned instead. For example:

https://github.com/ably/ably-flutter/blob/21c49b9883bb3cfb451774376f842f87726b61f3/android/src/main/java/io/ably/flutter/plugin/AblyMessageCodec.java#L348-L357

It would be nicer during development to have an exception be thrown when serialization fails. We should consider throwing an exception when serialization/ deserialization fails.

┆Issue is synchronized with this [Jira Story](https://ably.atlassian.net/browse/SDK-1132) by [Unito](https://www.unito.io)

ikurek:
Working link to mentioned comment: https://github.com/ably/ably-flutter/pull/140#discussion_r684479199

PR #409, opened 2022-05-20
Replaced silent serialization errors with exceptions

ikurek:
This closes https://github.com/ably/ably-flutter/issues/149. Previously, serialization of enums and values without assigned encoder/decoder failed silently, returning `null` when proper value was not found. I've replaced these `null` returns with thrown exceptions, so it's clear that an error has occurred, and should be fixed


KacperKluka (approved):

ikbalkaya (approved):
LGTM

PR #421, opened 2022-05-26
Fix missing App ID build error in example app

### LINKED ISSUE #418: Building Example App on Android fails
Building Example App on Android fails with the following error:
> Launching lib/main.dart on Android SDK built for x86 64 in debug mode...
> lib/main.dart:1
> 
> FAILURE: Build failed with an exception.
> 
> * What went wrong:
> Execution failed for task ':app:processDebugGoogleServices'.
> > Missing Google App Id. Please follow instructions on https://firebase.google.com/ to get a valid config file that contains a Google App Id
> 
> * Try:
> Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.
> 
> * Get more help at https://help.gradle.org
> 
> BUILD FAILED in 12s
> Exception: Gradle task assembleDebug failed with exit code 1
> Exited

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-1933) by [Unito](https://www.unito.io)

ikurek:
This should close #418. I've used the [PN test Firebase app](https://console.firebase.google.com/u/0/project/ably-flutter-pn-tests) to get a valid App ID and updated the `google-services.json`


example/android/app/src/google-services.json
```diff
@@ -7,7 +7,7 @@
   "client": [
     {
       "client_info": {
-        "mobilesdk_app_id": "",
+        "mobilesdk_app_id": "1:805978203344:android:351562e0e66401655aaad5",
```
KacperKluka, line 10: Is it safe to commit it to the public repo? Shouldn't this be a secret? :thinking:

ikurek: This is simply an identifier of the app on Firebase, it's not a secret. The whole file can contain some secrets (like `api_key` field), but according to Google, it's safe to keep these files, even with all the data, in the repository. See https://groups.google.com/g/firebase-talk/c/bamCgTDajkw

QuintinWillison (approved):
Is it worth documenting somewhere that, for the included example app, a placeholder has been used for `client.client_info.mobilesdk_app_id`? I'm wondering if there's any chance developers might copy the source verbatim and get confused, perhaps believing that this app id is required to make the solution work, not realising that they need to swap in their own app id. 🤔

ikurek:
@QuintinWillison We already have this documented in [Setting up the example app](https://github.com/ably/ably-flutter/blob/main/PushNotifications.md#setting-up-the-example-app), the docs state that you need to replace the entire file, so I think it should be clear 🙂

ikbalkaya (approved):

Issue #429, opened 2022-06-14
Android application got crash on notification receive in killed state using Ably notifications


Asharulislam:
I am not using firebase_messaging package.

2022-06-15 16:34:16.617 26304-26365/com.appiskey.dang.dev E/AndroidRuntime: FATAL EXCEPTION: pool-12-thread-1
    Process: com.appiskey.dang.dev, PID: 26304
    java.lang.RuntimeException: Methods marked with @UiThread must be executed on the main thread. Current thread: pool-12-thread-1
        at io.flutter.embedding.engine.FlutterJNI.ensureRunningOnMainThread(FlutterJNI.java:1280)
        at io.flutter.embedding.engine.FlutterJNI.setPlatformMessageHandler(FlutterJNI.java:854)
        at io.flutter.embedding.engine.dart.DartExecutor.onAttachedToJNI(DartExecutor.java:87)
        at io.flutter.embedding.engine.FlutterEngine.<init>(FlutterEngine.java:296)
        at io.flutter.embedding.engine.FlutterEngine.<init>(FlutterEngine.java:262)
        at io.flutter.embedding.engine.FlutterEngine.<init>(FlutterEngine.java:242)
        at io.flutter.embedding.engine.FlutterEngine.<init>(FlutterEngine.java:162)
        at io.ably.flutter.plugin.push.ManualFlutterApplicationRunner.<init>(ManualFlutterApplicationRunner.java:52)
        at io.ably.flutter.plugin.push.FirebaseMessagingReceiver.sendMessageToFlutterApplication(FirebaseMessagingReceiver.java:68)
        at io.ably.flutter.plugin.push.FirebaseMessagingReceiver.lambda$onReceive$0$FirebaseMessagingReceiver(FirebaseMessagingReceiver.java:36)
        at io.ably.flutter.plugin.push.-$$Lambda$FirebaseMessagingReceiver$Q7xqX52FAypmyYY-RYC6oZwb4P8.run(Unknown Source:6)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
        at java.lang.Thread.run(Thread.java:923)

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-2050) by [Unito](https://www.unito.io)

Asharulislam:
[...]

Issue #339, opened 2022-02-24
Split specific push providers as separate Flutter plugins


ikbalkaya:
As a follow up to  https://github.com/ably/ably-flutter/issues/292 we need to actually implement this feature
(This issue description will be updated)

PR #431, opened 2022-06-29
Main thread fix

ikbalkaya:
Regression from https://github.com/ably/ably-flutter/pull/291
Closes https://github.com/ably/ably-flutter/issues/429


KacperKluka:
@ikbalkaya are you planning to also fix that ANR issue that is being reintroduced to the SDK with this PR? It would not be good to make a new release that deliberately adds a bug that we've fixed before :sweat_smile:

ikbalkaya:
> @ikbalkaya are you planning to also fix that ANR issue that is being reintroduced to the SDK with this PR? It would not be good to make a new release that deliberately adds a bug that we've fixed before 😅

Yes, it is intended to be resolved when we decouple push notification code from library code as I added in description https://github.com/ably/ably-flutter/issues/339 But for now this bug takes the priority on ANR bug so we want to fix this. More discussion on  [this internal message](https://ably-real-time.slack.com/archives/C0178QGC093/p1655718602091409)

KacperKluka (approved):
LGTM :+1:

PR #433, opened 2022-07-05
Fix for running didFinishLaunchingWithOptions

Neelansh-ns:
Added fix for running `didFinishLaunchingWithOptions` in other registered plugins like uni_links.

Problem:

Since `didFinishLaunchingWithOptions` was returning NO, other plugins such as `uni_links`, also get the `initialAppLink` from `didFinishLaunchingWithOptions` were not able to run the function. This resulted in initialAppLink being `null` always.

### DIFF
--- a/ios/Classes/AblyFlutter.m
+++ b/ios/Classes/AblyFlutter.m
@@ -767,7 +767,7 @@ static const FlutterHandler _getFirstPage = ^void(AblyFlutter *const ably, Flutt
         PushHandlers.pushNotificationTapLaunchedAppFromTerminatedData = notification;
     }
     
-    return NO;
+    return YES;
 }
 
 #pragma mark - Push Notifications Registration - UIApplicationDelegate
@@ -785,7 +785,7 @@ static const FlutterHandler _getFirstPage = ^void(AblyFlutter *const ably, Flutt
 
 - (BOOL)application:(UIApplication *)application didReceiveRemoteNotification:(NSDictionary *)userInfo fetchCompletionHandler:(void (^)(UIBackgroundFetchResult))completionHandler {
     [_pushNotificationEventHandlers application:application didReceiveRemoteNotification:userInfo fetchCompletionHandler:completionHandler];
-    return NO;
+    return YES;
 }
 
 @end


QuintinWillison:
Thanks, @Neelansh-ns, for the contribution. ❤️ 
We'll take a proper look at this in due course, but my immediate observations...

From Apple's [documentation for `application:didFinishLaunchingWithOptions:`](https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1622921-application?language=objc) return value:

> NO if the app cannot handle the URL resource or continue a user activity, otherwise return YES. The return value is ignored if the app is launched as a result of a remote notification.

From Apple's [documentation for `application:didReceiveRemoteNotification:fetchCompletionHandler:`](https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1623013-application?language=objc), it looks like any return value is ignored as the return type is defined as `void`. Did you find that changing this return value had any impact on the problem you were solving with this fix, or is there a chance that just the `application:didFinishLaunchWithOptions:` change was the only remedy required?

We'll need to ascertain whether there was a good reason why an explicit choice to return `NO` was made. I expect that your fix is probably correct and that the current behaviour was likely lazy implementation on our behalf, for something that didn't have any visibility during testing due to the way that we have been testing. 🤔

Neelansh-ns:
> Did you find that changing this return value had any impact on the problem you were solving with this fix, or is there a chance that just the application:didFinishLaunchWithOptions: change was the only remedy required?

Yes @QuintinWillison , on changing the return type to YES, `didFinishLaunchingWithOptions` function runs for other plugins. You can try `uni_links` package. That's a great example to test this out. I have the same in my app.  I also confirmed by removing `ably_flutter` as a dependency.

arturjaworski (approved):

JakubJankowski (approved):

JakubJankowski:
Actually, let me take another look at why we're returning a bool in the `didReceiveRemoteNotification`, even though docs say it should be void.

arturjaworski (changes requested):
Documentation for `application:didReceiveRemoteNotification:fetchCompletionHandler:` says that it should return `void` type. It looks like some old issue from our side. Please make a change to introduce `void` type.

`application:didFinishLaunchingWithOptions:` change looks good :+1:

Neelansh-ns:
Hi @arturjaworski 

The apple documentation says that it should have a return void type, but the header file `FlutterPlugin.h`, has a return type `BOOL`.
Refer: https://github.com/flutter/engine/blob/main/shell/platform/darwin/ios/framework/Headers/FlutterPlugin.h#L97

So when we override the function with the `completionhandler`, we have to make the return type as `BOOL` itself instead of `void`.

The documentation asks us to return `YES` if the plugin handles the request.

Screenshot 2022-07-13 at 10 25 00 PM

```
- (BOOL)application:(UIApplication*)application
    didReceiveRemoteNotification:(NSDictionary*)userInfo
          fetchCompletionHandler:(void (^)(UIBackgroundFetchResult result))completionHandler
```

JakubJankowski:
Yeah, good point. LGTM 👍

arturjaworski (approved):
@Neelansh-ns You are right. Looks good to me 👍

QuintinWillison:
I'm landing this PR despite the docs workflow failing, due to the fact that we've identified that it's failing because it's unable to access the org-level secret `ABLY_AWS_ACCOUNT_ID_SDK` - which is as expected. GitHub state:

> Secrets are not passed to workflows that are triggered by a pull request from a fork.

PR #447, opened 2022-10-05
Expose error info in AblyException toString

### LINKED ISSUE #441: Expose `href` from `ErrorInfo` in bubbled exceptions
We have had a customer report getting this exception when making a call to get message history:

```
Non-fatal Exception: io.flutter.plugins.firebase.crashlytics.FlutterError: AblyException: java.lang.Exception: action not permitted, app = NMIcRg (40160 ). Error thrown null.
       at Platform.invokePlatformMethod(platform.dart:50)
       at PlatformObject.invokeRequest(platform_object.dart:66)
       at RealtimeChannel.history(realtime_channel.dart:53)
       at AblyOfficialPlugin._syncHistory(ably_official_plugin.dart:586)
```

@mattheworiordan asked in [this internal Slack thread](https://ably-real-time.slack.com/archives/CURL4U2FP/p1664854795741829):

> Do you know if it’s possible to include the error URL in these messages?  If so, they would have been taken to https://help.ably.io/error/40160, and had a better developer experience.

To which @QuintinWillison responded with:

> Related, I included `REST: Support Hyperlink on Request Failure` here:
> https://github.com/ably/features/pull/63
> The Android SDK does offer the `href` member on `ErrorInfo` (`TI4`) but we will need to dig deeper into both the Java and Flutter codebases to work out where/whether it gets bubbled up meaningfully enough.
> It also relates closely to `TI5` also, though that talks about "log entries" which is not that helpful. In this scenario, the actual log entry is presumably written or the exception is otherwise caught by the application code. So we may be more describing requirements of toString-like implementations down the stack. :thinking_face:

Creating this issue here, because it's in using this SDK for an Android app that we've had a customer experience pain here, however I suspect that most of the work to fix this may well be in our underlying [Java/Android SDK](https://github.com/ably/ably-java/).

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-2682) by [Unito](https://www.unito.io)

davyskiba:
Exposing full 'ErrorInfo' contents in 'AblyException.toString()' primarily to expose 'href'

Resolves #441

### DIFF
--- a/lib/src/error/src/ably_exception.dart
+++ b/lib/src/error/src/ably_exception.dart
@@ -31,7 +31,7 @@ class AblyException implements Exception {
 
   @override
   String toString() {
-    if(errorInfo != null){
+    if (errorInfo != null) {
       return 'AblyException ($errorInfo)';
     }
 
--- a/lib/src/error/src/ably_exception.dart
+++ b/lib/src/error/src/ably_exception.dart
@@ -32,11 +32,11 @@ class AblyException implements Exception {
   @override
   String toString() {
     if (errorInfo != null) {
-      return 'AblyException ($errorInfo)';
+      return 'AblyException: ($errorInfo)';
     }
 
     if (message == null) {
-      return 'AblyException (${(code == null) ? "" : '$code '})';
+      return 'AblyException: (${(code == null) ? "" : '$code '})';
     }
 
     return 'AblyException: $message (${(code == null) ? "" : '$code '})';
--- a/lib/src/error/src/ably_exception.dart
+++ b/lib/src/error/src/ably_exception.dart
@@ -31,9 +31,14 @@ class AblyException implements Exception {
 
   @override
   String toString() {
+    if(errorInfo != null){
+      return 'AblyException ($errorInfo)';
+    }
+
     if (message == null) {
       return 'AblyException (${(code == null) ? "" : '$code '})';
     }
+
     return 'AblyException: $message (${(code == null) ? "" : '$code '})';
   }
 }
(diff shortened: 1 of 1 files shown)


lib/src/error/src/ably_exception.dart
```diff
@@ -31,9 +31,14 @@ class AblyException implements Exception {
 
   @override
   String toString() {
+    if (errorInfo != null) {
+      return 'AblyException ($errorInfo)';
```
JakubJankowski, line 35: Should we maybe add a colon after `AblyException`?

JakubJankowski (approved):
LGTM 👍

lib/src/error/src/ably_exception.dart
QuintinWillison, line 35: Have you confirmed that if `errorInfo` _is_ not `null` then `code` and `message` _do not_ need to be communicated to the user? Currently the structure/API of this `AblyException` class is a bit confusing - especially given all three fields are extracted from the constructor that takes a `PlatformException`, implying that they all carry discrete informational value. 🤔

ikbalkaya (approved):
I don't have any additional comment on this as code looks good to me, but I think Quintin's question is an important one to answer.

lib/src/error/src/ably_exception.dart
davyskiba: Good catch, I've thought that PlatformException takes care of that, but it doesn't even hold the type of errorInfo.
52b1c98c20251eeb10d53c51868aae1f257f96c0

arturjaworski (approved):

QuintinWillison (approved):

Issue #459, opened 2022-12-14
How to handle when user auth changes?


yahya-uddin:
Currently I am using authCallback similar as follows:

```
realtime = ably.Realtime(
  options: ably.ClientOptions(
    key: '<API KEY>',
    authCallback: (ably.TokenParams tokenParams) async {
      // authService.token is the currently authenticated user token
      ably.TokenRequest tokenRequest = await createTokenRequest(tokenParams, authService.token);
      return tokenRequest;
    },
  ),
);
```

The issue is how do I handle when the user logs out, and logs back in with another user.

For example are there any methods like:

```
realtime.logout(); 
realtime.login();
```

The current workaround I am using is when the user logs out or changes I close the existing Realtime object, and recreate the Realtime object, so it is forced to re-authorise.

```
realtime.close();
realtime = ably.Realtime(
  options: ably.ClientOptions(
    key: '<API KEY>',
    authCallback: (ably.TokenParams tokenParams) async {
      ably.TokenRequest tokenRequest = await createTokenRequest(tokenParams, authService.token);
      return tokenRequest;
    },
  ),
);
```

yahya-uddin:
It seems like your supposed to use the authorise method:
https://ably.com/docs/api/realtime-sdk/authentication?lang=javascript#authorize

But there seems to be no Flutter equivalant.

Seems like a very common use-case, as a user can log-out and a new user can login any time.

yahya-uddin:
I notice the Auth object is commented out and links to an abstract class:
https://github.com/ably/ably-flutter/blob/1faba4ff8f8a8e434819d48753761b111c8a9b90/lib/src/platform/src/realtime/realtime.dart#L62

yahya-uddin:
Any update on this issue?

ikbalkaya:
Hey @yahya-uddin Sorry about late response on this. I just wanted to let you know that we are aware of this issue and we are going to aim to fix it as soon as possible.

Issue #444, opened 2022-10-04
authHeaders type conversion on Android Plugin is incorrect


ikbalkaya:
Please see this [SO question ](https://stackoverflow.com/questions/73888275/flutter-ably-java-util-hashmap-cannot-be-cast-to-io-ably-lib-types-param)for context.

I did a quick investigation and could reproduce the problem by creating a map of `authHeader`s

On Android Plugin `AblyMessageCodec` class has `decodeClientOptions` which has force type casts in it where it tries to convert from a `HashMap` to a `Param[]`

There are also other fields that are force casted to this and needs checking

┆Issue is synchronized with this [Jira Bug](https://ably.atlassian.net/browse/SDK-2686) by [Unito](https://www.unito.io)

Issue #439, opened 2022-09-28
java.util.HashMap cannot be cast to io.ably.lib.types.Param


jerinho:
Received this error
`java.lang.ClassCastException: java.util.HashMap cannot be cast to io.ably.lib.types.Param`

from this code

```
ably.ClientOptions options = ably.ClientOptions(
  clientId : ...,
  authUrl : ...,
  authHeaders : {
    'Authorization' : ...
  },
);
ably.Realtime realtime = await ably.Realtime(options : options);
```

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-2670) by [Unito](https://www.unito.io)

PR #460, opened 2023-02-22
Enable authorise

ikbalkaya:
This PR implements `Auth` interface for `ARTRest` and `ARTRealtime`

Includes implementation and platform codecs for iOS and Android. It also includes tests for each function. 

Note: Some integration tests failing (not related to changes I did)

Fixes https://github.com/ably/ably-flutter/issues/459
Fixes https://github.com/ably/ably-flutter/issues/444
Fixes https://github.com/ably/ably-flutter/issues/439


ikbalkaya:
https://github.com/ably/ably-java/issues/926 currently blocking this for Android

lib/src/platform/src/codec.dart
```diff
@@ -384,6 +384,24 @@ class Codec extends StandardMessageCodec {
...
   }
 
+  Map<String, dynamic> _encodeAuthOptions(
+      final AuthOptions authOptions) {
+    final jsonMap = <String, dynamic>{};
+    jsonMap[TxAuthOptions.tokenDetails] = _encodeTokenDetails(authOptions
+        .tokenDetails);
+    jsonMap[TxAuthOptions.authUrl] = authOptions.authUrl;
+    jsonMap[TxAuthOptions.authMethod] = authOptions.authMethod;
+    //jsonMap[TxAuthOptions.authCallback] = authOptions.authCallback;
```
owenpearson, line 394: What's the reason for leaving `authCallback` commented out here? If this is intentional I think you should add a comment explaining why it's not supported

lib/src/authentication/src/auth.dart
```diff
@@ -63,7 +63,7 @@ abstract class Auth {
   /// with, the default values. To understand why an Ably [TokenRequest] may be
   /// issued to clients in favor of a token, see
   /// [Token Authentication explained](https://ably.com/docs/core-features/authentication/#token-authentication).
-  Future<TokenDetails> requestToken({
+  Future<TokenDetails?> requestToken({
```
owenpearson, line 66: What's the reason for making these optional? According to the spec these methods should never return nil

ios/Classes/codec/AblyFlutterWriter.m
```diff
@@ -297,6 +304,37 @@ +(NSString *) encodePresenceAction: (ARTPresenceAction) action {
...
+};
+
+static AblyCodecEncoder encodeTokenDetails = ^NSMutableDictionary*(ARTTokenDetails *const details) {
+    NSMutableDictionary<NSString *, NSObject *> *dictionary = [[NSMutableDictionary alloc] init];
+    
+    WRITE_VALUE(dictionary, TxTokenDetails_token, [details token]);
+
+    WRITE_VALUE(dictionary, TxTokenDetails_issued, [details issued]?@((long)([[details issued] timeIntervalSince1970]*1000)):nil);
+    WRITE_VALUE(dictionary, TxTokenDetails_expires, [details expires]?@((long)([[details expires] timeIntervalSince1970]*1000)):nil);
+     //WRITE_VALUE(dictionary, TxTokenDetails_expires, [details expires]);
```
owenpearson, line 329: Some commented out code left here, did you mean to remove this?

ios/Classes/codec/AblyFlutterReader.m
```diff
@@ -155,6 +157,39 @@ use an explicitly received null from Dart (manifesting this side as a nil id)
...
+
+    return authOptions;
+};
+
+static AblyCodecDecoder readTokenParams = ^ARTTokenParams*(NSDictionary *const dictionary) {
+    ARTTokenParams *const tokenParams = [ARTTokenParams new];
+
+    READ_VALUE(tokenParams, capability, dictionary, TxTokenParams_capability);
+    READ_VALUE(tokenParams, clientId, dictionary, TxTokenParams_clientId)
+   // READ_VALUE(tokenParams, nonce, dictionary, TxTokenParams_nonce);
```
owenpearson, line 185: Likewise here

test_integration/lib/test/rest/rest_request_token_test.dart
```diff
@@ -0,0 +1,30 @@
...
+    logLevel: LogLevel.verbose,
+  );
+
+  final rest = Rest(
+    options: clientOptionsForToken,
+  );
+
+  final token = await rest.auth?.requestToken();
+
+  if (token == null) {
```
owenpearson, line 25: I know we probably don't need to test these auth methods extensively since they're already tested by the underlying libraries but I think we ought to do a bit more than just checking that they return something. For example I think it would be better to test that you can successfully initialise a new realtime client using a token obtained from `requestToken`

owenpearson (changes requested):
Looks good so far! I've left a few comments, also it looks like there's some CI failures including the formatting check

lib/src/platform/src/codec.dart
ikbalkaya: Good point. I suppose I was trying to stay in line with the return type of `invoke` which was optional. But looks like there is `invokeRequest` that does have a non-null return value. Changes here 8300e231bb8244d85c72e5d2be6ca47b34df61b5

lib/src/authentication/src/auth.dart
ikbalkaya: Good point. I suppose I was trying to stay in line with the return type of `invoke` which was optional. But looks like there is `invokeRequest` that does have a non-null return value. Changes here 8300e231bb8244d85c72e5d2be6ca47b34df61b5

ios/Classes/codec/AblyFlutterReader.m
ikbalkaya: a3748101667c1d06f1e12538477620c31d90b49c

android/src/main/java/io/ably/flutter/plugin/AuthMethodHandler.java
```diff
@@ -0,0 +1,102 @@
...
+
+        executor.execute(() -> {
+            try {
+                final Auth.TokenDetails tokenDetails = getAuth(ablyMessage, type)
+                        .requestToken(tokenParams, options);
+                result.success(tokenDetails);
+               // handler.post(() -> result.success(tokenDetails));
+            } catch (AblyException e) {
+                result.error(String.valueOf(e.errorInfo.code), e.errorInfo.message, e);
+               // handler.post(() ->result.error(String.valueOf(e.errorInfo.code), e.errorInfo.message, e));
```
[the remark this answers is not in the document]
ikbalkaya: a3748101667c1d06f1e12538477620c31d90b49c

lib/src/platform/src/codec.dart
ikbalkaya: That was another one I had disabled for debugging as I wasn't sure how callbacks were handled on this particular SDK. Turns out that [method_call_handler.dart](https://github.com/ably/ably-flutter/blob/d5476c2b8c6a29b1da22f9871f46d41cd5597b29/lib/src/platform/src/method_call_handler.dart#L4) handles downstream method calls such as this callback. I have made a comment and removed commented out code 95d040f8a0c15cd4bd5f5a0db1855bde1421f755

test_integration/lib/test/rest/rest_request_token_test.dart
ikbalkaya: Removed some unnecessary tests and improved some 4768cf3966977460616432dc0c8da88adf997b5d and 6ac176204355356fe63d82b857e253255117919b

android/src/main/java/io/ably/flutter/plugin/AuthMethodHandler.java
```diff
@@ -0,0 +1,99 @@
...
+                (Auth.AuthOptions) ablyMessage.message.get(PlatformConstants.TxTransportKeys.options);
+
+        executor.execute(() -> {
+            try {
+                final Auth.TokenDetails tokenDetails = getAuth(ablyMessage, type)
+                        .authorize(tokenParams, options);
+                result.success(tokenDetails);
+            } catch (AblyException e) {
+                result.error(String.valueOf(e.errorInfo.code), e.errorInfo.message, e);
+               // handler.post(() ->result.error(String.valueOf(e.errorInfo.code), e.errorInfo.message, e));
```
owenpearson, line 44: There's another commented out line here

owenpearson (changes requested):

test_integration/lib/test/realtime/realtime_authorize_test.dart
```diff
@@ -0,0 +1,50 @@
...
+    key: appKey,
+    environment: 'sandbox',
+    logLevel: LogLevel.verbose,
+  );
+
+  final ablyForToken = Rest(
+    options: clientOptionsForToken,
+  );
+  /* get first token */
+  final firstToken = await ablyForToken.auth?.requestToken();
```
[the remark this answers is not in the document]
ikbalkaya: That's a good point. Since auth is optional it won't. But I guess it should not be optional 🤔

android/src/main/java/io/ably/flutter/plugin/AuthMethodHandler.java
ikbalkaya: 567f17b6d427ed3173f673ebfe355d4490dcc53c

lib/src/platform/src/realtime/realtime.dart
```diff
@@ -19,6 +20,8 @@ class Realtime extends PlatformObject {
     _connection = Connection(this);
     _channels = RealtimeChannels(this);
     push = Push(realtime: this);
+    // auth = RealtimeAuth(this);
```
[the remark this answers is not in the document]
ikbalkaya: 567f17b6d427ed3173f673ebfe355d4490dcc53c

.github/workflows/check.yaml
```diff
@@ -13,7 +13,7 @@ jobs:
 
       - uses: subosito/flutter-action@v2
         with:
-          flutter-version: '3.0.x'
+          flutter-version: '3.3.3'
```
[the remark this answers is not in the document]
ikbalkaya: As discussed, this should no longer be the case :)

owenpearson (changes requested):
The code looks good now! There's just one failing test which needs to be fixed: [rest instance with token has a valid handle and successfully publishes messages](https://github.com/ably/ably-flutter/actions/runs/4818935007/jobs/8581527675?pr=460#step:5:8036)

owenpearson (approved):
LGTM, nice work 🚀

Issue #465, opened 2023-04-29
Realtime connection with authUrl causes crash on iOS


motiapp:
This basic connection code causes a fatal crash of my iOS app every time that **connect()** is called on an **ably.Realtime** instance. I have tried turning off autoConnect and manually connecting, and it crashes either way. It happens every time on both the Simulator and a iPhone X device. When running on the simulator I get a crash report in a system dialog window (not from flutter's console).

My connection code is basic and works just fine on the Android simulator.
<pre><code>
final options = ably.ClientOptions(
  authUrl: ApiUrls.ablyAuthUrl, // api auth endpoint url
  authMethod: "get",
  authHeaders: {
    "Authorization": "Bearer $accessToken", // session token from FirebaseAuth
    "Content-Type": "application/json",
  },
  echoMessages: false,
);
final realtime = ably.Realtime(options: options);
log("Ably connection state: ${_realtime?.connection.state}");
realtime?.connection.on().listen((stateChange) {
  log("Ably connection state changed: ${stateChange.event}");
  final isConnected = stateChange.event == ably.ConnectionEvent.connected;
});
</code></pre>

Does this even work on iOS for anyone right now?

iOS app crash log:
<pre><code>
-------------------------------------
Translated Report (Full Report Below)
-------------------------------------

Incident Identifier: 4CE4E757-6E44-4DC4-B86B-46B3DC44AEFC
CrashReporter Key:   51B41356-FD63-BD11-3BAF-7A9FA60DACDB
Hardware Model:      MacBookPro18,2
Process:             Runner [60368]
Path:                /Users/USER/Library/Developer/CoreSimulator/Devices/1200B6F6-D6A1-4074-92BE-8D8869AD7211/data/Containers/Bundle/Application/BC25DC43-6B0D-4A92-8CAF-2CB9446F4771/Runner.app/Runner
Identifier:          com.***********
Version:             4.2.3 (65)
Code Type:           ARM-64 (Native)
Role:                Foreground
Parent Process:      launchd_sim [98415]
Coalition:           com.apple.CoreSimulator.SimDevice.1200B6F6-D6A1-4074-92BE-8D8869AD7211 [236917]
Responsible Process: SimulatorTrampoline [14671]

Date/Time:           2023-04-29 04:04:37.7352 -0400
Launch Time:         2023-04-29 04:04:22.5390 -0400
OS Version:          macOS 13.2.1 (22D68)
Release Type:        User
Report Version:      104

[...]

PR #466, opened 2023-05-02
Read authUrl as NSURL

ikbalkaya:
`authUrl` was being read as string which wasn't treated as an error as type mismatch is allowed in Objective C. 

This PR fixes this using authURL as NSURL

Also it turned out that ably-java didn't like `authUrl` having token as query string parameter. That's why I made a change on ably-java and upgraded its version here 

Fixes https://github.com/ably/ably-flutter/issues/465

### DIFF
--- a/test_integration/lib/test/realtime/realtime_auth_url_test.dart
+++ b/test_integration/lib/test/realtime/realtime_auth_url_test.dart
@@ -1,3 +1,5 @@
+import 'dart:async';
+
 import 'package:ably_flutter/ably_flutter.dart';
 import 'package:ably_flutter_integration_test/app_provisioning.dart';
 import 'package:ably_flutter_integration_test/factory/reporter.dart';
@@ -30,16 +32,16 @@ Future<Map<String, dynamic>> testCreateRealtimeWithAuthUrl({
       autoConnect: false,
       logLevel: LogLevel.verbose);
   final realtime = Realtime(options: options);
-  var connected = false;
+  final completer = Completer<void>();
   realtime.connection.on().listen((stateChange) {
-    if (stateChange.event == ConnectionEvent.connected) {
-      connected = true;
+    if (stateChange.current == ConnectionState.connected) {
+      completer.complete();
     }
   });
   await realtime.connect();
-  await Future<void>.delayed(const Duration(seconds: 5));
-  if (!connected) {
+  await completer.future.timeout(const Duration(seconds: 30), onTimeout: () {
     throw Error();
-  }
+  });
+
   return {'handle': await realtime.handle};
 }
--- a/test_integration/lib/test/realtime/realtime_auth_url_test.dart
+++ b/test_integration/lib/test/realtime/realtime_auth_url_test.dart
@@ -1,4 +1,3 @@
-
 import 'dart:io';
 
 import 'package:ably_flutter/ably_flutter.dart';
--- a/test_integration/lib/test/realtime/realtime_auth_url_test.dart
+++ b/test_integration/lib/test/realtime/realtime_auth_url_test.dart
@@ -1,4 +1,5 @@
 import 'dart:async';
+import 'dart:ffi';
 
 import 'package:ably_flutter/ably_flutter.dart';
 import 'package:ably_flutter_integration_test/app_provisioning.dart';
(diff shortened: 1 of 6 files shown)


test_integration/lib/test/realtime/realtime_auth_url_test.dart
```diff
@@ -0,0 +1,42 @@
...
+import 'package:ably_flutter/ably_flutter.dart';
+import 'package:ably_flutter_integration_test/app_provisioning.dart';
+import 'package:ably_flutter_integration_test/factory/reporter.dart';
+
+//Make sure that creating realtime with authURL won't crash
+Future<Map<String, dynamic>> testCreateRealtimeWithAuthUrl({
+  required Reporter reporter,
+  Map<String, dynamic>? payload,
+}) async {
+  reporter.reportLog('init start');
```
owenpearson, line 12: can this log be removed?

test_integration/lib/test/realtime/realtime_auth_url_test.dart
```diff
@@ -0,0 +1,42 @@
...
+    HttpHeaders.contentTypeHeader: "application/json",
+    HttpHeaders.acceptLanguageHeader: "ios",
+  };
+  final options = ClientOptions(
+    authUrl: 'https://auth.ably.io/auth',
+    authHeaders: headers,
+    echoMessages: false,
+  );
+  final realtime = Realtime(options: options);
+  realtime.connection.on().listen((stateChange) {
```
owenpearson, line 37: Does this test actually wait for the client to be connected? It looks like you're just adding a listener to log `stateChange`s to the console but not actually checking whether the client becomes connected

test_integration/lib/test/realtime/realtime_auth_url_test.dart
```diff
@@ -0,0 +1,42 @@
...
+  );
+  final accessToken = await ablyForToken.auth.requestToken();
+
+  final Map<String, String> headers = {
+    HttpHeaders.authorizationHeader: "Bearer $accessToken",
+    HttpHeaders.contentTypeHeader: "application/json",
+    HttpHeaders.acceptLanguageHeader: "ios",
+  };
+  final options = ClientOptions(
+    authUrl: 'https://auth.ably.io/auth',
```
owenpearson, line 32: I don't think `auth.ably.io` is a valid domain. did you mean to mock the http response? it may be easier to just use. the echo server

owenpearson (changes requested):
Looks good so far! Just a few changes needed to the test fixture

test_integration/lib/test/realtime/realtime_auth_url_test.dart
ikbalkaya: I think it should stay as this is the standard for every test.

test_integration/lib/test/realtime/realtime_auth_url_test.dart
ikbalkaya: Not really. This test is only verifying that app won't crash in case of an authUrl is passed as described https://github.com/ably/ably-flutter/pull/466/commits/746fef4aab5901ff126f048ec7dc56e1358cb048

test_integration/lib/test/realtime/realtime_auth_url_test.dart
ikbalkaya: As above

ikbalkaya: Test for connections were added

test_integration/lib/test/realtime/realtime_auth_url_test.dart
ikbalkaya: Tests for connections were added

test_integration/lib/test/realtime/realtime_auth_url_test.dart
```diff
@@ -0,0 +1,45 @@
...
+  realtime.connection.on().listen((stateChange) {
+    if (stateChange.event == ConnectionEvent.connected) {
+      connected = true;
+    }
+  });
+  await realtime.connect();
+  await Future<void>.delayed(const Duration(seconds: 5));
+  if (!connected) {
+    throw Error();
+  }
```
owenpearson, line 43: It's a bit hacky just waiting for 5 seconds and then checking if we're connected - would it be possible to use a completion listener to wait for the connected event instead:
```dart
final completer = Completer<void>();
realtime.connection.on().listen((stateChange) {
  if (stateChange.current == ConnectionState.connected) {
    completer.complete();
  }
});
await realtime.connect();
await completer.future;
```

owenpearson (changes requested):

test_integration/lib/test/realtime/realtime_auth_url_test.dart
ikbalkaya: That's a good idea. I made change 3de55d70531ed8fb7c8e92f78831181f574a9a6b . This is with 30 seconds timeout, otherwise test wouldn't finish

owenpearson: nice, thanks!

owenpearson (approved):
LGTM, thanks

PR #470, opened 2023-06-14
Fix java/android custom fallbacks

### LINKED ISSUE #469: ClientOptions.fallbackHosts broken on android
When the `fallbackHosts` client option is used in an android application the following exception is thrown by the java codec:

```
java.lang.ClassCastException: java.util.ArrayList cannot be cast to java.lang.String[]
```

The underlying issue is that ably-java expects a fixed sized `String[]` but ably-flutter attempts to pass the fallback hosts as a dynamic `ArrayList<String>`.

sacOO7:
- Fixes #469


owenpearson (approved):
LGTM, nice one

sacOO7:
Thanks!

PR #473, opened 2023-07-06
[SDK-3701] Fix/missing client options android decoder

### LINKED ISSUE #472: Update android decoder for updated clientOptions
Once this issue -> https://github.com/ably/ably-java/issues/956 is resolved, need to update android decoder for the same.

sacOO7:
Fixed #472

### DIFF
--- a/example/android/build.gradle
+++ b/example/android/build.gradle
@@ -1,5 +1,5 @@
 buildscript {
-    ext.kotlin_version = '1.9.0'
+    ext.kotlin_version = '1.5.31'
 
     repositories {
         google()
--- a/example/android/build.gradle
+++ b/example/android/build.gradle
@@ -1,5 +1,5 @@
 buildscript {
-    ext.kotlin_version = '1.5.31'
+    ext.kotlin_version = '1.9.0'
 
     repositories {
         google()
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -23,7 +23,7 @@ apply plugin: 'com.android.library'
 
 dependencies {
     // https://github.com/ably/ably-java/
-    implementation 'io.ably:ably-android:1.2.28'
+    implementation 'io.ably:ably-android:1.2.31'
 
     // https://firebase.google.com/docs/cloud-messaging/android/client
     implementation 'com.google.firebase:firebase-messaging:23.0.4'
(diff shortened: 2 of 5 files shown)


AndyTWF:
Looks like we've got some failing CI here?

sacOO7:
> Looks like we've got some failing CI here?

Yes, I will check what might be going wrong 👍

example/android/build.gradle
```diff
@@ -1,5 +1,5 @@
 buildscript {
-    ext.kotlin_version = '1.5.31'
+    ext.kotlin_version = '1.9.0'
```
owenpearson, line 2: I'm gonna need some more context around this commit - what exactly has changed so that this now requires kotlin 1.9.0? For context kotlin 1.9.0 was released last week. ideally the version here should be set as low as possible to ensure that the library continues to work with older versions of kotlin

sacOO7: This is just an example app. It was failing, because it was explicitly asking to update this version.

https://github.com/ably/ably-flutter/actions/runs/5478663966/jobs/10045729475

owenpearson: yeah but what has changed which means we need to upgrade the kotlin version now? presumably it was working before?

sacOO7: Because out hunch is, if example app can't accept older versions of kotlin, then our sdk users might face the same issue

AndyTWF: This is a problem on `main` too - the source is the `fluttertoast` package we use for then example app.

In this PR https://github.com/ponnamkarthik/FlutterToast/pull/429 the minimum Kotlin version was changed. There aren't any tags on GitHub, but if we lock the package to `<=8.2.0` that should fix it :)

sacOO7: Okay, so it has nothing to do with the change we made `io.ably:ably-android:1.2.31` right?

sacOO7: So, I should change the kotlin version to `8.2.0` or it's fine to keep it as `1.9.0` ?

sacOO7: Okay, I have updated the lib version from [pubspec.yaml](https://github.com/ably/ably-flutter/pull/473/files#diff-565b0869896732da3b937c64bc8fd5f0ca37ea1f629d579c29fc70e0f1e3e48e)

AndyTWF (approved):
Changes LGTM - thanks 👍🏻 

Looks like we're failing on Android 29, perhaps related to https://github.com/ReactiveCircus/android-emulator-runner/issues/324, although the error is slightly different.

Can you have a look into this? :)

sacOO7:
> Changes LGTM - thanks 👍🏻
> 
> Looks like we're failing on Android 29, perhaps related to [ReactiveCircus/android-emulator-runner#324](https://github.com/ReactiveCircus/android-emulator-runner/issues/324), although the error is slightly different.
> 
> Can you have a look into this? :)

Actually I am retrying the build. Most likely related to spawning the emulator. Since, it's a emulator related issue, not sure what can be done to fix it, other than retrying : (

example/android/build.gradle
owenpearson: thanks both 👍

owenpearson (approved):
LGTM, thanks

PR #479, opened 2023-09-15
fix: invoke `result.success` after `pushReset` 

ttypic:
To avoid timeouts invoke `result.success` after `pushReset`

### DIFF
--- a/android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java
+++ b/android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java
@@ -648,6 +648,7 @@ public class AblyMethodCallHandler implements MethodChannel.MethodCallHandler {
       instanceStore.getPush(ablyMessage.handle)
               .getActivationContext()
               .reset();
+      result.success(null);
     } catch (AblyException e) {
       handleAblyException(result, e);
     }


sacOO7:
Nice fix 👍

android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java
```diff
@@ -648,6 +648,7 @@ private void pushReset(@NonNull MethodCall call, @NonNull MethodChannel.Result r
       instanceStore.getPush(ablyMessage.handle)
               .getActivationContext()
               .reset();
```
ttypic, line 650: The main problem with `reset` is that it clears ably client from `activationContext` and it breaks everything until app is reloaded. There are 2 possible solutions:
- Reinstall ably client in flutter (I've checked locally and it works)
```java
instanceStore.getPush(ablyMessage.handle)
              .getActivationContext().setAbly((AblyRest) instanceStore.getAblyClient(ablyMessage.handle));
```
- Do not clear ably client in reset (it looks for me that it is unnecessary, because ably client doesn't store any state) [PR here](https://github.com/ably/ably-java/pull/968)

sacOO7: Based on my observations while reading the code, I understood that `reset` breaks activation/deactivation state machine. It was never been a proper solution to the problem. As per PR (https://github.com/ably/ably-flutter/pull/357), it was introduced as an experimental feature that was never tested as such and I couldn't see any test cases available for the same

ttypic: @sacOO7 actually reset works in one particular very specific scenario: when ably environment has changed. If this happened `activate`/`deactivate` stop working. I agree that it's not very good solution, but it looks like it works. Basically it doesn't break state machine it simple delete it and recreate.

PR #489, opened 2023-10-25
feat: protect from multiple `MethodChannel.Result` invocation

ttypic:
Related to [SDK-3906](https://ably.atlassian.net/browse/SDK-3906)

[SDK-3906]: https://ably.atlassian.net/browse/SDK-3906?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sacOO7 (approved):
LGTM!
Just in case, you might like to add a test to avoid a case where method call consistently throws exception.
Based on what I know, we shouldn't be posting callbacks on main threads, it can block app UI execution and main thread exceptions should be handled by developers themselves. Since this is an old working code, there might be some reason : )

PR #499, opened 2023-12-06
feat: setOnBackgroundMessage() is now a Future<void> that will await for the function to complete.

ttypic:
Copied https://github.com/ably/ably-flutter/pull/498


sacOO7 (approved):
LGTM

PR #503, opened 2024-02-21
Fix realtime connection fields

### LINKED ISSUE #502: How do we get connectionId from connection?
It looks like the Connection object has an id property, but that and the other properties are not populated (except the state). What is the intended way to populate these fields? Can we somehow help adding these fields? The JS lib seems to properly give us the info we need.

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-4629) by [Unito](https://www.unito.io)

sacOO7:
- Fields available to use => Connection `id`, `key` and `recoveryKey`.
- Fixed #502

### DIFF
--- a/ios/Classes/codec/AblyPlatformConstants.h
+++ b/ios/Classes/codec/AblyPlatformConstants.h
@@ -79,6 +79,7 @@ extern NSString *const AblyPlatformMethod_realtimeAuthGetClientId;
 extern NSString *const AblyPlatformMethod_connectionId;
 extern NSString *const AblyPlatformMethod_connectionKey;
 extern NSString *const AblyPlatformMethod_connectionRecoveryKey;
+extern NSString *const AblyPlatformMethod_connectionIdUpdated;
 
 extern NSString *const AblyPlatformMethod_pushActivate;
 extern NSString *const AblyPlatformMethod_pushDeactivate;
--- a/ios/Classes/codec/AblyPlatformConstants.h
+++ b/ios/Classes/codec/AblyPlatformConstants.h
@@ -75,6 +75,12 @@ extern NSString *const AblyPlatformMethod_realtimeAuthAuthorize;
 extern NSString *const AblyPlatformMethod_realtimeAuthCreateTokenRequest;
 extern NSString *const AblyPlatformMethod_realtimeAuthRequestToken;
 extern NSString *const AblyPlatformMethod_realtimeAuthGetClientId;
+
+extern NSString *const AblyPlatformMethod_connectionId;
+extern NSString *const AblyPlatformMethod_connectionKey;
+extern NSString *const AblyPlatformMethod_connectionRecoveryKey;
+extern NSString *const AblyPlatformMethod_connectionErrorReason;
+
 extern NSString *const AblyPlatformMethod_pushActivate;
 extern NSString *const AblyPlatformMethod_pushDeactivate;
 extern NSString *const AblyPlatformMethod_pushReset;
--- a/ios/Classes/codec/AblyPlatformConstants.h
+++ b/ios/Classes/codec/AblyPlatformConstants.h
@@ -79,7 +79,6 @@ extern NSString *const AblyPlatformMethod_realtimeAuthGetClientId;
 extern NSString *const AblyPlatformMethod_connectionId;
 extern NSString *const AblyPlatformMethod_connectionKey;
 extern NSString *const AblyPlatformMethod_connectionRecoveryKey;
-extern NSString *const AblyPlatformMethod_connectionErrorReason;
 
 extern NSString *const AblyPlatformMethod_pushActivate;
 extern NSString *const AblyPlatformMethod_pushDeactivate;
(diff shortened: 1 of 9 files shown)


lib/src/platform/src/realtime/connection.dart
```diff
@@ -37,7 +37,8 @@ class Connection extends PlatformObject {
 
   /// A unique public identifier for this connection, used to identify this
   /// member.
-  String? id;
+  Future<String?> get id async =>
```
ttypic, line 40: It works, but this way we introduce breaking changes, and it's not align with the spec. Is there any chance we can subscribe to connection object changes (something similar to what we do for state)? I believe at least `id` and `errorReason` can't be changed without changing connection state, so we can simply add those fields in our listener

ttypic (changes requested):
Thanks, Sachin. It works well. However, I don't really like that we've changed existing fields to async props, even if these fields never worked properly.

I think we should either create new async methods with different names as a workaround in the Connection instead or come up with a solution for making them work as fields (similar to `state`).

maratal:
I'm not sure what is happening in this PR since as I said I'm not familiar with the flutter codebase, but can't `id` be set on `.connected` event? I believe I've seen this handler.

lib/src/platform/src/realtime/connection.dart
sacOO7: We can do this, but there would be inconsistency for other similar fields. Not every field will change for change in the state. Change can be made subject to discussion -> https://ably-real-time.slack.com/archives/C030C5YLY/p1708586146096219?thread_ts=1708354649.094169&cid=C030C5YLY

sacOO7: Okay, updated -> https://github.com/ably/ably-flutter/pull/503/commits/7609df0cda0e530b320e97a10d2f2bc2f7d2776e. This is still subject to how we decide to set class fields in the future.

sacOO7: Please note that this can cause inconsistency in retrieved `id` value if accessed before setting it inside a callback. Imagine a scenario where we have 
realtime.connection.on('connected', ()=> {
id = realtime.connection.id; // this can return old `id` value since new one is yet to be updated via callback : (
}

lib/src/platform/src/realtime/connection.dart
```diff
@@ -24,6 +24,7 @@ class Connection extends PlatformObject {
     on().listen((event) {
       _state = event.current;
       _errorReason = event.reason;
+      invoke<String?>(PlatformMethod.connectionId).then((_id) => id = _id);
```
ttypic, line 27: Could you move this back? This way, id will be updated with a delay, which isn't good. I suggested that we send a syntactic event on the platform side with the id (not only the state), subscribe to it here on the syntactic event, and retrieve the id from the event.

sacOO7: This is not a better solution either. We still can't guarantee delay in the event. Also, `ConnectionEvent` doesn't include `id` according to the spec.

sacOO7: I feel it's better if we keep `id` as async

sacOO7: https://sdk.ably.com/builds/ably/specification/main/features/#RTN4:~:text=enum%20ConnectionEvent

ttypic: Flutter SDK is a hack itself :) I believe it's very transparent solution, we can guarantee, that when connection state is become connected, `id` field will be set (what else do you think sdk user may expect? 

>ConnectionEvent doesn't include id according to the spec => https://sdk.ably.com/builds/ably/specification/main/features/#RTN4:~:text=enum%20ConnectionEvent

It doesn't matter it won't be spec connectionevent, it will be internal synthetic event. Public of course should be implemented according to the spec without id.

sacOO7: updated

ios/Classes/codec/AblyPlatformConstants.h
```diff
@@ -75,6 +75,12 @@ extern NSString *const AblyPlatformMethod_realtimeAuthAuthorize;
 extern NSString *const AblyPlatformMethod_realtimeAuthCreateTokenRequest;
 extern NSString *const AblyPlatformMethod_realtimeAuthRequestToken;
 extern NSString *const AblyPlatformMethod_realtimeAuthGetClientId;
+
+extern NSString *const AblyPlatformMethod_connectionId;
+extern NSString *const AblyPlatformMethod_connectionKey;
+extern NSString *const AblyPlatformMethod_connectionRecoveryKey;
+extern NSString *const AblyPlatformMethod_connectionErrorReason;
```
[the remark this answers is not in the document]
sacOO7: sure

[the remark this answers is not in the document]
sacOO7: done

sacOO7:
Link to internal convo - https://ably-real-time.slack.com/archives/C030C5YLY/p1708354649094169

sacOO7:
Marking as resolved, since it's already fixed as a part of https://github.com/ably/ably-flutter/pull/505

PR #505, opened 2024-02-26
[ECO-4629] fix: propagate id and key fields for connection object for Android and iOS

ttypic:
Resolves https://github.com/ably/ably-flutter/issues/502
Jira issue [ECO-4629](https://ably.atlassian.net/browse/ECO-4629) 

Add `key` and `id` to `ConnectionStateChange` event, updated example app

[ECO-4629]: https://ably.atlassian.net/browse/ECO-4629?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/android/src/main/java/io/ably/flutter/plugin/dto/EnrichedConnectionStateChange.java
+++ b/android/src/main/java/io/ably/flutter/plugin/dto/EnrichedConnectionStateChange.java
@@ -1,0 +1,15 @@
+package io.ably.flutter.plugin.dto;
+
+import io.ably.lib.realtime.ConnectionStateListener;
+
+public class EnrichedConnectionStateChange {
+    public final ConnectionStateListener.ConnectionStateChange stateChange;
+    public final String connectionId;
+    public final String connectionKey;
+
+    public EnrichedConnectionStateChange(ConnectionStateListener.ConnectionStateChange stateChange, String connectionId, String connectionKey) {
+        this.stateChange = stateChange;
+        this.connectionId = connectionId;
+        this.connectionKey = connectionKey;
+    }
+}
--- a/lib/src/platform/src/realtime/connection.dart
+++ b/lib/src/platform/src/realtime/connection.dart
@@ -13,6 +13,10 @@ class Connection extends PlatformObject {
 
   ErrorInfo? _errorReason;
 
+  String? _id;
+
+  String? _key;
+
   /// @nodoc
   /// Instantiates a connection with [realtime] client instance.
   ///
@@ -21,9 +25,11 @@ class Connection extends PlatformObject {
   Connection(this.realtime)
       : _state = ConnectionState.initialized,
         super() {
-    on().listen((event) {
-      _state = event.current;
-      _errorReason = event.reason;
+    _onConnectionStateChange().listen((event) {
+      _state = event.stateChange.current;
+      _errorReason = event.stateChange.reason;
+      _id = event.connectionId;
+      _key = event.connectionKey;
     });
   }
 
@@ -37,7 +43,7 @@ class Connection extends PlatformObject {
 
   /// A unique public identifier for this connection, used to identify this
   /// member.
-  String? id;
+  String? get id => _id;
 
   /// A unique private connection key used to recover or resume a connection,
   /// assigned by Ably.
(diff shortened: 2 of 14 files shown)


android/src/main/java/io/ably/flutter/plugin/dto/EnrichedConnectionStateChange.java
```diff
@@ -0,0 +1,15 @@
+package io.ably.flutter.plugin.dto;
+
+import io.ably.lib.realtime.ConnectionStateListener;
+
+public class EnrichedConnectionStateChange {
```
sacOO7, line 5: Ideally, this class should extend `ConnectionStateListener.ConnectionStateChange`

lib/src/realtime/src/connection_state_change.dart
```diff
@@ -35,3 +35,20 @@ class ConnectionStateChange {
     this.retryIn,
   });
 }
+
+/// @nodoc
+@immutable
+class EnrichedConnectionStateChange {
```
sacOO7, line 41: `EnrichedConnectionStateChange` can extend `ConnectionStateChange` so no need to filter incoming stream messages using `map`

android/src/main/java/io/ably/flutter/plugin/dto/EnrichedConnectionStateChange.java
sacOO7, line 5: `EnrichedConnectionStateChange ` can be redefined to extend `ConnectionStateListener.ConnectionStateChange`

lib/src/realtime/src/connection_state_change.dart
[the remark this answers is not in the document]
ttypic: I don't think it's a good idea, both `EnrichedConnectionStateChange` and `ConnectionStateChange` are data-classes, inheritance it that case is a bad practice.

android/src/main/java/io/ably/flutter/plugin/dto/EnrichedConnectionStateChange.java
ttypic: That's not a good idea and the problem of ably-java implementation. This class `ConnectionStateListener.ConnectionStateChange` definitely should be final, it's internal data class, there is no reason somebody should extend it.

lib/src/realtime/src/connection_state_change.dart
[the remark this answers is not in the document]
sacOO7: This is also a stateChange `event`. It's not uncommon pattern to extend base-events to create a custom event.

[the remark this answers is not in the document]
sacOO7: Anyways, I was also suggesting in terms of code readability.

lib/src/platform/src/realtime/connection.dart
```diff
@@ -21,9 +25,11 @@ class Connection extends PlatformObject {
   Connection(this.realtime)
       : _state = ConnectionState.initialized,
         super() {
-    on().listen((event) {
-      _state = event.current;
-      _errorReason = event.reason;
+    _onConnectionStateChange().listen((event) {
+      _state = event.stateChange.current;
+      _errorReason = event.stateChange.reason;
+      _id = event.connectionId;
```
maratal, line 31: I think it should be `_id = this.realtime.connectionId` where `connectionId` is something like [this](https://github.com/ably/ably-flutter/blob/2e1839d609ca076ff6e7c3e4f039c818c3bec25d/android/src/main/java/io/ably/flutter/plugin/AblyMethodCallHandler.java#L579), but not async, wdyt @ttypic @sacOO7 ?

maratal: `private void connectionId(@NonNull MethodCall methodCall, @NonNull MethodChannel.Result result)` can't be called from dart code? via `invoke` or something

sacOO7: We can connect over DM

android/src/main/java/io/ably/flutter/plugin/dto/EnrichedConnectionStateChange.java
maratal: Why not to add those properties in `ConnectionStateListener.ConnectionStateChange` directly?

sacOO7 (approved):
LGTM

PR #520, opened 2024-04-16
[ECO-4748] Adds privacy manifest for iOS plugin

umair-ably:
- Includes the same changes from our ably-cocoa repo [here](https://github.com/ably/ably-cocoa/pull/1900)
- Bumps cocoa dependency to v1.2.30

### DIFF
--- a/example/ios/Podfile.lock
+++ b/example/ios/Podfile.lock
@@ -1,9 +1,9 @@
 PODS:
-  - Ably (1.2.29):
+  - Ably (1.2.30):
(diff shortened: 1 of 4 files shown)


ttypic (approved):
LGTM

ttypic (changes requested):
@umair-ably could you update lock-files for example app and integration test?

```sh
cd example/ios
pod update Ably
cd ../../test_integration/ios
pod update Ably
```
should work

Issue #527, opened 2024-06-17
Crash on IPhone 7 plus


ashishp-dev:
[NSTaggedPointerString stringValue]: unrecognized selector sent to instance 0xbeda7c7407f36903
 *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[NSTaggedPointerString stringValue]: unrecognized selector sent to instance 0xbeda7c7407f36903'
*** First throw call stack:
(0x180927c60 0x198153ee4 0x1809f86e8 0x1808c154c 0x1808c075c 0x10634145c 0x1808d28c8 0x1809060f8 0x1063411fc 0x106340c6c 0x106318aa4 0x106317504 0x106316ec4 0x1063157ec 0x1805e8094 0x180595964 0x10631576c 0x1063149b0 0x1063140a8 0x10a9359ac 0x10a939638 0x111dd61e4 0x111845794 0x1805e7094 0x1805e8094 0x180594d44 0x180594994 0x1808e3014 0x1808a04f8 0x1808b3174 0x1a13ee988 0x1830b67d8 0x182e4fc70 0x10474901c 0x1060f04d0)
libc++abi: terminating with uncaught exception of type NSException
* thread #1, queue = 'io.ably.main', stop reason = signal SIGABRT
    frame #0: 0x00000001bb207bbc libsystem_kernel.dylib`__pthread_kill + 8
libsystem_kernel.dylib`:
->  0x1bb207bbc <+8>:  b.lo   0x1bb207bd8               ; <+36>
    0x1bb207bc0 <+12>: stp    x29, x30, [sp, #-0x10]!
    0x1bb207bc4 <+16>: mov    x29, sp
    0x1bb207bc8 <+20>: bl     0x1bb20360c               ; cerror_nocancel
Target 0: (Soul Train Dev) stopped.

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-4840) by [Unito](https://www.unito.io)

VeskeR:
Hi @ashishp-dev !

Could you please provide more context regarding the error? Specifically, what operation are you trying to perform (e.g. publish/receive a message), and which lines of code are causing the app to crash? Additionally, could you let me know what version of ably-flutter you are using?

Thank you!

ashishp-dev:
[...]

PR #528, opened 2024-06-20
fix(iOS): `transportParams` serialization

ttypic:
Resolves https://github.com/ably/ably-flutter/issues/527

Fixed `transportParams` serialization logic, it returned `NSDictionary` with `NSString` values instead of `ARTStringifiable`.


maratal (approved):
LGTM

umair-ably (approved):

Issue #533, opened 2024-07-25
Library doesn't respect RSH3a3a


maratal:
Library requests and feeds apple token to the ably-cocoa against the [spec](https://sdk.ably.com/builds/ably/specification/main/features/#RSH3a3). Shouldn't be doing that.

maratal:
Related issue in ably-cocoa - https://github.com/ably/ably-cocoa/issues/1951

PR #534, opened 2024-07-25
Removed push token handling to respect RSH3a3a

maratal:
Removed unnecessary push token handling. Token is requested by ably-cocoa and received by flutter, which sends it to ably via the first available `ARTRealtime` object. All instances have the same value for the `device.push` property, since the `device` itself is stored in a static variable.

Closes https://github.com/ably/ably-flutter/issues/533


ios/Classes/AblyInstanceStore.m
```diff
@@ -65,20 +65,21 @@ -(ARTPaginatedResult *) getPaginatedResult:(NSNumber *const) handle {
...
-    
-    for (id restHandle in _restInstances) {
-        ARTRest *const rest = _restInstances[restHandle];
-        [ARTPush didRegisterForRemoteNotificationsWithDeviceToken:deviceToken rest:rest];
-    }
-    for (id realtimeHandle in _realtimeInstances) {
-        ARTRealtime *const realtime = _realtimeInstances[realtimeHandle];
+// Set device token for the first created realtime object, device of which is available for others, because it's static
+-(void)didRegisterForRemoteNotificationsWithDeviceToken:(NSData *const)deviceToken {
+    ARTRealtime *const realtime = _realtimeInstances.allValues.firstObject;
```
ttypic, line 70: do we actually need this code? In what circumstances we have to call `didRegisterForRemoteNotificationsWithDeviceToken` or `didFailToRegisterForRemoteNotificationsWithError`? Shouldn't this be a part of state machine logic?

ios/Classes/AblyFlutter.m
```diff
@@ -814,16 +792,13 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
 }
 
 #pragma mark - Push Notifications Registration - UIApplicationDelegate
-/// Save the deviceToken provided so we can pass it to the first Ably client which gets created, in createRealtime or createRest.
+
 -(void)application:(UIApplication *)application didRegisterForRemoteNotificationsWithDeviceToken:(NSData *)deviceToken {
-    // Set deviceToken on all existing Ably clients, and a property which used for all future Ably clients.
```
ttypic, line 819: OOI, why have you decided to remove these comments?

maratal: Because the code setting deviceToken on all existing Ably clients was removed. I've left the comment about the single client though.

ios/Classes/AblyInstanceStore.m
maratal: Actually, no, we don't. But I've tried to simplify what is already here. I think you are right and we should remove all the APNs stuff from flutter.

maratal: I've changed how it works - https://github.com/ably/ably-flutter/pull/534/commits/18d12079312fec8bf63e8c869ba7d3b8aa24e67d

maratal: Now it stores the realtime/rest handle through which the token was requested and stores the token using the same realtime/rest instance. I think we should leave this code since removing it will break user's apps.

PR #542, opened 2024-08-15
ci: update platform versions

owenpearson:
fixes some issues with no longer supported versions

android CI is still broken even after these changes, but this at least gets iOS running again


ttypic:
Sorry missed this PR, changes already been made in other PR

Issue #551, opened 2024-10-04
Error on Android Build


LaBolaCuba:
Hi, i'm using ably with Flutter, on iOS is working fine but when i build for android this happend:

```
E/flutter (21393): <asynchronous suspension>
E/flutter (21393): #1      Platform.invokePlatformMethod (package:ably_flutter/src/platform/src/platform.dart:52:14)
E/flutter (21393): <asynchronous suspension>
E/flutter (21393): #2      Realtime.createPlatformInstance (package:ably_flutter/src/platform/src/realtime/realtime.dart:35:9)
E/flutter (21393): <asynchronous suspension>
E/flutter (21393): #3      PlatformObject._acquireHandle.<anonymous closure> (package:ably_flutter/src/platform/src/platform_object.dart:39:37)
E/flutter (21393): <asynchronous suspension>
E/flutter (21393): #4      PlatformObject._acquireHandle.<anonymous closure> (package:ably_flutter/src/platform/src/platform_object.dart:39:37)
E/flutter (21393): <asynchronous suspension>
E/flutter (21393): 
E/flutter (21393): [ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: MissingPluginException(No implementation found for method createRealtime on channel io.ably.flutter.plugin)
E/flutter (21393): #0      MethodChannel._invokeMethod (package:flutter/src/services/platform_channel.dart:332:7)
E/flutter (21393): <asynchronous suspension>
E/flutter (21393): #1      Platform.invokePlatformMethod (package:ably_flutter/src/platform/src/platform.dart:52:14)
E/flutter (21393): <asynchronous suspension>
E/flutter (21393): #2      Realtime.createPlatformInstance (package:ably_flutter/src/platform/src/realtime/realtime.dart:35:9)
E/flutter (21393): <asynchronous suspension>
E/flutter (21393): #3      PlatformObject._acquireHandle.<anonymous closure> (package:ably_flutter/src/platform/src/platform_object.dart:39:37)
E/flutter (21393): <asynchronous suspension>
E/flutter (21393): #4      PlatformObject._acquireHandle.<anonymous closure> (package:ably_flutter/src/platform/src/platform_object.dart:39:37)
E/flutter (21393): <asynchronous suspension>
E/flutter (21393): 
```

How i can solve this?, i'm going creazy with this. Ably on android don't work at all.

Env:
```
Flutter 3.24.3 • channel stable • https://github.com/flutter/flutter.git
[...]

PR #554, opened 2024-10-10
fix: add missing warn logLevel serialization

ttypic:
Resolves https://github.com/ably/ably-flutter/issues/551

## Summary by CodeRabbit

- **New Features**
	- Enhanced logging capabilities by adding support for the `warn` log level in the codec's logging mechanism.

- **Bug Fixes**
	- Improved handling of log levels to ensure accurate processing of the `warn` log level.


sacOO7 (approved):
LGTM. 
Do we need to add test for logging btw, to avoid future exceptions?

ttypic:
thanks! I don't think it's necessary to cover this with test. If we cover only `warn` level it won't be very useful, if we cover all serialization keys in flutter it will be too much work

PR #560, opened 2025-01-26
[ECO-5119] Opt out default push handler

### LINKED ISSUE #557: Crash app when open notification
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=2, address=0x16d527fd0)
    frame #0: 0x0000000180aa5c18 CoreFoundation`__CFStrAllocateMutableContents
CoreFoundation`:
->  0x180aa5c18 <+0>:  stp    x20, x19, [sp, #-0x20]!
    0x180aa5c1c <+4>:  stp    x29, x30, [sp, #0x10]
    0x180aa5c20 <+8>:  add    x29, sp, #0x10
    0x180aa5c24 <+12>: add    x8, x0, #0x8
    
dependencies:
    ably_flutter: ^1.2.35

   
    
    In android/src/main/AndroidManifest.xml
    `
...
<meta-data
            android:name="com.google.firebase.messaging.default_notification_channel_id"
            android:value="test" />
        <meta-data
            android:name="com.google.firebase.messaging.default_notification_icon"
            android:resource="@mipmap/ic_notification" />
        <meta-data
            android:name="com.google.firebase.messaging.default_notification_color"
            android:resource="@color/primary" />
        <receiver android:name="io.ably.flutter.plugin.push.FirebaseMessagingReceiver"
            tools:node="remove">
        </receiver>
...
`

In iOS/AppDelegate.swift
`
...
override func application(_ application: UIApplication, didRegisterForRemoteNotificationsWithDeviceToken deviceToken: AblyData) {
       AblyInstanceStore.sharedInstance().didRegisterForRemoteNotifications(withDeviceToken: deviceToken)
    }

    override func application(_ application: UIApplication, didFailToRegisterForRemoteNotificationsWithError error: Error) {
       AblyInstanceStore.sharedInstance().didFailToRegisterForRemoteNotificationsWithError_error = error;
       NSLog("application:didFailToRegisterForRemoteNotificationsWithError was called with error: %@", error.localizedDescription)
    }
...
`

maratal:
Closes #557 

Some users use APNs via FCM, so there should be an option to turn-off default handling of push notifications in iOS Ably Flutter to give user an ability to handle it themselves or via FCM framework (similar opt-out exists in android's Ably Flutter via  AndroidManifest).

### DIFF
--- a/example/ios/Runner/AppDelegate.swift
+++ b/example/ios/Runner/AppDelegate.swift
@@ -1,16 +1,16 @@
 import UIKit
 import Flutter
 
-@UIApplicationMain
+@main
 @objc class AppDelegate: FlutterAppDelegate {
-  override func application(
-    _ application: UIApplication,
-    didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?
-  ) -> Bool {
-    GeneratedPluginRegistrant.register(with: self)
-    return super.application(application, didFinishLaunchingWithOptions: launchOptions)
-  }
-    
+    override func application(
+        _ application: UIApplication,
+        didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?
+    ) -> Bool {
+        GeneratedPluginRegistrant.register(with: self)
+        return super.application(application, didFinishLaunchingWithOptions: launchOptions)
+    }
+
     override func application(_ application: UIApplication, didFailToRegisterForRemoteNotificationsWithError error: Error) {
         NSLog("application:didFailToRegisterForRemoteNotificationsWithError was called with error: %@", error.localizedDescription)
     }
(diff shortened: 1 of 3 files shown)


lawrence-forooghian:
Hey @maratal, I'm here in response to us being asked if we intend to merge this into `main`. I haven't been following this issue closely, but as I understand it:

- the Flutter SDK has some code that is called in response to _some_ iOS lifecycle event (which?)
- this code was causing a crash

So, my questions:

- did we fix the crash?
- why do we need to give the option of disabling this handling of the iOS lifecycle event? is it because this handling causes a crash, or is it because this handling prevents users from handling push notifications themselves, or both?

lawrence-forooghian:
If we make the handling opt-out, what guidance would we give to customers for this option? It would have to be something like "some users have experienced crashes that may be related to this SDK's default handling of push notifications; here's how to opt out". Is that the plan?

If we were to make it opt-in, what would be the consequences for the existing users of the library?

As for reproducing the crash, do we have any customer who has been able to reproduce it?

PR #567, opened 2025-06-14
NullPointerException fix.

maratal:


paddybyers:
I'm guessing we'd want equivalent checks for `error()` and `endOfStream()`.

I guess the root problem is some lack of synchronisation somewhere, but for now we just need to stop it from crashing

Issue #572, opened 2025-06-26
Use stable IDs for `StreamsChannel` class


ttypic:
There is some evidence that the `StreamsChannel` class, which is used to stream values from platform calls (methods like `subscribe`, `onXXX`), can cause the app to crash. According to the error stack trace, the issue appears to be due to a type cast error.

Upon reviewing the code, the only plausible cause seems to be a collision in the generated IDs. Currently, `StreamsChannel` uses a simple counter to generate unique IDs. On the platform side, these IDs are parsed and set, then used to send received values. Listeners for these IDs are registered on the Dart side.

It appears that, in some cases, a collision may occur where the platform-side code outlives the Dart-side code. As a result, the platform sends an event with an old ID, which is received by a newly registered listener on the Dart side. This listener expects a different value type, leading to a crash due to the mismatched type.

The root cause of this collision is unclear and difficult to reproduce. According to crash reports, it seems to affect only Samsung devices.

To fix this issue, we can use stable IDs instead of a simple counter

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-5427) by [Unito](https://www.unito.io)

PR #573, opened 2025-06-26
[ECO-5427] fix: use stable 64-bit ID generation for streams and improve platform reset handling

ttypic:
Fixes https://github.com/ably/ably-flutter/issues/572

- Replaced `int` identifiers with `long` to support 64-bit IDs in `StreamsChannel`
- Introduced `StableIdGenerator` for generating unique 64-bit IDs
- Adjusted `StreamsChannel` and platform/device communication to use stable IDs
- Improved Ably client reset logic to avoid conflicts after reinitialization


sacOO7 (approved):
Overall PR looks good to me

PR #601, opened 2026-05-26
ci: GHA workflow security cleanup

emptyhammond:
Routine hygiene pass over the GitHub Actions workflows in this repo, based on an automated audit. The changes are split into four commits, one per finding type.

- **Disable credential persistence on checkout** — every `actions/checkout` step now sets `persist-credentials: false`. None of these jobs push back to the repo, so leaving the `GITHUB_TOKEN` in the runner's git config after checkout serves no purpose.
- **Declare minimum permissions** — each workflow now sets an explicit `permissions:` block (top-level `contents: read`) so jobs no longer receive the default broad `GITHUB_TOKEN` scopes. Where a job legitimately needs more (e.g. `id-token: write` for OIDC, `deployments: write` for docs/features uploads), it keeps a job-level override.
- **Pin actions to commit SHAs** — third-party `uses:` references (`actions/checkout`, `actions/setup-java`, `subosito/flutter-action`, `aws-actions/configure-aws-credentials`, `ably/sdk-upload-action`, `reactivecircus/android-emulator-runner`, and the reusable workflows `ably/features` and `dart-lang/setup-dart`) are pinned to a specific commit SHA with the original tag retained as a trailing comment. This prevents an upstream tag from being silently re-pointed.
- **Stop inheriting all secrets into the features workflow** — the `features.yml` call to `ably/features/.../sdk-features.yml` previously used `secrets: inherit`, which forwards every repo/org secret. It now forwards only `ABLY_AWS_ACCOUNT_ID_SDK`, the single secret the called workflow declares.

No behaviour changes are expected; this is a hardening-only PR.

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -5,6 +5,9 @@ on:
     branches:
       - main
 
+permissions:
+  contents: read
+
 jobs:
   check:
     runs-on: ubuntu-latest
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -11,6 +11,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          persist-credentials: false
 
       - uses: subosito/flutter-action@v2
         with:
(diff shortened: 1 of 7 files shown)


ttypic (approved):
LGTM

